PR #14, opened 2019-01-14
Reserve compute resources for system daemons

to-bar:


core/core/src/ansible/roles/master/defaults/main.yml
```diff
@@ -1,8 +1,13 @@
...
+  evictionHard:
+    memory.available: 200Mi
+# Reserve resources for Kubernetes system daemons that are not run as pods (kubelet.service + docker.service cgroups)
+  kubeReserved:
+    cpu: 50m
+    memory: 256Mi
+# Reserve resources for other non-pod processes (like OS daemons and user login sessions)
+  systemReserved:
+    cpu: 50m
+    memory: 768Mi
```
toszo: Please drop a comment how this values was calculated or estimated.

core/core/src/ansible/roles/worker/defaults/main.yml
```diff
@@ -0,0 +1,13 @@
...
+  evictionHard:
+    memory.available: 200Mi
+# Reserve resources for Kubernetes system daemons that are not run as pods (kubelet.service + docker.service cgroups)
+  kubeReserved:
+    cpu: 50m
+    memory: 256Mi
+# Reserve resources for other non-pod processes (like OS daemons and user login sessions)
+  systemReserved:
+    cpu: 50m
+    memory: 768Mi
```
toszo: Please drop a comment how this values was calculated or estimated.

PR #25, opened 2019-01-28
Added app deployments with authentication service.

seriva:


core/core/src/ansible/roles/deployments/tasks/applications/auth-service/validation.yml
```diff
@@ -0,0 +1,10 @@
+- name: Verify number of instances
```
toszo: I think it is not valid validation now.

core/core/src/ansible/roles/deployments/templates/auth-service/auth-service.yml.j2
```diff
@@ -0,0 +1,242 @@
...
+      schedulerName: default-scheduler
+      securityContext:
+        fsGroup: 1000
+        runAsNonRoot: true
+        runAsUser: 1000
+      terminationGracePeriodSeconds: 60
+      volumes:
+      - configMap:
+          defaultMode: 365
+          name: {{ auth_service_name }}
```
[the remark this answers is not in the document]
toszo: Don't remember exactly but there was some issue with keycloak initialization when there was consistency in using quotes.

core/core/src/ansible/roles/deployments/tasks/main.yml
```diff
@@ -0,0 +1,13 @@
+---
+
+- name: Include deployments
+  include_tasks: "applications/{{ item.name }}/main.yml"
+  vars:
+    data: "{{ item }}"
+  loop: "{{ kubernetes.deployments }}"
```
to-bar: It seems there is a bug or strange behavior of Ansible: loop is parsed even when condition is not met.
Workaround to consider is:
```
loop: "{{ kubernetes.deployments if kubernetes.deployments else [] }}"
when:
  - kubernetes.deployments is defined
  - kubernetes.deployments
```

core/core/src/ansible/roles/deployments/tasks/main.yml
```diff
@@ -0,0 +1,13 @@
...
+  when:
+      - kubernetes.deployments is defined and kubernetes.deployments | length > 0
+
```
to-bar: How about having just one empty line? ;)

core/core/src/ansible/roles/deployments/templates/auth-service/auth-service.yml.j2
```diff
@@ -0,0 +1,242 @@
...
+    stop-embedded-server
+  keycloak.sh: |
+    #!/usr/bin/env bash
+
+    set -eu
+
+    /opt/jboss/keycloak/bin/jboss-cli.sh --file=/scripts/keycloak.cli
+
+    exec /opt/jboss/tools/docker-entrypoint.sh -b 0.0.0.0  -c standalone-ha.xml
+    exit "$?"
```
to-bar: $? is exit code (always integer) so I think quotation marks are not needed

core/core/src/ansible/roles/master/tasks/registry-secrets.yml
```diff
@@ -0,0 +1,7 @@
+- name: Apply docker-registry secrets
+  shell: kubectl --kubeconfig=/home/{{ admin_user.name }}/.kube/config create secret docker-registry {{ item.name }} --docker-server={{ item.server_url }} --docker-username={{ item.username }} --docker-password="{{ item.password }}" --docker-email={{ item.email }} --namespace={{ item.namespace }} --dry-run -o yaml | kubectl apply -f -
+  become_user: "{{ admin_user.name }}"
+  loop: "{{ kubernetes.image_registry_secrets }}"
```
to-bar: Same as for deployments, workaround is:
`loop: "{{ kubernetes.image_registry_secrets if kubernetes.image_registry_secrets else [] }}"`

core/core/src/ansible/roles/deployments/tasks/main.yml
toszo: Will not be changed.

core/core/src/ansible/roles/master/tasks/registry-secrets.yml
toszo: Will not be changed.

core/core/src/ansible/roles/deployments/templates/auth-service/auth-service.yml.j2
toszo: But also are not bad here.

PR #31, opened 2019-01-31
Set Filebeat to be started after Docker

to-bar:


core/core/src/ansible/roles/docker/handlers/main.yml
```diff
@@ -1,5 +1,5 @@
 ---
-- name: Restart docker
+- name: Restart Docker
   service:
```
lukurde: Other `service` modules were replaced with native `systemd`. Why not here?

core/core/src/ansible/roles/filebeat/tasks/main.yml
```diff
@@ -15,3 +15,7 @@
   systemd:
     name: filebeat
     state: started
+
+- name: Verify Filebeat is running
+  command: systemctl is-active filebeat
+  changed_when: false
```
lukurde: This task will never be changed nor failed. Shouldn't we register output/exit code of systemctl and make a condition that'll fail the task if Filebeat is not active?

core/core/src/ansible/roles/filebeat/tasks/configure-filebeat.yml
```diff
@@ -7,12 +7,26 @@
...
+        path: /etc/systemd/system/filebeat.service.d
+        state: directory
+    - name: Copy drop-in configuration file (extra-dependencies.conf)
+      template:
+        dest: /etc/systemd/system/filebeat.service.d/extra-dependencies.conf
+        src: extra-dependencies.conf.j2
+      register: filebeat_unit_dependencies
+      notify: Restart Filebeat service
+  when: inventory_hostname in groups['master'] or inventory_hostname in groups['worker']
+
```
lukurde: This block should be placed before "Copy configuration file (filebeat.yml)" <- there's a handler for restarting Filebeat.

first change configuration
then enable and restart services

lukurde: Whoops. My bad, handlers are flushed later in main.yml. Looks good.

to-bar: That's ok since restart is triggered later, in main.yml by flush_handlers task.

core/core/src/ansible/roles/filebeat/tasks/main.yml
to-bar: It fails whenever the command returns non-zero exit code (when service is not running), so the verification is correct.

PR #36, opened 2019-02-04
Feature/postgresql replication documentation

erzetpe:


docs/home/HOWTO.md
```diff
@@ -147,6 +149,25 @@ sudo -u postgres -i
...
+      enable: yes
+      user: your-postgresql-replication-user
+      password: your-postgresql-replication-password
+      max_wal_senders: 10 # optional - default value 5
+      wal_keep_segments: 34 # optional - default value 32
+```
+If enable is set to yes in replication then Epiphany will automatically create cluster of master and slave server with replication user with name and password
+specified in data.yaml.
+
+To override default values of max_wal_senders and wal_keep_segments provide values for wal_keep_segments and max_wal_senders in replication section of postgresql in data.yaml.
```
to-bar: I think this line can be removed, it's clear from the context and example section that these defaults values can be overridden.

docs/home/HOWTO.md
```diff
@@ -147,6 +149,25 @@ sudo -u postgres -i
 And then configure database server using psql according to your needs and
 PostgreSQL documentation, to which link you can find at <https://www.postgresql.org/docs/>
 
+### How to configure PostgreSQL replication
+
+In order to configure PostgreSQL replication to your data.yaml add block similar to the one below to core section:
```
lukurde:
```suggestion
In order to configure PostgreSQL replication add postgresql block to your data.yaml (core section) similar to the one below:
```

to-bar: I think this would sound better: In order to configure PostgreSQL replication add to your data.yaml a block similar...

PR #38, opened 2019-02-04
grammar drama, windows specific info added, conflict resolved

lukurde:


docs/home/HOWTO.md
```diff
@@ -133,10 +133,28 @@ To get it from the registry and run it:
...
+```LOCAL_DATA_DIR``` should be the host input directy for your data YAMLs and certificates.  ```LOCAL_BUILD_DIR``` should be the host directory where you want the Epiphany engine to write its build output. If everything is ok you will be presented with a Bash prompt from which one can run the Epiphany engine. Note that when filling in your data YAMLs one needs to specify the paths from the container's point of view.
 
 [`Azure specific`] Ensure that you have already enough resources/quotas accessible in your region/subscription on Azure before you run Epiphany - depending on your configuration it can create large number of resources.
 
+### Note for Windows users
+
+Watch out for the end-of-lines symbols conversion. By default Git for Windows sets "core.autocrlf" to "true". Mounting such files with Docker results in "^M" end-of-line character in the config files.
```
to-bar: I would suggest: Watch out for the line endings conversion.

For future changes you can also consider using code-span syntax for keywords like `core.autocrlf` and `true` instead of double quotes.

More info about code-span:
https://github.github.com/gfm/#code-span
You can check this is common approach also used by Git: https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration

to-bar: I would suggest: Watch out for the line endings conversion.

For future changes you can also consider using code-span syntax for keywords like `core.autocrlf` and `true` instead of double quotes.

More info about code-span:
https://github.github.com/gfm/#code-span
You can check this is common approach also used by Git: https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration

PR #98, opened 2019-02-18
Join node when kubelet config.yaml does not exist

### LINKED ISSUE #92: Scaling cluster horizontally failes

to-bar:
Fixes #92


to-bar:
Testing scenario:
1. Run on a worker node (here vm-k8s-node-001): `sudo kubeadm reset`
2. Run on master: `kubectl get nodes | grep node-001`
`vm-k8s-node-001     Ready    <none>   47m     v1.13.0`

Conclusion: checking whether a node should be joined to cluster using only `get nodes` can be insufficient

PR #110, opened 2019-02-19
Rabbitmq improvements

toszo:
#108, #109


lukurde:
I like the port configurability in data.yaml. Didn't find obvious issues with the rest but someone more knowledgeable in C# should review this.

Issue #24, opened 2019-01-24
Upstream monitoring


toszo:
Research and support different third party monitoring components as an upstream for Epiphany Platform.

mmakow:
Current (initial) scope: push data from Prometheus to upstream Prometheus. 

Other tools will be added if needed. Requires discussions with @patrick-ryan1 and applications (starting with APM)

toszo:
For the step 1 - federation of Prometheus - only documentation changes was applied. 
https://github.com/epiphany-platform/epiphany/blob/develop/docs/home/HOWTO.md#how-to-configure-scalable-prometheus-setup

PR #119, opened 2019-02-20
Item: #24 Desc: Add information to how-to how configure manually fede…

erzetpe:
…ration of Prometheus instances


docs/home/HOWTO.md
```diff
@@ -305,6 +306,37 @@ https://prometheus.io/docs/prometheus/latest/querying/examples/
 
 Right now we are only supporting email messages, but we are working heavily on introducing integration with Slack and Pager Duty.
 
+## How to configure Prometheus federation
```
toszo: In title and line below you can explain what is federation and why you need it

Issue #147, opened 2019-03-04
ZooKeeper 3.4.12 package not available


przemyslawdy:
Cannot download ZooKeeper 3.4.12 package.
Status code 404 for http://www.us.apache.org/dist/zookeeper/zookeeper-3.4.12/zookeeper-3.4.12.tar.gz
Possible solutions:
- upgrade to stable version 3.4.13
- or change yaml files to use older release from the archives (change url variable to https://archive.apache.org/dist/zookeeper)

PR #150, opened 2019-03-06
Fix ZooKeeper distribution URL

to-bar:
ZooKeeper v3.4.12 is no longer available at http://www.us.apache.org/dist/zookeeper


to-bar:
Fixes #147

PR #312, opened 2019-07-05
License file generation

seriva:


toszo:
Name of this PR is not really helping to understand what changes are.

core/src/epicli/cli/epicli.py
```diff
@@ -18,7 +19,10 @@ def main():
         formatter_class=argparse.RawDescriptionHelpFormatter)
 
     # setup some root arguments
-    parser.add_argument('--version', action='version', version=VERSION)
+    parser.add_argument('--version', action='version', help='Shows the CLI version', version=VERSION)
+    parser.add_argument('--licenses', action='version',
```
toszo: action='version' same as for --version, is it correct?

seriva: Yes the version action just returns a string so Im using that to just return the license info under a different command. No need to make it more complicated:)

PR #333, opened 2019-07-08
Docker install fix, Monitoring fix

toszo:
Fixes for monitoring (Prometheus + Kafka) #332, #331, #330
Docker install fix for RHEL, default sec rules for VMs in AWS #317


core/src/epicli/data/common/ansible/playbooks/roles/jmx_exporter/defaults/main.yml
```diff
@@ -0,0 +1,9 @@
+jmx_exporter_user: jmx-exporter
```
erzetpe: Why this is not in epiphany configuration defaults?

core/src/epicli/data/common/ansible/playbooks/roles/jmx_exporter/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+  become: yes
+  file:
+    path: "{{ item }}"
+    recurse: yes
+    owner: root
+    group: "{{ jmx_exporter_group }}"
+    mode: 0750
+    state: directory
+  with_items:
+  - /opt/jmx-exporter/jars
```
erzetpe: Shouldn't this be also configurable or use default parameters somehow?

core/src/epicli/data/common/defaults/configuration/jmx-exporter.yml
```diff
@@ -3,3 +3,9 @@ version: 0.3.0
 title: "JMX exporter"
 name: default
 specification:
+  description: "Service that runs JMX Exporter"
```
erzetpe: Configs here and in role default, why?

PR #343, opened 2019-07-10
CLI - upgrade, backup, restore commands; username moved to group_vars

toszo:


core/src/epicli/cli/engine/PatchEngine.py
```diff
@@ -0,0 +1,62 @@
...
+        inventory_path = get_inventory_path(build_directory)
+        self.ansible_command.run_playbook(inventory=inventory_path, playbook_path=backup_role_path)
+
+    def run_recovery(self, build_directory):
+        backup_role_path = os.path.join(build_directory, 'ansible', 'recovery.yml')
+        inventory_path = get_inventory_path(build_directory)
+        self.ansible_command.run_playbook(inventory=inventory_path, playbook_path=backup_role_path)
+
+    @staticmethod
+    def get_inventory_path(build_directory):
```
seriva: I suggest moving this to build_saver.py where all these get_{whatever-path}_methods live.

core/src/epicli/cli/epicli.py
```diff
@@ -103,6 +129,24 @@ def run_init(args):
...
+def run_upgrade(args):
+    Config().output_dir = args.build_directory
+    with PatchEngine() as engine:
+        engine.run_upgrade(Config().output_dir)
+
+
+def run_backup(args):
+    Config().output_dir = args.build_directory
+    with PatchEngine() as engine:
+        engine.run_backup(Config().output_dir)
```
seriva: I suggest not passing down the path since the whole idea of this global config is to access it throughout the CLI without passing down the same parameters all the time.

core/src/epicli/cli/epicli.py
```diff
@@ -103,6 +129,24 @@ def run_init(args):
...
+
+
+def run_recovery(args):
+    Config().output_dir = args.build_directory
+    with PatchEngine() as engine:
+        engine.run_recovery(Config().output_dir)
```
seriva: See above.

Issue #357, opened 2019-07-16
Haproxy TLS through communication


erzetpe:
Haproxy should enable forwarding of TLS communication.

PR #373, opened 2019-07-24
Item: #357  Desc: Add TLS through mode for haproxy

erzetpe:


core/src/epicli/data/common/ansible/playbooks/roles/haproxy/templates/haproxy_RedHat.cfg.j2
```diff
@@ -61,25 +61,37 @@ frontend {{ front.name }}
...
   {%- endfor %}
   {%- endif %}
 {%- endfor %}
 
-    {%- if specification.stats is defined %}
+{%- if specification.stats is defined %}
     {%- if specification.stats.enable %}
 frontend stats
-    mode https
+    mode http
```
toszo: Why http and no https? 
It will break the sd config for Prometheus in core/core/src/ansible/roles/haproxy_exporter/templates/prometheus-haproxy-exporter.service.j2

erzetpe: As this is stated in documentation, there is no such mode as https. 

mode { tcp|http|health }

More ind documentation below:
http://cbonte.github.io/haproxy-dconv/1.8/configuration.html#4.2-mode

PR #380, opened 2019-07-30
Added missing default values for logging machine

toszo:


core/src/epicli/data/aws/defaults/infrastructure/virtual-machine.yml
```diff
@@ -400,4 +400,64 @@ specification:
        destination_to_port: 65536
        destination_port_range: "0"
        source_address_prefix: "10.1.2.0/24"
-       destination_address_prefix: "*"
\ No newline at end of file
+       destination_address_prefix: "*"
```
lukurde: destination_address_prefix (in multiple places):
sometimes it's "*", sometimes it's "0.0.0.0/0"

Shouldn't we pick one and stick with it?

toszo: Done

Issue #384, opened 2019-07-30
Introduce canal network plugin


toszo:
Releted to #364

Issue #364, opened 2019-07-20
Pod fails to resolve DNS names when Calico installed on Azure


toszo:
To reproduce: 
1. Login to cluster installed on Azure with calico plugin
2. kubectl apply -f https://k8s.io/examples/admin/dns/busybox.yaml
3. kubectl exec -ti busybox -- nslookup kubernetes.default
4. kubectl exec -ti busybox -- nslookup google.com
(https://kubernetes.io/docs/tasks/administer-cluster/dns-debugging-resolution/)

Think that should be investigated at first: 
IP_AUTODETECTION_METHOD for calico - since by default is set to "first_found" which can detect not appropriate IP.

toszo:
https://docs.projectcalico.org/v3.8/reference/public-cloud/azure#about-calico-on-azure

toszo:
Created #384 to introduce canal plugin

to-bar:
According to Calico documentation, Azure does not support Calico networking because it doesn't allow traffic with unknown source IPs.
We decided to introduce Canal. It's is a shorthand for saying "Calico and Flannel", a common practice which sets up Calico to handle policy management and Flannel to manage the network itself. This combination brings in Calico's support for the NetworkPolicy feature of Kubernetes, while utilizing Flannel's UDP-based network traffic to provide for an easier setup experience that works in a wider variety of host network environments without special configuration.

PR #385, opened 2019-07-30
Item: #384, Kubernetes canal - cni plugin

toszo:
Introduction of canal - which is a combination of flannel networking + calico policies. This implementation is required due to lack of support calico networking on Azure. 

Related:
#364


toszo:
Added canal cni to legacy also.

PR #423, opened 2019-08-06
Opened ports documentation

erzetpe:


docs/home/SECURITY.md
```diff
@@ -20,4 +21,68 @@ Additional to users created by each component Epiphany creates also users and gr
   - rabbitmq/rabbitmq
   - zookeeper/zookeeper
   - kafka/kafka
+  
+Other accounts created by each component you can find in these components documentation.
+
+
+### Ports used by components in Epiphany
+
+Below you can find list of ports used in Epiphany on per component basis:
```
seriva: Suggest adding info that these are default ports since some of them might be changed with config.

PR #427, opened 2019-08-07
AzureCLI integration

seriva:
- Added login to Azure-CLI and selection of subscription.
- Added creation of service principle.
- Added creation of the resource group.
- Added possibility to run without service principal


core/src/epicli/data/common/defaults/epiphany-cluster.yml
```diff
@@ -11,8 +11,10 @@ specification:
     key_path: /root/.ssh/epiphany-operations/id_rsa # YOUR-SSH-KEY-PATH
   cloud:
     subscription_name: YOUR-SUB-NAME
+    resource_group_name: YOUR-RESOURCE-GROUP-NAME
     vnet_address_pool: 10.1.0.0/20
     use_public_ips: False # When not using public IPs you have to provide connectivity via private IPs (VPN)
+    use_service_principle: False
```
erzetpe: should be probably use_service_principal

erzetpe: do we also need this for AWS?

core/src/epicli/data/common/defaults/epiphany-cluster.yml
```diff
@@ -11,8 +11,10 @@ specification:
     key_path: /root/.ssh/epiphany-operations/id_rsa # YOUR-SSH-KEY-PATH
   cloud:
     subscription_name: YOUR-SUB-NAME
+    resource_group_name: YOUR-RESOURCE-GROUP-NAME
```
erzetpe: shouldn't this be created based on cluster name, similar to this what we have on AWS? we also are not using this parameter on AWS in terraform templates probably?

seriva: This is already configurable for AWS as well

PR #487, opened 2019-09-12
Feature/azure-k8n-storage

seriva:
- Added shared storage for k8n


core/src/epicli/data/azure/defaults/infrastructure/storage-share.yml
```diff
@@ -0,0 +1,9 @@
+kind: infrastructure/storage-share
+version: 0.3.0
```
lukurde: Is that correct? Other versions in this PR were bumped to 0.4.0

core/src/epicli/data/common/ansible/playbooks/roles/kubernetes_master/templates/azure/k8s-persistent-volume.yml.j2
```diff
@@ -1,10 +1,14 @@
...
-  name: "pv0001" 
+metadata:
+  name: {{ specification.storage.name }}
+  labels:
+    name: {{ specification.storage.name }}
 spec:
+  storageClassName: azurefile
   capacity:
-    storage: "15Gi" 
+    storage: {{ specification.storage.capacity }}Gi
```
lukurde: Are we sure we want the suffix hard-coded to Gibibytes?

core/src/epicli/data/azure/defaults/infrastructure/storage-share.yml
seriva: Fixed

core/src/epicli/data/common/ansible/playbooks/roles/kubernetes_master/templates/azure/k8s-persistent-volume.yml.j2
seriva: As per Tomasz request yes.

lukurde: Ok, then

PR #515, opened 2019-09-20
Roles to use private repo

toszo:
Prometheus, Grafana, Exporters, Kafka, Zookeeper, Kubernetes, Applications, Calico version to 3.8.1


core/src/epicli/data/common/defaults/configuration/feature-mapping.yml
```diff
@@ -9,39 +9,39 @@ specification:
     - name: kubernetes-node
       enabled: yes
     - name: elasticsearch
-      enabled: yes
+      enabled: no
```
seriva: Shouldnt these be turned back on to yes?

toszo: It should, will fix it

PR #517, opened 2019-09-20
Add scripts and repository role for setup and teardown

erzetpe:


core/src/epicli/data/aws/defaults/infrastructure/virtual-machine.yml
```diff
@@ -223,14 +223,14 @@ specification:
        destination_port_range: "22"
        source_address_prefix: "0.0.0.0/0"
        destination_address_prefix: "0.0.0.0/0"
-     - name: node_exporter
-       description: Allow node_exporter traffic
+     - name: repository
```
seriva: Seems you replaced the nope_exporter entry with this one. Shouldn't there be both?

core/src/epicli/cli/engine/ansible/AnsibleRunner.py
```diff
@@ -47,11 +47,20 @@ def run(self):
...
+
+        if repository_setup_play_result != 0:
+            return
+
         common_play_result = self.ansible_command.run_playbook_with_retries(inventory=inventory_path,
                                                                             playbook_path=os.path.join(
                                                                                 get_ansible_path(
                                                                                     self.cluster_model.specification.name),
-                                                                                "common.yml"), retries=5)
+                                                                                "common.yml"), retries=1)
```
seriva: Is there a reason we only want one retry for the other playbooks now?

core/src/epicli/cli/engine/ansible/AnsibleRunner.py
```diff
@@ -47,11 +47,20 @@ def run(self):
 
         self.ansible_vars_generator.run()
 
+        repository_setup_play_result = self.ansible_command.run_playbook_with_retries(inventory=inventory_path,
```
seriva: I think you need to check the offline flag here.  Dont we only want to execute this when we are in online mode?

core/src/epicli/data/aws/defaults/infrastructure/virtual-machine.yml
erzetpe: Need to fix.

erzetpe: Fixed.

PR #529, opened 2019-09-26
Feature/offline mode

seriva:
Move Skopeo for copying to repo-host by Ansible and add message for long running repo playbook.


seriva:
- Move Skopeo for copying to repo host and add message for long running repo task

PR #654, opened 2019-10-25
Add preflight to check OS version before installation

erzetpe:


core/src/epicli/data/common/ansible/playbooks/preflight.yml
```diff
@@ -0,0 +1,9 @@
+---
+# Ansible playbook that makes sure the base items for all nodes are installed
```
lukurde:
```suggestion
# Ansible playbook that checks if the preflight conditions are met
```

PR #657, opened 2019-10-28
Feature/offline upgrade

seriva:
- #640 
- #644


core/src/epicli/cli/helpers/build_saver.py
```diff
@@ -10,8 +10,11 @@
 MANIFEST_FILE_NAME = 'manifest.yml'
 SP_FILE_NAME = 'sp.yml'
 INVENTORY_FILE_NAME = 'inventory'
+INVENTORY_FILE_NAME_LEGACY = 'inventory/development'
```
toszo: For legacy builds this path will not always be like that. Possible is: inventosy/staging, inventory/production, or other (was defined by user)

seriva: But the it will be in the folder inventory/?

erzetpe: This should be always in the folder inventory, but names can differ.

seriva: Fixed

PR #669, opened 2019-10-31
Replacing jq from github on upgraded machines with jq from repository

erzetpe:


core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/common/upgrade-common.yml
```diff
@@ -0,0 +1,23 @@
...
+    state: present
+  when: ansible_os_family == "Debian"
+
+- name: Install jq RedHat family packages
+  yum:
+    name:
+      - jq
+    update_cache: yes
+    state: present
+  when: ansible_os_family == "RedHat"
```
toszo: I think you can use "package" ansible module since name of the package is the same.

Issue #680, opened 2019-11-05
[upgrade] TASK [upgrade : verify-upgrade] failing - syntax error


przemyslavic:
> 16:12:33 INFO cli.engine.ansible.AnsibleCommand - FAILED - RETRYING: verify-upgrade | Verify kubelet version from API server and get node status (1 retries left).
16:12:37 INFO cli.engine.ansible.AnsibleCommand - fatal: [vm-qa-upgrades-rhel-master-001]: FAILED! => {"attempts": 30, "changed": false, "cmd": "set -o pipefail && kubectl get nodes vm-qa-upgrades-rhel-master-001 | # get values only for STATUS and VERSION columns, example output: 'Ready v1.14.6' awk 'NR==1 { for (col=1; col<=NF; col++) { columns[$col] = col } };\n     NR>1  { print $columns[\"STATUS\"], $columns[\"VERSION\"] }'", "delta": "0:00:00.004863", "end": "2019-11-04 16:12:37.005603", "msg": "non-zero return code", "rc": 1, "start": "2019-11-04 16:12:37.000740", "stderr": "/bin/sh: -c: line 1: unexpected EOF while looking for matching `''\n/bin/sh: -c: line 2: syntax error: unexpected end of file", "stderr_lines": ["/bin/sh: -c: line 1: unexpected EOF while looking for matching `''", "/bin/sh: -c: line 2: syntax error: unexpected end of file"], "stdout": "", "stdout_lines": []}

przemyslavic:
fixed in #689

PR #689, opened 2019-11-06
Set imageRepository in kubeadm-config ConfigMap to use local registry

to-bar:
* Set imageRepository in kubeadm-config ConfigMap to use local registry
* Fix for added comment in verify-upgrade (#680)
* Check if Docker is installed like a pro
* Reconfigure Docker only if necessary
* Reconfigure Docker for pulling images from local registry


to-bar:
Fixes #680

Issue #694, opened 2019-11-07
epicli apply does not remove from build files removed from sources


to-bar:
Steps to Reproduce: 

1. Remove file from sources, in my case it was: `core/src/epicli/data/common/ansible/playbooks/roles/image_registry/meta/main.yml`
2. Re-run for already existing build `epicli apply -f ${workspaceFolder}/clusters/YOUR_DATA_YAML.yml`

The file still exists in the build (`clusters\build\k8s-test-redhat\ansible\roles\image_registry\meta\main.yml`)

PR #696, opened 2019-11-08
Work on upgrade

seriva:
- Fix for #694.
- Documentation.
- Bumped versions and added change log entry.
- Added check --offline-requirements is an actual valid path.


seriva:
- Fix for #694.
- Documentation.
- Bumped versions and added change log entry.
- Added check --offline-requirements is an actual valid path.

PR #729, opened 2019-11-18
IP change on existing Epiphany cluster

toszo:


tools/ip_change/kubernetes/start.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
erzetpe: Maybe we can use some more descriptive name for this script?

Issue #731, opened 2019-11-19
[upgrade] kube-scheduler pod crashing after upgrading k8s master to 1.14


przemyslavic:
kube-scheduler pod crashing after upgrading k8s master to 1.14
port 10251 already in use by another kube-scheduler process
>[operations@vm-qa-upgrades-ubu-master-001 ~]$ kubectl logs kube-scheduler-vm-qa-upgrades-ubu-master-001 -n=kube-system
I1118 22:45:02.705911       1 serving.go:319] Generated self-signed cert in-memory
failed to create listener: failed to listen on 0.0.0.0:10251: listen tcp 0.0.0.0:10251: bind: address already in use

PR #735, opened 2019-11-19
Fix for 'port already in use by another kube-scheduler process'

to-bar:
Stop kubelet service before updating Docker (#731)


to-bar:
Fixes #731

Issue #869, opened 2020-01-28
Issue installing Debian packages


samiralajmovic:
Several packages are no installed when you attempt to install packages from the requirements.txt file.

```
08:10:19 INFO cli.engine.ansible.AnsibleCommand - TASK [common : Install debian family packages] **************************************************************************
08:10:19 INFO cli.engine.ansible.AnsibleCommand - fatal: [avt]: FAILED! => {"cache_update_time": 1579507811, "cache_updated": false, "changed": false, "msg": "'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\"      install 'netcat' 'curl' 'tmux' 'fping' 'iftop' 'htop' 'sshpass' 'vim' 'sysstat' 'python-setuptools' 'a
pt-transport-https' 'software-properties-common' 'ebtables' 'ethtool' 'jq'' failed: E: Unable to correct problems, you have held broken packages.\n", "rc": 100, "stderr": "E: Unable to correct problems, you have held broken packages.\n", "stderr_lines": ["E: Unable to correct problems, you have held broken packages."], "stdout": "Reading package lists...\nBuildin
g dependency tree...\nReading state information...\nSome packages could not be installed. This may mean that you have\nrequested an impossible situation or if you are using the unstable\ndistribution that some required packages have not yet been created\nor been moved out of Incoming.\nThe following information may help to resolve the situation:\n\nThe following
packages have unmet dependencies:\n curl : Depends: libcurl4 (= 7.58.0-2ubuntu3.8) but it is not going to be installed\n vim : Depends: libpython3.6 (>= 3.6.5) but it is not going to be installed\n", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information...", "Some packages could not be installed. This may mean that
 you have", "requested an impossible situation or if you are using the unstable", "distribution that some required packages have not yet been created", "or been moved out of Incoming.", "The following information may help to resolve the situation:", "", "The following packages have unmet dependencies:", " curl : Depends: libcurl4 (= 7.58.0-2ubuntu3.8) but it is n
ot going to be installed", " vim : Depends: libpython3.6 (>= 3.6.5) but it is not going to be installed"]}
```

I found the function `deplist_cmd` in `data/common/ansible/playbooks/roles/repository/files/download-requirements/ubuntu-18.04/download-requirements.sh` which is responsible for fetching dependencies. 

```sh
[...]

PR #871, opened 2020-01-28
Fix the missing dependency repository issue for ubuntu

samiralajmovic:
Perhaps there was a reason not to include `--recurse` flag, but it fixes the issue for me.

fixes https://github.com/epiphany-platform/epiphany/issues/869


to-bar:
See https://github.com/epiphany-platform/epiphany/issues/869#issuecomment-585201329

PR #917, opened 2020-02-14
HA Kubernetes proposal

sk4zuzu:


toszo:
One thing missing here is the actual decision - which way we are going to implement this feature but I assume it will be based on the discussion.

PR #941, opened 2020-02-26
Health Monitor design doc

toszo:
Fixed incorrect line ending in config-pre.sh (was failing on Mac)


core/src/epicli/.devcontainer/cert/config-pre.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+
```
seriva: Not some crappy Windows EOL?

PR #961, opened 2020-03-03
Add nfs role in epicli

ks4225:
Hi Epiphany Team,

This pull request is an attempt at creating NFS VMs in the Epiphany (epicli) process.

We intend to use Epiphany in building clusters that will host Ellipse.
The motivation for this feature is that in order to deploy Ellipse, NFS is currently a critical piece.
Till now we have configured NFS external to Epiphany, but it makes sense (at least for us) for Epiphany to handle this too.

We are also aware of other options for persistent volumes like Azure Files, but they have historically caused issues in production for Ellipse. NFS is a solution that has been working consistently for us, and why that is the option we have gone for here.

The changes are currently specific to handling the scenario we are using (an Azure based cluster).
It allows for simple configuration of the NFS export definitions.

We would appreciate feedback on how we can progress this forward and eventually merge into Epiphany.

Thanks,
DI Team


core/src/epicli/data/common/defaults/epiphany-cluster.yml
```diff
@@ -1,94 +1,101 @@
...
+          address_pool: 10.1.9.0/24
+    opendistro_for_elasticsearch:
+      count: 0
+      machine: logging-machine
+      configuration: default
+      subnets:
+        - availability_zone: eu-west-2a
+          address_pool: 10.1.10.0/24
+    nfs:
+      count: 1
```
seriva: I think this is something we probably want to set to 0 for defaults since this cluster component is more of an edge case at this point.

toszo:
@seriva please create feature branch for NFS feature - for now, it will not be merged to develop because needs adjustments.

seriva:
@ks4225 I created a feature branch for this (feature/nfs-storage) and you can do a  pull request on that branch. If you need any help/advice with the comments from above let us know.

PR #997, opened 2020-03-12
Add nfs role in epicli

ks4225:
Originally created here: https://github.com/epiphany-platform/epiphany/pull/961

I believe I have addressed the points raised. Please let me know if there are still some gaps in the implementation.

Thanks,
DI team


core/src/epicli/data/common/defaults/configuration/nfs.yml
```diff
@@ -0,0 +1,12 @@
+kind: configuration/nfs
+title: "NFS Node Config"
+name: default
+specification:
+  nfs_defs:
+    - nfs_path: "/srv/nfs4/exports/"
+      nfs_src: "/exports"
+      nfs_export: "*(rw,sync,no_subtree_check,no_root_squash)"
+  # nfs_defs is a list of NFS definitions to be created
```
seriva: Can you put the comments behind the actual lines pls? That way when using epicli init they will be included for sure.

seriva:
You dont need to add the nfs entry in the minimal-cluster-config.yml`s. nfs is more of an edge case and the entry will still by added when people run epicli init with --all.

core/src/epicli/data/common/defaults/configuration/nfs.yml
ks4225: I have moved the comments now

ks4225:
I have applied the recommended changes.

ks4225:
Are you able to give us some insight into what will happen from now; in the feature branch and how the changes will eventually be released?

Is there anything additional you will require from our team?

Thanks,
DI team

PR #1057, opened 2020-03-24
Add k8s pgbouncer to epiphany

plirglo:
- added liveness and readiness basic checks
- added pgbouncer init container
- pgbouncer gets user credentials from database using init container


core/src/epicli/data/common/defaults/configuration/applications.yml
```diff
@@ -66,11 +66,11 @@ specification:
 ## --- Pgpool ---
 
   - name: pgpool # this service requires PostgreSQL to be installed in cluster
-    enabled: no
+    enabled: yes
```
seriva: Just wondering of by default we dont want to disable this like the rest of the application deployments? What if someone does not install postgress?

plirglo: Fixed, thx

PR #1065, opened 2020-03-25
Migration from PostgreSQL installed from Software Collections to installed from PostgreSQL official repository

erzetpe:


docs/home/howto/DATABASES.md
```diff
@@ -1,3 +1,70 @@
...
+rpm -qa | grep rh-postgresql10-postgresql-server
+```
+
+1). Stop instance of PostgreSQL installed from Software Collections database with command:
+
+```bash
+systemctl stop postgresql
+```
+
+2). Prepare backup of your PostgreSQL data directory -  - e.g. with tar command:
```
plirglo: directory -  - e.g. should be directory - e.g. (?)

docs/home/howto/DATABASES.md
```diff
@@ -1,3 +1,70 @@
...
+
+```bash
+tar -cf backup.tar /var/opt/rh/rh-postgresql10/lib/pgsql/data/
+```
+
+But any other tool that will provide you reliable backup that will prevent you from data loose can be chosen.
```
plirglo: loss instead loose

seriva: Maybe he has data on the loose?

PR #1091, opened 2020-03-30
Documentation - postgres migration replication new structure

erzetpe:


docs/home/howto/DATABASES.md
```diff
@@ -94,7 +99,53 @@ specification:
...
+    ...
+    replication:
+      enabled: true
+      replication_user_name: your_postgresql_replication_user
+      replication_user_password: your_postgresql_replication_password
+      specification.extensions.replication.replication_user_password
+      use_repmgr: false
+    ...
+title: PostgreSQL
+version: 0.6.0
```
seriva: I would just leave versions of yml snippets absent.

docs/home/howto/DATABASES.md
```diff
@@ -94,7 +99,53 @@ specification:
...
+```
+
+If `enabled` is set to `yes` in `replication`, then Epiphany will automatically create cluster of primary and secondary server
```
plirglo: Please add something about two database nodes

PR #1099, opened 2020-03-31
Use resources from config file

pprach:


plirglo:
I don't see any changes in applications' configuration. I think if we remove these parameters from deployment.yaml we need to move them to configurations (as you did reference), please see how thi is done for pgpool:
https://github.com/epiphany-platform/epiphany/blob/feature/postgres-ha/core/src/epicli/data/common/defaults/configuration/applications.yml

PR #1142, opened 2020-04-08
Use backup of yum.repos.d to identify added repositories

### LINKED ISSUE #1135: 2ndquadrant yum repos remain enabled on repository host after teardown
```
# cat /var/tmp/enabled-system-repos.txt
2ndquadrant-dl-default-release-pg10
2ndquadrant-dl-default-release-pg10-debug
rhui-REGION-client-config-server-7
rhui-REGION-rhel-server-releases
rhui-REGION-rhel-server-rh-common
```

to-bar:
Fixes #1135

### DIFF
--- a/CHANGELOG-0.6.md
+++ b/CHANGELOG-0.6.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - [#1154](https://github.com/epiphany-platform/epiphany/issues/1154) - Node exporter is not installed on logging vms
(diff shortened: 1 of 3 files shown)


seriva:
Can you add entry to changelog? On develop there should be an entry for 061 already.

to-bar:
> 
> 
> Can you add entry to changelog? On develop there should be an entry for 061 already.

Good point, added.

PR #1217, opened 2020-04-28
reversing unneeded delegate_to privilege elevation (fix)

sk4zuzu:


seriva:
pls add entry to changelog for this fix under 0.6.1

PR #1236, opened 2020-05-05
Enable Ansible profile_tasks plugin for timing tasks

### LINKED ISSUE #811: Measure execution time of Ansible tasks
There is a built-in plugin which allows to measure execution time of each task and shows summary for each role sorted descending:
https://docs.ansible.com/ansible/latest/plugins/callback/profile_tasks.html
http://techadminblog.com/get-execution-time-of-each-task-ansible/

It allows to easily identify long running tasks as candidates for optimization.

to-bar:
Resolves #811

### DIFF
--- a/core/src/epicli/cli/engine/ApplyEngine.py
+++ b/core/src/epicli/cli/engine/ApplyEngine.py
@@ -20,8 +20,9 @@ from cli.engine.ansible.AnsibleRunner import AnsibleRunner
 class ApplyEngine(Step):
     def __init__(self, input_data):
         self.file = input_data.file
-        self.skip_infrastructure = input_data.no_infra if hasattr(input_data, 'no_infra') else False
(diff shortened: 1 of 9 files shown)


core/src/epicli/cli/engine/ApplyEngine.py
```diff
@@ -20,8 +20,9 @@
 class ApplyEngine(Step):
     def __init__(self, input_data):
         self.file = input_data.file
-        self.skip_infrastructure = input_data.no_infra if hasattr(input_data, 'no_infra') else False
-        self.skip_config = input_data.skip_config if hasattr(input_data, 'skip_config') else False
+        self.skip_infrastructure = getattr(input_data, 'no_infra', False)
```
seriva, line 23: Sweet, if there are other places you can add this feel free:)

PR #1237, opened 2020-05-07
Existing network support in Azure

ks4225:
These changes only apply to Azure Cloud clusters.

When creating an Epiphany Cluster, we may want to integrate the cluster with existing network resources. The changes here allow the selection of an existing network (using resource group + vnet + subnet) so that the cluster is terraformed into that existing network. Vnets and subnets will not be created when using this option.


sunshine69:
Can I have some added into the README Howto section to illustrate/show how to use the new feature? 

Does it include the generation of initial manifest with option `--full`? If not I suggest to include it or at least an skeleton example.

ks4225:
I have added a readme section in `CLUSTER.md` (which includes an example). Since this feature will probably be an edge case, I haven't added it into the YAML generation.

ks4225:
When re-running this, I noticed the Azure network validation occurs before the Azure login logic.
As a workaround, I have duplicated the required login logic, but happy to follow any suggestion on how to fix it properly.

docs/home/howto/CLUSTER.md
```diff
@@ -291,6 +291,21 @@ To setup the cluster do the following steps from the provisioning machine:
...
+An existing network can be specified in the cluster's YAML file as follows:
+
+```yaml
+network:
+      existing_rg: resource-group-name
+      existing_vnet: vnet-name
+      existing_subnet: subnet-name
+```
+
+If these values are specific, the `use_network_security_groups` will be ignored (i.e. effectively always be false).
```
mkyc, line 305: @ks4225 sorry for late response. Did you mean `specified` instead of `specific`? I'm trying to understand intent of that change first.

ks4225: Yes, sorry, typo there. It should be "specified".

core/src/epicli/cli/engine/ApplyEngine.py
```diff
@@ -161,6 +169,45 @@ def dry_run(self):
 
         return [*self.configuration_docs, *self.infrastructure_docs]
 
+    def azure_subnet_query_if_required(self):
```
seriva, line 172: This needs to be refactored for sure since its mostly code duplication and on this level we should not interact with the provider correctly. Thats what we build the abstraction layer for.

docs/home/howto/CLUSTER.md
mkyc: so that should be fixed.

ks4225:
Thanks @seriva, I understand what you are asking and I agree. Just haven't had time to revisit this PR for a while as it's not a priority for us at the moment.

rafzei:
Im closing this after approval from @ks4225.

PR #1242, opened 2020-05-08
Add git-lfs package to VS Code dev container

to-bar:


to-bar:
> 
> 
> Wouldn't this be better to just disable git lfs in our repository? We are not storing the binary files, etc. and sometimes lfs in the past was problematic.

We store svg and png files:
```
# find . -type f \( -iname \*.png -o -iname \*.svg \) | wc -l
34
```
Using LFS we save some space (not much so far):
```
# find . -type f \( -iname \*.png -o -iname \*.svg \) -exec du -ch {} + | grep total$
136K    total
```
After conversion from LFS pointers to files:
```
# find . -type f \( -iname \*.png -o -iname \*.svg \) -exec du -ch {} + | grep total$
1.9M    total
```

The drawback of LFS is that it requires additional package (git-lfs). Thus, I created this PR to reduce this drawback :)

PR #1276, opened 2020-05-18
Additional settings for epirepo teardown

ar3ndt:


core/src/epicli/data/common/defaults/configuration/repository.yml
```diff
@@ -3,4 +3,11 @@ title: "Epiphany requirements repository"
...
-  download_done_flag_expire_minutes: 120
\ No newline at end of file
+  download_done_flag_expire_minutes: 120
+  teardown:
+    disable_http_server: true
+    remove:
+      files: true
+      helm_charts: true
+      images: true
+      packages: true
```
to-bar, line 13:
```suggestion
      files: false
      helm_charts: false
      images: false
      packages: false
```
I would vote for leaving epirepo files, they can be reused for offline installation.

ar3ndt: ok

PR #1290, opened 2020-05-21
Install Helm only on 1st kubernetes_master host

ar3ndt:


core/src/epicli/data/common/ansible/playbooks/roles/repository/tasks/Debian/setup.yml
```diff
@@ -20,13 +20,15 @@
...
-  include_tasks: "create-helm-repo.yml"
+- name: Helm binary installation and Helm repository creation
+  block:
+    - name: Include Helm binary installation 
+      include_role:
+        name: helm
+        tasks_from: install-bin
+    - name: Include Helm repository creation
+      include_tasks: "create-helm-repo.yml"
+  when: inventory_hostname in groups['kubernetes_master'][0]
```
sk4zuzu, line 31: Please consider using the `target_repository_hostnames[0]` defined here https://github.com/epiphany-platform/epiphany/blob/cc57b54f4e3694aa8a6b55217bf8a72a9191d726/core/src/epicli/data/common/ansible/playbooks/roles/repository/tasks/gather-facts.yml
The cause for this is currently we have repository on single host (for performance reasons) and this host can theoretically switch places inside the kubernetes_master group in inventory (for example in AWS while adding more master nodes to kubernetes cluster) with other nodes.

core/src/epicli/data/common/ansible/playbooks/roles/repository/tasks/RedHat/setup.yml
```diff
@@ -23,13 +23,15 @@
...
-  include_tasks: "create-helm-repo.yml"
+- name: Helm binary installation and Helm repository creation
+  block:
+    - name: Include Helm binary installation 
+      include_role:
+        name: helm
+        tasks_from: install-bin
+    - name: Include Helm repository creation
+      include_tasks: "create-helm-repo.yml"
+  when: inventory_hostname in groups['kubernetes_master'][0]
```
sk4zuzu, line 34: Same as for Debian.
Also this code is not specific to either Debian or Redhat or in fact any distro since helm binary is statically linked. I'd suggest putting that code in a single place outside of Redhat or Debian specific code.

core/src/epicli/data/common/ansible/playbooks/roles/repository/tasks/setup.yml
```diff
@@ -76,3 +76,7 @@
 
 - name: Set up repositories
   include_tasks: "{{ ansible_os_family }}/setup.yml"
+
+- name: Include Helm repository creation
+  include_tasks: "create-helm-repo.yml"
+  when: inventory_hostname in target_repository_hostnames[0]
```
sk4zuzu, line 82: This is a "string" match, because `target_repository_hostnames[0]` is a string (not a list), is this intended? I think using `==` here (a strict match) would be the correct approach. :+1:

core/src/epicli/data/common/ansible/playbooks/roles/repository/tasks/setup.yml
```diff
@@ -76,3 +76,7 @@
 
 - name: Set up repositories
   include_tasks: "{{ ansible_os_family }}/setup.yml"
+
+- name: Include Helm repository creation
+  include_tasks: "create-helm-repo.yml"
+  when: inventory_hostname == target_repository_hostnames[0]
```
to-bar, line 82:
```suggestion
  when: inventory_hostname in target_repository_hostnames
```
How about keeping the same condition for consistency?
```
- name: Produce final list of hosts to deploy repositories on
  set_fact:
    target_repository_hostnames: >-
      {{ groups.repository | difference(groups.kubernetes_master) | union([resolved_repository_hostname]) | unique }}
```

PR #1311, opened 2020-06-01
[DOC] Epicli UX design document inserted.

mkyc:


docs/design-docs/cli/cli-ux.md
```diff
@@ -0,0 +1,378 @@
...
+to indicate processing or some not important for this document output. 
+
+## Story
+
+### epicli installation
+
+To increase user base we need to provide brew formulae to allow simple installation. 
+```
+> brew install epicli
+```
```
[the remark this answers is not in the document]
mkyc: Good point about dependencies. However, this design does not consider technical implementation. I used "brew" to highlight importance of smooth user experience. It's super nice when all you need to do before you use tool is `brew install tool-you-want` (ex.: https://learn.hashicorp.com/terraform/getting-started/install.html#install-terraform). I'm okay with situation when we in fact run `docker run something` inside `epicli something`.

docs/design-docs/cli/cli-ux.md
```diff
@@ -0,0 +1,378 @@
...
+
+It could also be `epicli init -p aws -t nonprod -c postgresql ....` or `epicli --no-interactive -p aws` for non-interactive run. 
+
+#### inspect .epicli/
+
+Previous command generated files in ~/.epicli directory. 
+```
+> ls –la ~/.epicli
+config
+environemts/
```
erzetpe, line 64:
```suggestion
environmets/
```

docs/design-docs/cli/cli-ux.md
```diff
@@ -0,0 +1,378 @@
...
+To increase user base we need to provide brew formulae to allow simple installation. 
+```
+> brew install epicli
+``` 
+ 
+### environment initialization and deployment
+
+#### init
+
+As before user should be able to start interaction with epicli with `epicli init` command. In case of no parameters interactive version would be opened.
```
erzetpe, line 40: Jhipster has really nice interactive mode. We can try to achieve something similar to their implementation.

docs/design-docs/cli/cli-ux.md
```diff
@@ -0,0 +1,378 @@
...
+How many Kubernetes Nodes do you want?: 2
+Do you want PostgreSQL relational database?: Yes
+Do you want RabbitMQ message broker?: No
+Name your new Epiphany environment: test1
+There is already environment called test1, please provide another name: test2
+[...]
+Your new environment configuration was generated! Go ahead and type: 'epicli status' or 'epicli apply. 
+```
+
+It could also be `epicli init -p aws -t nonprod -c postgresql ....` or `epicli --no-interactive -p aws` for non-interactive run.
```
erzetpe, line 56: Looking at the number of parameters and possible options I think that this will be hard to handle.

docs/design-docs/cli/cli-ux.md
```diff
@@ -0,0 +1,378 @@
...
+You are not connected to your environment. Please type 'epicli init cloud' to provide authorization informations!    
+```
+
+As output is saying for now this command only uses local files in ~/.epicli directory. 
+
+#### init cloud
+
+Follow instructions to provide cloud provider authentication.  
+```
+> epicli init cloud
```
erzetpe, line 126: Here also some option to select provider would be good.

docs/design-docs/cli/cli-ux.md
```diff
@@ -0,0 +1,378 @@
...
+> epicli apply
+[...]
+---
+Environment “test2” was updated! Plese type “epicli status” to see status or “epicli components” to list components. To login to kubernetes cluster as root please type “epicli components kubernetes login”. 
+Command “epicli status” will synchronize every time now, so no need to run “epicli status update”
+```
+ 
+## Daemon
+
+We should also consider scenario with web browser management tool. It might look like:
```
erzetpe, line 367: Yup, we have planned this some time ago. :)

erzetpe: Quite long time ago. :)

docs/design-docs/cli/cli-ux.md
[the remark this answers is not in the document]
toszo: I think we can say that dependencies are "implementation detail" (it can be tackled by using docker image underneath - just an idea to show there will be solutions for that).

docs/design-docs/cli/cli-ux.md
mkyc: Yes it might. It should be investigated later.

Issue #1314, opened 2020-06-02
The VSCode devcontainer does not support custom CA certificate bundles


seriva:
Currently we can only include one custom CA when building the devcontainer. We need the ability to:

- Add multiple crt certs to /usr/local/share/ca-certificates/
- Add the ability to configure the full CA bundles for PIP and Boto3
- Original configuration for a single CA cert should be maintained

mkyc:
@seriva what about this task? Shouldn't it be closed?

seriva:
This was fixed already of part 0.7.0

PR #1315, opened 2020-06-02
Certificate bundle enhancement for VSCode devcontainer (#1314)

seriva:
- Added support for multiple crt certs for /usr/local/share/ca-certificates/
- Added the ability to configure the full CA bundles for PIP and Boto3
- Minor fix for Azure API proxy for SP login where PW contains ' character causing an string escape error
- Added debug flag when running Ansible task
- Updated documentation

### DIFF
--- a/docs/home/howto/PREREQUISITES.md
+++ b/docs/home/howto/PREREQUISITES.md
@@ -60,45 +60,6 @@ docker run -it -v LOCAL_DIR:/shared --rm epiphanyplatform/epicli:TAG
 
 Where `LOCAL_DIR` should be replaced with the local path to the directory for Epicli input (SSH keys, data yamls) and output (logs, build states).
 
-## Run Epicli directly from OS
-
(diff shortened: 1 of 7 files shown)


docs/home/howto/PREREQUISITES.md
```diff
@@ -60,45 +60,6 @@ docker run -it -v LOCAL_DIR:/shared --rm epiphanyplatform/epicli:TAG
 
 Where `LOCAL_DIR` should be replaced with the local path to the directory for Epicli input (SSH keys, data yamls) and output (logs, build states).
 
-## Run Epicli directly from OS
```
mkyc, line 63: Why that section got deleted?

seriva: Currently not all needed dependencies are Python based anymore. For running test for example we require Rust + additional dependencies that need to be installed. So for now the container approach is the only one we want to advertise.

mkyc: ok

PR #1342, opened 2020-06-07
[DOC] Epiphany C4 model

mkyc:


pyrkamarcin:
@mkyc Files like this: docs/architecture/current-c4/4_Epiphany_Component_Diagram_metrics_and_logs.png are not easy readable - arrows are too complicated and aren't understandable.

mkyc:
@pyrkamarcin I updated layout of those images to be generated little better. I see that there is still problem with one showing human interactions but as it is generated I do not want to spend too much time on it.

PR #1356, opened 2020-06-16
Modify backup / restore roles according to test results

plirglo:


core/src/epicli/data/common/defaults/configuration/backup.yml
```diff
@@ -1,10 +1,9 @@
 kind: configuration/backup
 title: Backup Config
 name: default
+provider: any
```
sk4zuzu, line 4: I don't think that we can put this in default config. :(
```suggestion
```

core/src/epicli/data/common/defaults/configuration/recovery.yml
```diff
@@ -1,11 +1,9 @@
 kind: configuration/recovery
 title: Recovery Config
 name: default
+provider: any
```
sk4zuzu, line 4: I don't think that we can put this in default config. :(
```suggestion
```

Issue #1349, opened 2020-06-10
Disable helm installation when deploying a non-K8s cluster


przemyslavic:
Adjusting the feature-mapping configuration I had to add the helm role to the repository host to be able to deploy a non-k8s cluster. Otherwise the installation fails on the `helm: Unpack Helm binary` task.

```
2020-06-09T16:08:14.2331424Z 16:08:14 INFO cli.engine.ansible.AnsibleCommand - TASK [helm : Unpack Helm binary] ***********************************************
2020-06-09T16:08:14.2507270Z 16:08:14 INFO cli.engine.ansible.AnsibleCommand - fatal: [ci-fixazurrhelflannel-load-balancer-vm-0]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'dict object' has no attribute 'version'\n\nThe error appears to be in '/shared/build/fixazurrhelflannel/ansible/roles/helm/tasks/install-bin.yml': line 2, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n---\n- name: Unpack Helm binary\n  ^ here\n"}
```
There is no reason to install helm if we don't have kubernetes.

mkyc:
@sk4zuzu this bug should go to "DoD Check"

sk4zuzu:
Couple of issues fixed here:

1. The `helm` inventory group has been reused, that allowed skipping all role tasks in ansible, when the `helm` role is absent from all features inside the `feature-mapping`.

2. The fix from 1. above did not completely solve the problem, there was an issue with installation of helm binary. Despite the disabling of the `helm` role, the helm binary is still being installed in the `repository` role. The `repository` role borrows code that installs helm binary and originally version string of the binary was defined in role vars of the `helm` role, which in turn were not available when `helm` role was disabled. This has been corrected.

3. Problem stated in 2. above is more of a general kind. We usually put some package in repository and define some version in config defaults (those things must match so there is no mistake). So the `download` role has been extended with the method to get the latest file from the repository and make definition of the artifact version in config defaults redundant (this code can be reused not only from the `helm` role).

4. Execution of the role has been narrowed to minimal number of hosts in the cluster.

5. Extra ability to install helm chart without custom `values.yaml` has been added.

6. Some minor refactor and cleanups like removing trailing witespace and restructuring code to improve readability has been done as well.

Here's an example how to disable helm and kubernetes and still deploy Epiphany:

```yaml
[...]

PR #1334, opened 2020-06-03
Fix multi master Helm conditions

ar3ndt:

PR #1360, opened 2020-06-16
Hotfix: removing hard dependency on Helm

sk4zuzu:


mkyc:
Does it solve some issue? Should it be linked?

sk4zuzu:
Yes it solves the issue. It was a mutli level problem really (including other not reported cases), so this PR contains more code. :+1:

mkyc:
I mean which issue. Should we link it to one of issues?

mkyc:
Yeah, sorry I didn't noticed links.

sk4zuzu:
The "official" issue it solves is linked to this PR #1349, but also the other #1334 becomes redundant I guess.

Issue #1373, opened 2020-06-19
[BUG] permission denied for shared directory in the container when no volume was mounted


przemyslavic:
**Describe the bug**
Assuming I want to run everything in the container without mounting any volume, there is an issue with permissions.
The shared directory is owned by root and therefore epicli can not write anything in it.

**To Reproduce**
1. execute `docker run -it image:tag`
3. execute `epicli init -p aws -n xxx`

**Expected behavior**
Configuration file has been generated and saved in the shared directory.

**Additional context**

Command output
```
epiuser@60d68dc3d233:/shared$ epicli init -p aws -n xxx
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/cli/epicli.py", line 126, in main
    return args.func(args)
  File "/usr/local/lib/python3.7/site-packages/cli/epicli.py", line 147, in run_init
    with InitEngine(args) as engine:
  File "/usr/local/lib/python3.7/site-packages/cli/engine/InitEngine.py", line 16, in __init__
    super().__init__(__name__)
  File "/usr/local/lib/python3.7/site-packages/cli/helpers/Step.py", line 8, in __init__
    self.logger = Log(step_name)
  File "/usr/local/lib/python3.7/site-packages/cli/helpers/Log.py", line 33, in __new__
    Log.instance = Log.__LogBase()
  File "/usr/local/lib/python3.7/site-packages/cli/helpers/Log.py", line 17, in __init__
    log_path = os.path.join(get_output_path(), config.log_file)
  File "/usr/local/lib/python3.7/site-packages/cli/helpers/build_saver.py", line 76, in get_output_path
    os.makedirs(Config().output_dir)
  File "/usr/local/lib/python3.7/os.py", line 223, in makedirs
    mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: '/shared/build/'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/epicli", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.7/site-packages/cli/epicli.py", line 128, in main
    logger = Log('epicli')
  File "/usr/local/lib/python3.7/site-packages/cli/helpers/Log.py", line 33, in __new__
    Log.instance = Log.__LogBase()
  File "/usr/local/lib/python3.7/site-packages/cli/helpers/Log.py", line 17, in __init__
    log_path = os.path.join(get_output_path(), config.log_file)
[...]

PR #1378, opened 2020-06-23
Docker image user permissions fix (#1373) 

rafzei:
+ got rid of empty continuation line warning

### DIFF
--- a/core/src/epicli/Dockerfile
+++ b/core/src/epicli/Dockerfile
@@ -1,13 +1,11 @@
 FROM python:3.7-slim
 
 ARG EPICLI_VERSION
-
 ARG USERNAME=epiuser
 ARG USER_UID=1000
 ARG USER_GID=$USER_UID
-ARG SHARED_DIR=/shared
 
-ENV DOCKER_CLI Yes
+ENV EPICLI_DOCKER_SHARED_DIR=/shared
 
 COPY /dist/ /epicli
 WORKDIR /epicli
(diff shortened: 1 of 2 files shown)


core/src/epicli/Dockerfile
```diff
@@ -5,6 +5,7 @@ ARG EPICLI_VERSION
 ARG USERNAME=epiuser
 ARG USER_UID=1000
 ARG USER_GID=$USER_UID
+ARG SHARED_DIR=/shared
```
seriva, line 8: The the dockerized Epicli expects the /shared director as output. Check /epiphany/core/src/epicli/cli/helpers/Config.py.

My suggestion would be to replace"

`ENV DOCKER_CLI Yes` 

With:

ENV EPICLI_DOCKER_SHARED_DIR $SHARED_DIR

And change the logic of /epiphany/core/src/epicli/cli/helpers/Config.py to reflect this. When people build there own container with a changed SHARED_DIR argument it will always be reflected properly in the Epicli config when dockerized and avoid issues. Also it makes more sence then `ENV DOCKER_CLI Yes`

core/src/epicli/cli/helpers/Config.py
```diff
@@ -4,11 +4,11 @@
 class Config:
     class __ConfigBase:
         def __init__(self):
-            self._docker_cli = bool(os.environ.get('DOCKER_CLI', ''))
+            self._docker_cli = bool(os.environ.get('EPICLI_DOCKER_SHARED_DIR',''))
 
             self._output_dir = None
             if self._docker_cli:
-                self._output_dir = '/shared/build/'
+                self._output_dir = os.environ.get('EPICLI_DOCKER_SHARED_DIR')
```
seriva, line 11: please add the /build dir to the path.

PR #1379, opened 2020-06-23
Feature/restore and backup validation

pyrkamarcin:
Adding validation for recovery manifest to prevent accepting of Kubernetes config object.


core/src/epicli/data/common/validation/configuration/recovery.yml
```diff
@@ -17,3 +17,75 @@ properties:
...
+            - enabled
+            - snapshot_name
+            additionalProperties: false
+            properties:
+              enabled:
+                "$id": "#/properties/specification/properties/components/properties/load_balancer/properties/enabled"
+                type: boolean
+              snapshot_name:
+                "$id": "#/properties/specification/properties/components/properties/load_balancer/properties/snapshot_name"
+                type: string
```
sk4zuzu, line 35: Would be amazing if we'd get validation for it (and all the others)! It can be either `latest` or `20200526-102034` (just digits and single '-'). :+1: What do you think?

sk4zuzu:
@pyrkamarcin good work! Feel free to add similar validation for backup as well. :+1:

core/src/epicli/data/common/validation/configuration/recovery.yml
pyrkamarcin: Great idea. It's done now in 90e478f.

core/src/epicli/data/common/validation/configuration/backup.yml
```diff
@@ -17,3 +17,66 @@ properties:
...
+            properties:
+              enabled:
+                "$id": "#/properties/specification/properties/components/properties/load_balancer/properties/enabled"
+                type: boolean
+          logging:
+            "$id": "#/properties/specification/properties/components/properties/logging"
+            type: object
+            required:
+            - enabled
+            additionalProperties: true
```
sk4zuzu, line 37: Looks like an overlook.
```suggestion
            additionalProperties: false
```

pyrkamarcin: Fix in 84c68b8

Issue #1264, opened 2020-05-13
PGAudit is missing shared preload library in the default configuration


przemyslavic:
**Describe the bug**
The missing configuration must be added to defaults. Otherwise pgaudit will not work.

```
  extensions:
    pgaudit:
      enabled: yes
      shared_preload_libraries:
        - pgaudit
```

**How to reproduce**
Execute `epicli apply ...` without configuration mentioned above.

**Expected behavior**
Shared preload libraries should be added by default when pgaudit is installed.

**Config files**
As above.

**Environment**
- Cloud provider: [All]
- OS: [All]

**epicli version**: [`epicli --version`]

**Additional context**
None

---

**DoD checklist**

* [ ] Changelog updated (if affected version was released)
* [ ] COMPONENTS.md updated / doesn't need to be updated
* [ ] Automated tests passed (QA pipelines)
  * [ ] apply
  * [ ] upgrade
* [ ] Case covered by automated test (if possible)
* [ ] Idempotency tested
* [ ] Documentation updated / doesn't need to be updated
* [ ] All conversations in PR resolved
* [ ] Backport tasks created / doesn't need to be backported

erzetpe:
This task is already done in [PR 1383](https://github.com/epiphany-platform/epiphany/pull/1383).

PR #1383, opened 2020-06-24
Fix for issue #1264

seriva:
Fix for issue #1264


core/src/epicli/data/common/defaults/configuration/postgresql.yml
```diff
@@ -62,7 +62,9 @@ specification:
                 when: replication
   extensions:
     pgaudit:
-      enabled: no
+      enabled: yes
```
to-bar, line 65:
```suggestion
      enabled: no
```
Any reason to enable additional logging by default? This impacts performance.

Issue #1191, opened 2020-04-23
Automatic Hashicorp Vault Agent installation, configuration and Kubernetes - Hashicorp Vault integration


erzetpe:
This is integration of Hashicorp Vault with Kubernetes according to design doc: This is installation of Hashicorp Vault, to provide Vault according to design doc: https://github.com/epiphany-platform/epiphany/blob/develop/docs/design-docs/k8s-vault-integration/k8s-vault-integration.md

Issue #1190, opened 2020-04-23
Automatic Hashicorp Vault installation and configuration


erzetpe:
This is installation of Hashicorp Vault, to provide Vault according to design doc: https://github.com/epiphany-platform/epiphany/blob/develop/docs/design-docs/k8s-vault-integration/k8s-vault-integration.md

PR #1393, opened 2020-06-29
Feature/vault documentation

erzetpe:


erzetpe:
This is documentation for issues: #1191, #1190.

docs/home/howto/SECURITY.md
```diff
@@ -367,3 +367,257 @@ cp /etc/kubernetes/admin.conf $HOME/
...
+```yaml
+kind: configuration/vault
+title: Vault Config
+name: default
+specification:
+  vault_enabled: true # enable Vault install
+  vault_system_user: vault # user name under which Vault service will be running
+  vault_system_group: vault # group name under which Vault service will be running
+  enable_vault_audit_logs: false # turn on audit logs that can be found at /opt/vault/logs/vault_audit.log
+  enable_vault_ui: false # enable Vault UI, shouldn't be used at production
```
mkyc, line 394: why?

docs/home/howto/SECURITY.md
```diff
@@ -367,3 +367,257 @@ cp /etc/kubernetes/admin.conf $HOME/
...
+  vault_script_autounseal: true # enable automatic unseal vault at the start of the service, shouldn't be used at production
+  vault_script_autoconfiguration: true # enable automatic configuration of Hashicorp Vault
```
mkyc, line 396: I believe that should be explained more. What auto configuration is doing? What is being configured?

docs/home/howto/SECURITY.md
```diff
@@ -367,3 +367,257 @@ cp /etc/kubernetes/admin.conf $HOME/
...
+
+Epiphany besides two already included in vault policies (root and default) provides two additional predefined policies:
+
+- admin - policy granting administration privileges, have sudo permission on Vault system endpoints
+- provisioner - policy granting permissions to create user secrets, adding secrets, enable authentication methods, but
+  without access to Vault system endpoints
+
+### Manual unsealing of the Vault
+
+When Hashicorp Vault starts it starts in sealed mode. This mean that Vault data is encrypted and Vault needs to generate key
```
mkyc, line 443: I would rephrase it a bit. Something like: "By design Hashicorp Vault starts in sealed mode."

mkyc, line 443: `data is encrypted and Vault needs to generate key` ... isn't it bit incorrect? I think it should be: `data is encrypted and operator needs to provide unsealing key` or something similar.

docs/home/howto/SECURITY.md
```diff
@@ -367,3 +367,257 @@ cp /etc/kubernetes/admin.conf $HOME/
...
+to decrypt and access data.
+
+Vault can be unsealed manually using command:
+
+```bash
+vault operator unseal
+```
+
+and passing three unseal keys from /opt/vault/init.txt file. In the future number of keys will be defined from the level
```
mkyc, line 452: `In the future releases, number ...`

docs/home/howto/SECURITY.md
erzetpe: Additional surface of attack, so when disabled more secure.

PR #1411, opened 2020-07-02
Add support for azure availability-set

sunshine69:


sunshine69:
I intended to add some documentation into the howto sections. I just did a test and it runs kosher, the availability set is generated and kube nodes get the availability set.

But I would like to add some unit test and doco as well. Not knowing this create the PR probably I click the wrong button :smile: however please hold a bit after I add some more documentation in probably on Monday.

docs/home/howto/CLUSTER.md
```diff
@@ -649,3 +649,109 @@ You can read more [here](https://www.confluent.io/blog/how-choose-number-topics-
...
+
+```yaml
+  components:
+    kafka:
+      count: 0
+    kubernetes_master:
+      count: 1
+    kubernetes_node:
+# This line tell we generate the availability-set terraform template
+      availability_set: kube-node-availability-set
```
mkyc, line 694: @sunshine69 on which components did you test such configuration?

sunshine69: The current prodeastus and devdr10, devcluster one I did create for all components type and it works for all. Kubernetes master, nodes, kfafka, postgres etc, each of them has its own availability set.

docs/home/howto/CLUSTER.md
```diff
@@ -649,3 +649,109 @@ You can read more [here](https://www.confluent.io/blog/how-choose-number-topics-
...
+      count: 0
+title: Epiphany cluster Config
+---
+kind: infrastructure/virtual-machine
+name: kubernetes-node-machine
+provider: azure
+specification:
+  size: Standard_DS3_V2
+# This tell to template the vm terraform to insert the id of the availability-set
+  availability_set: kube-node-availability-set
```
seriva, line 745: Do we really need to define this on this level as well? We know which VM definition is connected to which component definition already, having this in 2 places will be confusing I think.

core/src/epicli/data/azure/validation/infrastructure/availability-set.yml
```diff
@@ -0,0 +1 @@
+$ref: '#/definitions/unvalidated_specification'
```
mkyc, line 1: newline

docs/home/howto/CLUSTER.md
mkyc: Ok, one more thing, isn't that line repetition with line 680 in `kind: infrastructure/virtual-machine` section?

docs/home/howto/CLUSTER.md
mkyc: oh, that would also be my question.

sunshine69: In theory we don't. Ideally we don't. This is repeated information - while we tell the type `epiphany-cluster` and it should have enough information to generate the terraform config.

However the fact is if you don't do like this you need to implement a way to retrieve the information at the time of generating the terraform config for the host to from the `epiphany-cluster` data which turn out non trivial tasks and increase complexity.

However I will have a look again and see what I can do.

docs/home/howto/CLUSTER.md
sunshine69: Not sure what you mean by repetition, as of now we need to define in several places like the sample config.

I am trying to see what needs to be done to remove the settings in the `kind: infrastructure/virtual-machine` without adding too much complexity.

core/src/epicli/data/azure/validation/infrastructure/availability-set.yml
sunshine69: do you mean that we need a newline? If so then do we need it? Most of other file in same dir does not have new line anyway.

IMHO yaml doe snot specify we need it, it is a practice to add `...` for closing or `---` in opening but really need for stream communication of yaml data anyway.

docs/home/howto/CLUSTER.md
seriva: You should be able to implement that pretty easily in the InfrastructureBuilder.py

PR #1429, opened 2020-07-08
Change move to copy build dir during an upgrade

rafzei:


core/src/epicli/cli/engine/UpgradeEngine.py
```diff
@@ -42,11 +42,8 @@ def backup_build(self):
         backup_dir_name = f'backup_{int(round(time.time() * 1000))}'
         backup_temp_dir = os.path.join(self.build_dir, '../', backup_dir_name)
         self.backup_build_dir = os.path.join(self.build_dir, backup_dir_name )
-        self.logger.info(f'Backing up old build files to "{self.backup_build_dir}"')
-        shutil.move(self.build_dir, backup_temp_dir, copy_function = shutil.copytree) 
-        if not os.path.exists(self.build_dir):
-            os.mkdir(self.build_dir)        
-        shutil.move(backup_temp_dir, self.backup_build_dir, copy_function = shutil.copytree)
+        self.logger.info(f'Backing up build dir to "{self.backup_build_dir}"')
+        shutil.copytree(self.build_dir, self.backup_build_dir)
```
seriva, line 46: Did you consider the old 0.2.3 backup folders? Is this compatible with this?

core/src/epicli/cli/engine/UpgradeEngine.py
```diff
@@ -42,11 +42,8 @@ def backup_build(self):
         backup_dir_name = f'backup_{int(round(time.time() * 1000))}'
         backup_temp_dir = os.path.join(self.build_dir, '../', backup_dir_name)
```
seriva, line 43: Is backup_temp_dir  still used?

core/src/epicli/cli/engine/UpgradeEngine.py
rafzei: We are not supporting 0.2 versions anymore.

core/src/epicli/cli/engine/UpgradeEngine.py
```diff
@@ -40,13 +40,9 @@ def backup_build(self):
...
         backup_dir_name = f'backup_{int(round(time.time() * 1000))}'
-        backup_temp_dir = os.path.join(self.build_dir, '../', backup_dir_name)
         self.backup_build_dir = os.path.join(self.build_dir, backup_dir_name )
-        self.logger.info(f'Backing up old build files to "{self.backup_build_dir}"')
-        shutil.move(self.build_dir, backup_temp_dir, copy_function = shutil.copytree) 
-        if not os.path.exists(self.build_dir):
-            os.mkdir(self.build_dir)        
-        shutil.move(backup_temp_dir, self.backup_build_dir, copy_function = shutil.copytree)
+        self.logger.info(f'Backing up build dir to "{self.backup_build_dir}"')
+        shutil.copytree(self.build_dir, self.backup_build_dir)
```
sk4zuzu, line 45: Please take a look at this part of Python 3.7 documentation (https://docs.python.org/3.7/library/shutil.html#shutil.copytree):
```
shutil.copytree(src, dst, symlinks=False, ignore=None, copy_function=copy2, ignore_dangling_symlinks=False)
Recursively copy an entire directory tree rooted at src, returning the destination directory. The destination directory, named by dst, must not already exist;
```
It clearly states that the `directory must not exist`. Of course the name is a timestamp basically, but personally I would be extra paranoid here (but that just maybe me idk, please ignore if you think it's not a valid suggestion) \\:D/.

Issue #1413, opened 2020-07-02
[FEATURE REQUEST] Set protocol for Vault only in one place in configuration


erzetpe:
**Is your feature request related to a problem? Please describe.**
We should be able to set protocol for Vault only in one place, not like in the config below:

```yaml
---
kind: configuration/vault
title: Vault Config
provider: aws
name: default
specification:
  vault_enabled: true
  ...
  **tls_disable: true
  externalVaultAddr: http://external-vault:8200**
```

This can lead to inconsistency and confusing bugs.

**Describe the solution you'd like**
I would like to have protocol - http or https configured only in one place.

**Describe alternatives you've considered**

**Additional context**
No additional context.

przemyslavic:
It has been tested. The protocol selection (http / https) now depends on the value of the tls_disable parameter and the vault address is built based on it.

Issue #1412, opened 2020-07-02
[BUG] Certificate in Vault is also generated or copied even if flag in configuration tls_disable is set to true


erzetpe:
**Describe the bug**
Certificate in Vault is also generated or copied even if flag in configuration tls_disable is set to true in Vault configuration.

**To Reproduce**
1. Create Vault config in Epiphany configuration, e.g:

```yaml
---
kind: configuration/vault
title: Vault Config
provider: aws
name: default
specification:
  vault_enabled: true
  tls_disable: true
```

2.  Execute `epicli apply ...` with your configuration file and settings from abov

**Expected behavior**
With flag tls_disable set to true part of code related to certificates should be skipped.

**Config files**
Shown above.

**OS (please complete the following information):**
 - OS: All

**Cloud Environment (please complete the following information):**
 - Cloud Provider: All

**Additional context**
No any additional context.

przemyslavic:
It has been tested. The certificate generation/copying now depends on the value of the tls_disable parameter.

Issue #1398, opened 2020-06-30
[BUG] Vault installation fails when using canal/calico network plugin


przemyslavic:
**Describe the bug**
Vault installation fails when using canal/calico network plugin

**To Reproduce**
1. edit config file to use calico/canal network plugin
2. execute `epicli apply`

**Expected behavior**
The cluster was successfully deployed

**OS (please complete the following information):**
 - OS: [all]

**Cloud Environment (please complete the following information):**
 - Cloud Provider [e.g. AWS/Azure]

**Additional context**

```
2020-06-29T21:16:46.8872983Z 21:16:46 INFO cli.engine.ansible.AnsibleCommand - TASK [vault : Run configuration script] ****************************************
[...]

PR #1434, opened 2020-07-10
Fix issue with Vault and Kubernetes Calico/Canal communication

erzetpe:
This request is to solve issues: #1412, #1413, #1398


erzetpe:
Later on for sure, but right now it's just quick fix. This will be done with prior parametrization of other options in Vault with replacing bash script with something more suitable.

seriva:
Ok, make sure this is somewhere represented in some issue or task,

Issue #896, opened 2020-01-31
[Kubernetes upgrade] Update documentation


toszo:
Add upgrade related documentation as mentioned in comments in this taks.

mkyc:
@toszo I believe this task is good start to what Sapna wrote in email. It's visible that epicli is not simple tool for the team over there. 

We could start with documentation for her and after review we could proceed with CICD tooling in separate milestone.

rpudlowski93:
@toszo  Could Your provide description and what exactly need to be updated in documentation?

mkyc:
Helm documentation is also missing. @ar3ndt you can use this task.

toszo:
In Kubernetes 1.17, many of the API was removed. 
Information about Kubernetes deprecated API's (people still use `extensions/v1beta1` in deployments). We need to enclose information that all deployments, configmaps, secrets, policies, etc. have to be updated to use new Kubernetes APIs.

to-bar:
Please keep in mind that we need to add to the documentation information about not supported Epiphany versions (below v0.4.4) since 0.7.0 required K8s in version 1.14.6+. I created issue #1433 to add preflight check for this.

mkyc:
helm documentation already added in #1435.

PR #1435, opened 2020-07-10
helm documentation #896

ar3ndt:


ar3ndt:
> Can you also add an entry to:
> 
> /epiphany/docs/home/HOWTO.md

added

Issue #1424, opened 2020-07-03
[BUG] Terraform returning an error during deployments on Azure ("A retryable error occurred.")


przemyslavic:
**Describe the bug**
There is often an issue with Azure cluster deployment due to a terraform error saying "A retryable error occurred."

**To Reproduce**
1. execute `epicli apply -f test.yml` (configuration given below)

**Expected behavior**
The cluster has been deployed successfully.

**Config files**
```yaml
---
kind: epiphany-cluster
name: test
provider: azure
specification:
  admin_user:
    key_path: /path/to/id_rsa
    name: operations
  cloud:
    region: xxx
    subscription_name: xxx
    use_public_ips: true
    use_service_principal: false    
  components:
    kafka:
      count: 2
    kubernetes_master:
      count: 1
    kubernetes_node:
      count: 3
    load_balancer:
      count: 1
    logging:
      count: 2
    monitoring:
      count: 1
    postgresql:
      count: 2
    rabbitmq:
      count: 2
    ignite:
      count: 2
    opendistro_for_elasticsearch:
      count: 2
  name: test
  prefix: 'qa'
title: Epiphany cluster Config
```
**OS (please complete the following information):**
 - OS: [all]

**Cloud Environment (please complete the following information):**
 - Cloud Provider [MS Azure]

**Additional context**
Log:
```
13:08:18 INFO cli.engine.terraform.TerraformCommand - azurerm_virtual_machine.ci-retryable-monitoring-vm-0: Still creating... [1m40s elapsed]
13:08:22 INFO cli.engine.terraform.TerraformCommand - azurerm_virtual_machine.ci-retryable-ignite-vm-1: Still creating... [1m40s elapsed]
13:08:25 INFO cli.engine.terraform.TerraformCommand - azurerm_virtual_machine.ci-retryable-kubernetes-node-vm-0: Still creating... [1m40s elapsed]
13:08:28 INFO cli.engine.terraform.TerraformCommand - azurerm_virtual_machine.ci-retryable-kubernetes-node-vm-1: Still creating... [1m50s elapsed]
13:08:28 INFO cli.engine.terraform.TerraformCommand - azurerm_virtual_machine.ci-retryable-monitoring-vm-0: Still creating... [1m50s elapsed]
[...]

PR #1437, opened 2020-07-10
Fix: Terraform returning an error during deployments on Azure "A retryable error occurred." (#1424)

seriva:
- Extended logging to pipe stderr logs to proper log level
- Extended logging levels with colors
- Added retry to Terraform to resolve "RetryableError" thrown by Terraform


mkyc:
Change log entry, at least about logging improvements?

seriva:
Added the changelog.

Issue #1395, opened 2020-06-29
[BUG] The RabbitMQ deployment - clustering is not working properly 


przemyslavic:
**Describe the bug**
There is an issue with cluster formation.

**To Reproduce**
1. Deploy the cluster with the RabbitMQ application
2. Run `kubectl exec -it -n=queue rabbitmq-cluster-0 -- rabbitmqctl cluster_status` and `kubectl exec -it -n=queue rabbitmq-cluster-1 -- rabbitmqctl cluster_status`

**Expected behavior**
The RabbitMQ deployment is clustered.
The command result shows 2 running nodes.

**Additional context**
Actual command result:
```
[ec2-user@ec2-15-236-60-122 ~]$ kubectl exec -it -n=queue rabbitmq-cluster-0 -- rabbitmqctl cluster_status
Cluster status of node rabbit@10.244.3.16 ...
[{nodes,[{disc,['rabbit@10.244.3.16']}]},
 {running_nodes,['rabbit@10.244.3.16']},
 {cluster_name,<<"rabbit@rabbitmq-cluster-0.rabbitmq-cluster.queue.svc.cluster.local">>},
 {partitions,[]},
 {alarms,[{'rabbit@10.244.3.16',[]}]}]
```
```
[ec2-user@ec2-15-236-60-122 ~]$ kubectl exec -it -n=queue rabbitmq-cluster-1 -- rabbitmqctl cluster_status
Cluster status of node rabbit@10.244.2.15 ...
[{nodes,[{disc,['rabbit@10.244.1.10','rabbit@10.244.1.14',
                'rabbit@10.244.2.11','rabbit@10.244.2.15',
                'rabbit@10.244.3.12','rabbit@10.244.3.8']}]},
 {running_nodes,['rabbit@10.244.2.15']},
 {cluster_name,<<"rabbit@rabbitmq-cluster-0.rabbitmq-cluster.queue.svc.cluster.local">>},
 {partitions,[]},
 {alarms,[{'rabbit@10.244.2.15',[]}]}]
```
```
[ec2-user@ec2-15-236-60-122 ~]$ kubectl get pods -n=queue -o wide
NAME                 READY   STATUS    RESTARTS   AGE     IP            NODE                                                 NOMINATED NODE   READINESS GATES
rabbitmq-cluster-0   1/1     Running   0          4h12m   10.244.3.16   ec2-15-236-95-83.eu-west-3.compute.amazonaws.com     <none>           <none>
[...]

PR #1464, opened 2020-07-15
bump rabbitmq version from 3.7.10 to 3.8.3 #1395

ar3ndt:


przemyslavic:
And what about existing clusters with deployments in version 3.7.10? Upgrading/patching?

pyrkamarcin:
> And what about existing clusters with deployments in version 3.7.10? Upgrading/patching?

@ar3ndt correct me if I'm wrong, but there is no option to soft upgrade on this moment. Current upgrade process only patching image version on StatefulSet on memory manifest, then new version of pod has no created without manual delete old and lose data.

ar3ndt:
> > And what about existing clusters with deployments in version 3.7.10? Upgrading/patching?
> 
> @ar3ndt correct me if I'm wrong, but there is no option to soft upgrade on this moment. Current upgrade process only patching image version on StatefulSet on memory manifest, then new version of pod has no created without manual delete old and lose data.

yeah that would be kinda problematic with gracefully bumping it while there might be data in the queues.
Lets hold on with this

seriva:
Missing changelog entry.

ar3ndt:
> My general question is: is that possible to have value `3.8.3` in some uber-vars? Can we have it from single place and then just change `3.7.10` to `3.8.3` in one line?

I think so, maybe by the way of modularisation task.

Issue #1461, opened 2020-07-14
[BUG] Error upgrading AWS/RHEL/calico cluster from 0.6 to 0.7


przemyslavic:
**Describe the bug**
The cluster cannot be upgraded from version 0.6 to version 0.7 in the following configuration: `AWS/RHEL/calico`.
Ansible fails on task `[kubernetes_master : Check if NetworkManager service is loaded]`.

**To Reproduce**
1. execute `epicli apply -f test.yml` using `epiphanyplatform/epicli:0.6.0` docker image (configuration given below)
3. execute `epicli upgrade -b /path/to/build/directory/` (from develop branch)

**Expected behavior**
The cluster has been upgraded without errors.

**Config files**
Configuration that should be included in the yaml file:
```yaml
---
kind: configuration/kubernetes-master
name: default
provider: aws
specification:
  advanced:
    networking:
      plugin: calico
```

**OS (please complete the following information):**
 - OS: [e.g. RHEL7]

**Cloud Environment (please complete the following information):**
 - Cloud Provider [AWS]

**Additional context**
Log:

```
2020-07-13T19:55:34.5656419Z 19:55:34 INFO cli.engine.ansible.AnsibleCommand - TASK [kubernetes_master : Apply network plugin configured by user] *************
2020-07-13T19:55:34.7674392Z 19:55:34 INFO cli.engine.ansible.AnsibleCommand - included: /shared/build/06todevawrhcalico/ansible/roles/kubernetes_master/tasks/./cni-plugins/calico.yml for ec2-xx-xx-xx-xx.eu-west-3.compute.amazonaws.com
2020-07-13T19:55:34.8351576Z 19:55:34 INFO cli.engine.ansible.AnsibleCommand - 
2020-07-13T19:55:34.8358102Z 19:55:34 INFO cli.engine.ansible.AnsibleCommand - TASK [kubernetes_master : Check if /etc/NetworkManager/conf.d exists] **********
2020-07-13T19:55:36.2312123Z 19:55:36 INFO cli.engine.ansible.AnsibleCommand - ok: [ec2-xx-xx-xx-xx.eu-west-3.compute.amazonaws.com]
2020-07-13T19:55:36.3470696Z 19:55:36 INFO cli.engine.ansible.AnsibleCommand - 
[...]

PR #1472, opened 2020-07-16
Got rid of pipe and grep #1461

rafzei:


core/src/epicli/data/common/ansible/playbooks/roles/kubernetes_master/tasks/cni-plugins/calico.yml
```diff
@@ -19,24 +19,19 @@
...
-            set -o pipefail &&
-            systemctl list-units --type=service --state=loaded | grep -q NetworkManager.service
-          args:
-            executable: /bin/bash
-          register: network_manager_service_status
-          changed_when: false
+- name: Reload NetworkManager service
+  block:
+    - name: Check if NetworkManager service is loaded
+      command: systemctl show NetworkManager.service --property=LoadState
```
to-bar, line 25: I'm not sure if the new command is an equivalent of the previous one.

Check this:
```
[ec2-user@ec2-35-181-55-88 ~]$ systemctl list-units --type=service --state=loaded | grep filebeat
filebeat.service                                      loaded active running Filebeat sends log files to Logstash or directly to Elasticsearch.
[ec2-user@ec2-35-181-55-88 ~]$ systemctl show filebeat --property=LoadState
LoadState=loaded
[ec2-user@ec2-35-181-55-88 ~]$ sudo systemctl stop filebeat
[ec2-user@ec2-35-181-55-88 ~]$ systemctl list-units --type=service --state=loaded | grep filebeat
[ec2-user@ec2-35-181-55-88 ~]$ echo $?
1
[ec2-user@ec2-35-181-55-88 ~]$ systemctl show filebeat --property=LoadState
LoadState=loaded
```
It looks like `--property=SubState` or `--property=ActiveState` could work:
```
[ec2-user@ec2-35-181-55-88 ~]$ systemctl show filebeat --property=ActiveState
ActiveState=inactive
[ec2-user@ec2-35-181-55-88 ~]$ systemctl show filebeat --property=SubState
SubState=dead
[ec2-user@ec2-35-181-55-88 ~]$ sudo systemctl start filebeat
[ec2-user@ec2-35-181-55-88 ~]$ systemctl show filebeat --property=ActiveState
ActiveState=active
[ec2-user@ec2-35-181-55-88 ~]$ systemctl show filebeat --property=SubState
SubState=running
[ec2-user@ec2-35-181-55-88 ~]$ systemctl show NetworkManager.service --property=SubState
SubState=running
[ec2-user@ec2-35-181-55-88 ~]$ systemctl show NetworkManager.service --property=ActiveState
ActiveState=active
[ec2-user@ec2-35-181-55-88 ~]$ systemctl show NetworkManagerTest.service --property=ActiveState
ActiveState=inactive
[ec2-user@ec2-35-181-55-88 ~]$ echo $?
0
systemctl show NetworkManagerTest.service --property=SubState
SubState=dead
```
I would consider keeping grep, since it's simple. However, `set -o pipefail` is unnecessary but the task should have condition `failed_when: network_manager_service_status.rc > 1` and maybe it could have a more accurate name, e.g. `Check if NetworkManager service needs to be reloaded`.

rafzei: Both checking loadstate but `systemctl list-units --type=service --state=loaded` tell us that service is loaded and active. We have to keep in mind that service can be loaded even if its inactive, but it cannot be reloaded. I will fix that :+1:

PR #1488, opened 2020-07-27
More exporters

sunshine69:
New features:
- Added posgres-exporter role and playbook to implement postgres-exporter

Stability improvement:
- download-requirements role - re order execution so that the dependencies are satisfied
- set DEBIAN_FRONTEND=noninteractive to avoid prompting when installing libssl. This caused it hung forever.

Tested Outcome:
- postgres-exporter configured and deploy
- epicli can re-run many times without errors because of the debian apt system and download dependencies issues.

Not addresses yet to come.

- At times ansible run apt to install and conflict with local ubuntu auto upgrade which resulted an error because the dir /var/lib/dpkg is locked. If this happen, re-run after some minutes when the auto upgrade is completed should work but it might be better to handle it in epiphany.


seriva:
This seems to be combined with the code from this review where changes where requested:

https://github.com/epiphany-platform/epiphany/pull/1411

Please make pull request only with the changes for the exporters.

PR #1518, opened 2020-08-03
Implement ansible role postgres-exporter

sunshine69:


seriva:
I think this should probably install in the same way as the rest of the exporters by setting up the role inside the ```configuration/feature-mapping``` and adding it to the ```postgresql``` role mapping.

seriva:
@erzetpe and @plirglo Can you also weigh in here since you where implementing most of the Postgress configuration and plugins.

core/src/epicli/data/common/ansible/playbooks/roles/postgres_exporter/meta/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+dependencies:
+  - role: preflight_facts
```
erzetpe, line 3: Are you sure that preflight_facts will be used in your role?

core/src/epicli/data/common/ansible/playbooks/postgres_exporter.yml
```diff
@@ -0,0 +1,6 @@
+---
+- hosts: postgres_exporter[0]
```
plirglo, line 2: I think we need to handle also clustered postgres, we should iterate over the list.

core/src/epicli/data/common/ansible/playbooks/roles/postgres_exporter/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+
+postgres_instances: "--postgres.server={{ groups['postgres'][0] }}"
```
plirglo, line 3: We should handle also clustered postgres

sunshine69:
> I think this should probably install in the same way as the rest of the exporters by setting up the role inside the `configuration/feature-mapping` and adding it to the `postgresql` role mapping.

Not sure what you mean. I make this role using the similar role `kafka_exporter` approach.

postgres should handle postgres stuff only. The exporter is not part of postgres bu part of monitoring system and relates to prometheus - and postgres. Thus if you mean that we need to make the postgres role to handle the exporter then I don't think it is a good idea.

core/src/epicli/data/common/ansible/playbooks/roles/postgres_exporter/defaults/main.yml
sunshine69: Agreed. But there is a cost of reading postgres data in all postgres hosts and it might affect the postgres performance. Also I am not sure if it is OK to have a repetition of the monitoring data if we run it in all postgres hosts.

Kafka exporter does the similar thing but then Keith (ABB US team) decided to turn off as it produce redundant data and causing kafka significant load or issues.

While I agree the correctness but I need to look into how the exporter detect if they can be clustered? There is no mention or feature of postgres exporter supporting multiple read of the same cluster and report to the same prometheus. If I miss read some doco please let me know the resource that postgres exporter works fine for multiple read in different posgres hosts. More importantly what benefit do we get from that scenario? 

Also please remember that posgres cluster is not really HA service. It is one write and standby read. Each postgres node has different role at a time while it is running, it is not a equal role HA replacement. Thus installing in all postgres nodes does not make sense to me.

Handling cluster would be still only in one node, and automatically deploy/switch to the other node if the role of the node changes. But then it will be too complex and less reliable, not straight forward implementation.

Or if postgres exporter support a fail over connection string then we better install it as a kubernetes app and querying using this mechanism. 

I am delighted to see any opinion about it after getting my idea through.

core/src/epicli/data/common/ansible/playbooks/roles/postgres_exporter/meta/main.yml
sunshine69: preflight facts is started at the beginning inside epiphany code and not related to this role. So 100% it is used.

core/src/epicli/data/common/ansible/playbooks/roles/postgres_exporter/defaults/main.yml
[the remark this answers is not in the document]
sunshine69: Please see comment below with regard to the cluster support

plirglo: Maybe we don't need to collect data from all nodes in most cases but there are sill cases when we need to collect them eg:

1. Application might be configured to write data into node0, and read data from node1 - in this case good to see metrics from both nodes
2.. Cluster failover: node0 going down and node1 is promoted to be primary - in this case we will not see metrics from (new) primary node because there is no exporter.

I didn't find any other nice and lite solution for that, instead install exporter on all nodes. We need this metrics also after failover.

I will merge this PR into feature branch: feature/postgres-exporter and will continue work on this (modify a bit to add support for HA and run tests)
 
Thank you for your contribution, you did a good job. We really appreciate this!

PR #1521, opened 2020-08-04
make-download-requirements-more-stable

sunshine69:
- Change the order of executing the task download-requirements.yml after we populate the directory /tmp/epi-repository-setup-scripts.

I have hit a situation (many times when re-run from the current system) that this role failed because the script
core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/ubuntu-18.04/download-requirements.sh

this part

```
repos_backup_file="/tmp/epi-repository-setup-scripts/enable-system-repos.sh"
if [[ ! -f /etc/apt/sources.list ]]; then
    if [[ -f /var/tmp/enabled-system-repos.tar ]] && [[ -f ${repos_backup_file} ]]; then
        echol "OS repositories seems missing, restoring..."
        ${repos_backup_file}
    else
        echol "/etc/apt/sources.list seems missing, you either know what you're doing or you need to fix your repositories"
    fi
fi

```

It comes to the the part `/etc/apt/sources.list seems missing, you either know what you're doing or you need to fix your repositories`

As you can see this relies on the /tmp/epi-repository-setup-scripts/enable-system-repos.sh.

- Set DEBIAN_FRONTEND=noninteractive for download-requirements.sh

A similar PR has been addressed this issues but in different place and because
that run before this stage the problem is hidden because these dependencies
packages are already install (libssl). However if you re-run it on existing
system it might happen again. Or in some combination situation it happened
again.

IMHO there is no side effect of adding this change to be sure we are not
prompted by apt for whatever reason.

### DIFF
--- a/core/src/epicli/data/common/ansible/playbooks/roles/repository/tasks/setup.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/repository/tasks/setup.yml
@@ -39,9 +39,6 @@
         dest: /tmp/epi-download-requirements/download-requirements.sh
         mode: a+x
 
-    - name: Download Epiphany requirements
-      include_tasks: download-requirements.yml
-
   when:
     - not offline_mode
     - not custom_repository_url
@@ -71,6 +68,15 @@
     force: no # if target exists it will skip, default is 'yes'
   when:
     - offline_mode
+    - not custom_repository_url
+    - inventory_hostname in target_repository_hostnames
+
+# Put it down here as the script download-requirements.sh relies on some files
+# in the folder /tmp/epi-repository-setup-scripts that until now it should be available
+- name: Download Epiphany requirements
+  include_tasks: download-requirements.yml
+  when:
+    - not offline_mode
     - not custom_repository_url
     - inventory_hostname in target_repository_hostnames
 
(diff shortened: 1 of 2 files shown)


core/src/epicli/data/common/ansible/playbooks/roles/repository/tasks/setup.yml
```diff
@@ -74,6 +71,15 @@
     - not custom_repository_url
     - inventory_hostname in target_repository_hostnames
 
+# Put it down here as the script download-requirements.sh relies on some files
```
sk4zuzu, line 74: I don't think this change makes any difference, that's because  the `enable-system-repos.sh` script depends on the `/var/tmp/enabled-system-repos.tar` file and the `/var/tmp/enabled-system-repos.tar`  file is created at the `- name: Set up repositories`. This means the `[[ -f /var/tmp/enabled-system-repos.tar ]] && [[ -f ${repos_backup_file} ]]` condition will evaluate to false anyway regardless of the changed order. 🤔

sunshine69: In the `normal` situation your assumption is correct, that is the first time runs, no errors due to this.

However if the build fail, we re-run on existing cluster then I hit many times that the cyclic dependencies which I describe here. ssh in the system and found what I described in the change request.

No more investigation why and how the system ends up like this as no time. However consider the change wont do harm and most important, it fixes the re-run situation for me.

I am using this change and my system build reliability is sort of perfect now.

sk4zuzu: Ok then. I did some manual testing and I don't see any harm caused by this. I'm OK with merging it. :+1:😇

sk4zuzu (approved):

rafzei (approved):
Looks good to me as well

PR #1528, opened 2020-08-05
Rook ceph cluster description

ar3ndt:


docs/design-docs/rook-ceph/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Prerequisite
+
+
+Since version 1.4 lvm package present on the nodes is required.
+Example installation command:
+
+RHEL:
+
+`yum install lvm2 -y`
```
mkyc, line 17: isn't it done by epiphany already?

mkyc (approved):
One little comment.

docs/design-docs/rook-ceph/README.md
```diff
@@ -0,0 +1,213 @@
...
+After creating basic setup (`common.yaml`, `operator.yaml`, `cluster.yaml`) install toolbox (`toolbox.yaml`) as well for checking the ceph cluster status.
+
+**IMPORTANT**:
+
+ensure the **osd** container is created and running. It requires a storage device to be available on the nodes.
+
+During cluster startup it searches for the devices available and creates osd containers for them.
+
+Kubelet nodes have to use a default flag `enable-controller-attach-detach` set to `true`. Otherwise PVC will not attach to the pod.
+
```
rpudlowski93, line 89: I could recommend to add here the location of the file where we can find the flag, it will help: 

/var/lib/kubelet/kubeadm-flags.env  on every worker nodes with kubelet. After that we need to restart kubelet: systemctl restart kubelet

mkyc (dismissed):

docs/design-docs/rook-ceph/README.md
rpudlowski93: good question, if no, should we add here also command for ubuntu? sudo apt-get install lvm2?

rpudlowski93 (approved):

docs/design-docs/rook-ceph/README.md
mkyc: Let's not bother about it.

mkyc (approved):

rpudlowski93 (approved):

Issue #1443, opened 2020-07-13
EKS cluster autoscaling tests


mkyc:
Prepare manual load tests to ensure that cluster scales up and down as expected. Load tests will be included later into epiphany code but not at that moment.

**Is your feature request related to a problem? Please describe.**
EKS clusters we will provide in epiphany need to have horizontal autoscaling enabled and tested. 

**Describe the solution you'd like**
There should be some Kubernetes Deployment defined with images applying some high load on cluster. With this deployment we should be able to scale number of replicas up and down, and check if vertical autoscaler is responding accordingly and scales nodes up and down. 

Deployment and description of commands should be committed to sub-directory with terraform scripts of EKS cluster. There is no need to integrate it with any test pipeline at this point. 

**Describe alternatives you've considered**
Unknown. 

**Additional context**
No.

mkyc:
This task is extremely similar to #1441 and code might be shared by both tasks.

przemyslavic:
I'm having problems deploying a cluster.
Looks like an issue installing chart.
```
Error: failed to download "stable/cluster-autoscaler" (hint: running `helm repo update` may help)

  on cluster-autoscaler.tf line 1, in resource "helm_release" "cluster-autoscaler":
   1: resource "helm_release" "cluster-autoscaler" {
```

mum0l:
@przemyslavic  - I am working on that.. Looks like there is an issue when deploying 2 clusters.
Simplest solution which is working properly is to re-use the "cluster-autoscaler" role for other deployments, but @pyrkamarcin  mention that this solution might be not secure, so I am looking for other possibilities to run multiply roles  - one per cluster. Sadly its quite tought

mum0l:
Just make a new PR: https://github.com/epiphany-platform/epiphany/pull/1623 - please review

mum0l:
I already test the autoscaling feature, but it will be good if someone else will check it.
Some docs, which I did recently:
https://github.com/mum0l/epiphany/blob/feature/eks_autoscaler_fix/docs/design-docs/eks/README.md#testing-the-autoscaling-feature

mum0l:
@przemyslavic could you test it again?

przemyslavic:
[...]

PR #1568, opened 2020-08-19
1443 eks cluster autoscaler

mum0l:
Finally! working version of EKS cluster autoscaler.


mkyc:
@mum0l please change merge target from develop to feature/eks-cluster-creation

mum0l:
@mkyc - good point - fixed!

feature/eks-cluster-creation/cluster-metrics-server.tf
```diff
@@ -0,0 +1,4 @@
+module "metrics_server" {
+  source = "cookielab/metrics-server/kubernetes"
+  version = "0.9.0"
+}
```
mkyc, line 4: newline

feature/eks-cluster-creation/cluster-metrics-server.tf
```diff
@@ -0,0 +1,4 @@
+module "metrics_server" {
```
mkyc, line 1: why we need it?

feature/eks-cluster-creation/aws-autoscaling-groups.tf
```diff
@@ -0,0 +1,57 @@
...
+
+    actions = [
+      "autoscaling:DescribeAutoScalingGroups",
+      "autoscaling:DescribeAutoScalingInstances",
+      "autoscaling:DescribeLaunchConfigurations",
+      "autoscaling:DescribeTags",
+      "ec2:DescribeLaunchTemplateVersions",
+    ]
+
+    resources = ["*"]
```
mkyc, line 30: shouldn't it have something more restrictive?

feature/eks-cluster-creation/aws-autoscaling-groups.tf
```diff
@@ -0,0 +1,57 @@
...
+    sid    = "clusterAutoscalerOwn"
+    effect = "Allow"
+
+    actions = [
+      "autoscaling:SetDesiredCapacity",
+      "autoscaling:TerminateInstanceInAutoScalingGroup",
+      "autoscaling:UpdateAutoScalingGroup",
+    ]
+
+    resources = ["*"]
```
mkyc, line 43: shouldn't it have something more restrictive?

feature/eks-cluster-creation/cluster-autoscaler.tf
```diff
@@ -0,0 +1,42 @@
...
+    name  = "image.tag"
+    value = var.eks_autoscaler_version
+  }
+  set {
+    name  = "extraArgs.scale-down-utilization-threshold"
+    value = var.cluster-autoscaler_scale-down-utilization-threshold
+  }
+  set {
+    name  = "rbac.serviceAccountAnnotations.eks\\.amazonaws\\.com/role-arn"
+    value = "arn:aws:iam::540350320930:role/cluster-autoscaler"
```
mkyc, line 38: should we have value here?

feature/eks-cluster-creation/cluster-autoscaler.tf
```diff
@@ -0,0 +1,42 @@
...
+  set {
+    name  = "image.tag"
+    value = var.eks_autoscaler_version
+  }
+  set {
+    name  = "extraArgs.scale-down-utilization-threshold"
+    value = var.cluster-autoscaler_scale-down-utilization-threshold
+  }
+  set {
+    name  = "rbac.serviceAccountAnnotations.eks\\.amazonaws\\.com/role-arn"
```
mkyc, line 37: why do you escape some of `.` ?

feature/eks-cluster-creation/aws-eks.tf
```diff
@@ -1,15 +1,14 @@
 module "eks" {
   source          = "terraform-aws-modules/eks/aws"
   version         = "12.2.0"
-  cluster_name    = var.eks_cluster_name
-  subnets         = module.vpc.private_subnets
+  cluster_name    = var.cluster_name
+  subnets         = module.vpc.public_subnets
```
mkyc, line 5: why this change?

feature/eks-cluster-creation/aws-eks.tf
```diff
@@ -1,15 +1,14 @@
...
   vpc_id          = module.vpc.vpc_id
   cluster_version = var.eks_cluster_version
-
+  enable_irsa     = true
```
mkyc, line 8: can you describe this value? I don't know what it does. Some comment here would be helpful.

feature/eks-cluster-creation/vars.tf
```diff
@@ -1,12 +1,16 @@
...
-// Cluster name
-variable "eks_cluster_name" {
-  default = "eks-terraform-cluster"
-}
-
 // Cluster version
 variable "eks_cluster_version" {
-  default = "1.17"
+  // default = "1.17"
+  default = "1.16"
```
mkyc, line 4: why downgraded k8s version?

mkyc (commented):
Some comments added. Also: 
* Can you add section to readme describing autoscaling part? 
* What are sources you used to implement this? 
* can you move this code to docs/design-docs/eks directory? Similar as is in [this branch for aks](https://github.com/epiphany-platform/epiphany/tree/feature/aks-cluster-creation/docs/design-docs/aks)
* Most important is, can you provide any test results? I've seen in commits with some load generation, but I need to see any test result here or in related task #1443

feature/eks-cluster-creation/main.tf
```diff
@@ -2,6 +2,10 @@
 # # https://learn.hashicorp.com/terraform/kubernetes/provision-eks-cluster#optional-configure-terraform-kubernetes-provider
 # # To learn how to schedule deployments and services using the provider, go here: ttps://learn.hashicorp.com/terraform/kubernetes/deploy-nginx-kubernetes.
 
+terraform {
+  required_version = ">= 0.12"
+}
+
```
rpudlowski93, line 8: I had the same in my AKS but I change the value to "=0.12.6" in order to pin exacly one version. Maybe it could be good standard. But still it is small detail.

feature/eks-cluster-creation/vars.tf
```diff
@@ -40,10 +65,10 @@ variable "worker1_data" {
   })
   default = {
     name = "eks-worker-1"
-    instance_type = "t2.small"
-    asg_desired_capacity = 2
+    instance_type = "t2.medium"
+    asg_desired_capacity = 1
     asg_min_size = 1
-    asg_max_size = 5
+    asg_max_size = 10
```
rpudlowski93, line 71: Maybe should we only declare variables in vars.tf file and define the variables in terraform.tfvars? It could be nice to have  one standard for AKS and EKS.

feature/eks-cluster-creation/cluster-metrics-server.tf
mum0l: Yes it is, as cluster autoscaler took metrics about current cluster usage, which are created by metrcis-server. Without that, cluster-autoscaler wont work.

feature/eks-cluster-creation/aws-autoscaling-groups.tf
mum0l: No, as we set only specific autoscaling actions on resources, so it shouldn't be a big deal. Sadly they do not include IAM polices in the module, so that's need to be created manually like that.

feature/eks-cluster-creation/aws-autoscaling-groups.tf
mum0l: No, as we set only specific autoscaling actions on resources, so it shouldn't be a big deal. Sadly they do not include IAM polices in the module, so that's need to be created manually like that.

feature/eks-cluster-creation/main.tf
mum0l: I think, I need to keep it as it is, as when I try to put this version, I get:
`Error: Unsupported Terraform Core version
  on main.tf line 6, in terraform:
   6: 	required_version = "=0.12.6"`

feature/eks-cluster-creation/vars.tf
mum0l: Good point, i just moved them into similarly created dir in ~/docs/design-docs/

feature/eks-cluster-creation/cluster-autoscaler.tf
mum0l: This value has been migrated into tfvars file.

feature/eks-cluster-creation/aws-eks.tf
mum0l: Just did it. This value is required to enable IAM Roles for Service Accounts which are created during terraform run.

feature/eks-cluster-creation/vars.tf
mum0l: This is just to show, that version 1.16 is supported too. I just changed it to 1.17

feature/eks-cluster-creation/main.tf
rpudlowski93: Yes exacly, since terraform installed the latest version i suppose and all the time there will be installed the latest version using this code, we don't even know which one exacly. So maybe pin to the latest version for example 0.13.0? Using >= is similar to using docker images without any tag which is danger in some cases in my opinion.

feature/eks-cluster-creation/aws-autoscaling-groups.tf
mkyc: I'm pretty sure enumerating exact resources would be better. Every security person would get triggered on `*` ;) That being said I do not think we need it now.

rpudlowski93 (approved):

Issue #1570, opened 2020-08-20
[BUG] Azure network security groups: eplicli apply makes changes when configuration file is not changed


to-bar:
**Describe the bug**
`epicli apply` is not idempotent on Azure. Odd runs attach network security groups to subnets and even runs detach.

**To Reproduce**
1. Prepare minimalistic config file (use defaults)
2. Execute `epicli apply -f az_nsg_association_bug.yml`
3. Check network security group association in Azure Portal (e.g. Home | repro-nsg-association-bug-kubernetes-master-vm-0 | Networking)
4. Repeat steps 2 and 3.

**Expected behavior**
Terraform should not modify associations (state) unless configuration has been changed. 

**Config files**
If applicable, add config files to help explain your problem.
[https://app.zenhub.com/files/164668263/af24eadf-1a93-45d7-b166-d880fb4612d9/download](https://app.zenhub.com/files/164668263/af24eadf-1a93-45d7-b166-d880fb4612d9/download)

**Cloud Environment**
Cloud Provider: Azure

**Additional context**
Found in version 0.7.1.

**More details**

1. After first run:

In Azure Portal (Home | repro-nsg-association-bug-kubernetes-master-vm-0 | Networking):

> Network security group repro-nsg-association-bug-kubernetes-master-nsg-0 (attached to subnet: repro-nsg-association-bug-kubernetes-master-subnet-0)
> Impacts 1 subnets, 1 network interfaces
> 
> Network security group repro-nsg-association-bug-kubernetes-master-nsg-0 (attached to network interface: repro-nsg-association-bug-kubernetes-master-nic-0)
> Impacts 1 subnets, 1 network interfaces

2. No changes in configuration.

During second run:

```
10:01:27 INFO cli.engine.terraform.TerraformCommand - azurerm_storage_share.repro-nsg-association-bug-k8s-ss: Refreshing state... [id=https://repronsgassociationbugk8s.file.core.windows.net/k8s]
10:01:39 INFO cli.engine.terraform.TerraformCommand - azurerm_subnet.repro-nsg-association-bug-kubernetes-master-subnet-0: Modifying... [id=/subscriptions/2d60775f-932a-4cf6-b9f0-548a8b43b368/resourceGroups/repro-nsg-association-bug-rg/providers/Microsoft.Network/virtualNetworks/repro-nsg-association-bug-vnet/subnets/repro-nsg-association-bug-kubernetes-master-subnet-0]
[...]

PR #1573, opened 2020-08-20
Fixs for changing Terraform templates between Epicli apply runs

seriva:
Small fix done while researching #1570. It does not fix this issue but it does make sure the Terraform templates do not change in subsequent Epicli apply runs with unchanged config yaml

 In this case the generation of a VM password on linux was removed as it is not needed by Terraform or used by Epiphany since we use SSH.

https://www.terraform.io/docs/providers/azurerm/r/virtual_machine.html#admin_password


TolikT (approved):
LGTM

to-bar (approved):

Issue #1580, opened 2020-08-21
[FEATURE REQUEST] Implement worker nodes downscale


atsikham:
**Is your feature request related to a problem? Please describe.**
This request is the result of #1497 spike. At the moment Epiphany has no ability to scale down worker nodes correctly, this ability should be added. Such task existed previously - #1202, that work should be continued.

**Describe the solution you'd like**
When `count` of k8s worker nodes is decreased in config file and new configuration applied, all necessary work, such as cordon/drain nodes should be done before node removal by Terraform. 

**Describe alternatives you've considered**
No

**Additional context**
- **[Documentation](https://github.com/epiphany-platform/epiphany/blob/develop/docs/home/howto/CLUSTER.md) about scaling process should be updated.**
- Should work for all cloud providers.

---

**DoD checklist**

- Changelog
  - [ ] updated
  - [ ] not needed
- COMPONENTS.md
  - [ ] updated
  - [ ] not needed
- Schema
  - [ ] updated
  - [ ] not needed
- Backport tasks
  - [ ] created
  - [ ] not needed
- Documentation
  - [ ] added
  - [ ] updated
  - [ ] not needed
- [ ] Feature has automated tests
- [ ] Automated tests passed (QA pipelines)
  - [ ] apply
  - [ ] upgrade
  - [ ] backup/restore
- [ ] Idempotency tested
- [ ] All conversations in PR resolved
- [ ] Solution meets requirements and is done according to design doc
- [ ] Usage compliant with license

seriva:
Not going todo anymore feature requests for Epiphany.

Issue #1578, opened 2020-08-21
[BUG] Update of 1 node with no replication to a few replicated nodes leads to data loss


atsikham:
**Describe the bug**
This bug is the result of #1497 spike.
Update of 1 node with no replication to a few replicated nodes leads to data loss. Also no `rabbitmqctl forget_cluster_node <node>` executed during downscale, so it leads to some issues, such of removed node disk presence in cluster information.

**To Reproduce**
1. deploy 1 rabbitmq node with disabled clustering (default configuration)
2. create some entities: users/queues/vhosts (see `More information` part with useful commands)
3. extend config file with a part from `Config files`
4. apply configuration with 3 instances (`count: 3`)
5. check created data, it's been lost
6. downscale to 2 instances, data is in place

**Expected behavior**
- `rabbitmqctl forget_cluster_node <node>` should be executed during downscale to be able avoid some results of #1574 
- Nothing should be changed in the cluster when applying the same configuration.
- Only specified changes should be added to the cluster when applying new configuration, upscale and downscale.
- **[Documentation](https://github.com/epiphany-platform/epiphany/blob/develop/docs/home/howto/CLUSTER.md) about scaling process should be updated.**

**Config files**
```yaml
---
kind: configuration/rabbitmq
title: "RabbitMQ"
name: default
provider: azure
specification:
  cluster:
    is_clustered: true
```

**Cloud Environment (please complete the following information):**
Azure, but should also be tested with other providers.

**Additional context**
Found in version 0.7.1.

**More information**
Verification commands
```
# check cluster status
rabbitmq-diagnostics cluster_status

# create user
rabbitmqctl add_user testuser testpassword
rabbitmqctl set_user_tags testuser administrator
rabbitmqctl set_permissions -p / testuser ".*" ".*" ".*"

# create vhost
rabbitmqctl add_vhost Some_Virtual_Host
rabbitmqctl set_permissions -p Some_Virtual_Host guest ".*" ".*" ".*"

# create queue
rabbitmq-plugins enable rabbitmq_management
wget http://127.0.0.1:15672/cli/rabbitmqadmin
chmod +x rabbitmqadmin
./rabbitmqadmin declare queue --vhost=Some_Virtual_Host name=some_outgoing_queue durable=true

# create exchange
./rabbitmqadmin declare exchange --vhost=Some_Virtual_Host name=some_exchange type=direct

# create binding
[...]

Issue #1577, opened 2020-08-21
[BUG] Incorrect number of cluster instances when scaling up/down 


atsikham:
**Describe the bug**
This bug is the result of #1497 spike. There is an incorrect number of postgresql cluster instances when scaling up/down.

**To Reproduce**
1. deploy 1 node of PostgreSQL
2. edit config file (see `Config files` part) to enable replication
3. scale up to 3 nodes (`count: 3`)
1 node is not joined to the cluster:

  ```text
  [postgres@atsikham-scale-test-postgresql-vm-0 ~]$ /usr/lib/postgresql/10/bin/repmgr cluster show
  ID | Name                                | Role    | Status    | Upstream                            | Location | Connection string
  ----+-------------------------------------+---------+-----------+-------------------------------------+----------+-------------------------------------------------------------------
  1  | atsikham-scale-test-postgresql-vm-0 | primary | * running |                                     | default  | host=10.1.6.4 user=epi_repmgr dbname=epi_repmgr connect_timeout=2
  2  | atsikham-scale-test-postgresql-vm-1 | standby |   running | atsikham-scale-test-postgresql-vm-0 | default  | host=10.1.6.6 user=epi_repmgr dbname=epi_repmgr connect_timeout=2

  [postgres@atsikham-scale-test-postgresql-vm-0 ~]$ /usr/lib/postgresql/10/bin/repmgr cluster crosscheck
                                 Name | Id |  1 |  2
  ------------------------------------+----+----+----
  atsikham-scale-test-postgresql-vm-0 |  1 |  ? |  ?
  atsikham-scale-test-postgresql-vm-1 |  2 |  ? |  ?


  [postgres@atsikham-scale-test-postgresql-vm-2 ~]$ /usr/lib/postgresql/10/bin/repmgr cluster show
  ERROR: unable to retrieve node records
  DETAIL: ERROR:  relation "repmgr.nodes" does not exist
  LINE 1: ...le, un.node_name AS upstream_node_name       FROM repmgr.nod...
                                                             ^

  [postgres@atsikham-scale-test-postgresql-vm-2 ~]$ /usr/lib/postgresql/10/bin/repmgr cluster crosscheck
  ERROR: unable to retrieve any node records
  ```
4. Scale down to 2 nodes, check cluster status
5. Remove the 3rd node's disk (bug #1574)
6. Scale up to 3 nodes, there are still 2 nodes in the cluster, but standby node is replaced:

  ```text
[...]

Issue #1576, opened 2020-08-21
[BUG][Kafka] Incorrect number of brokers/queues available after scaling up/down


atsikham:
**Describe the bug**
This bug is the result of #1497 spike.
After scaling up/down, numbers of Kafka brokers and queues are incorrect.

**To Reproduce**
Steps to reproduce the behavior (see `More information` part for some useful commands):
1. deploy 1 kafka instance
2. scale up to 3 instances
3. check the number of brokers - there is only 2 available
4. create 10 queues
5. scale kafka down to 2 instances, not all queues are listed on different nodes (in my case only 2 queues were created)
```
root@atsikham-scale-test-kafka-vm-1:~# /opt/kafka_2.12-2.3.1/bin/kafka-topics.sh --zookeeper localhost:2181 --list
topic2

root@atsikham-scale-test-kafka-vm-0:~# /opt/kafka_2.12-2.3.1/bin/kafka-topics.sh --zookeeper localhost:2181 --list
topic1
topic2
```

**Expected behavior**
- After scaling up/down, number of brokers and queues should be correct.
- **[Documentation](https://github.com/epiphany-platform/epiphany/blob/develop/docs/home/howto/CLUSTER.md) about scaling process should be updated.**

**Config files**
No

**Cloud Environment (please complete the following information):**
Azure, but should also be tested with other providers.

**Additional context**
Found in version 0.7.1.

**More information**
Verification commands
```shell
systemctl status kafka
systemctl status zookeeper
# list brokers
echo dump | nc localhost 2181 - check available brokers
/opt/zookeeper-3.4.12/bin/zkCli.sh -server localhost:2181 ls /brokers/ids
# list topics
/opt/kafka_2.12-2.3.1/bin/kafka-topics.sh --zookeeper localhost:2181 --list
# create a topic
/opt/kafka_2.12-2.3.1/bin/kafka-topics.sh --create --zookeeper localhost:2181 --replication-factor 1 --partitions 1 --topic topic1
```
Configuration files
- /opt/kafka/config/server.properties
- /opt/zookeeper/conf/zoo.cfg

przemyslavic:
Tested scenarios:
:heavy_check_mark: `epicli apply:` 3 kafka nodes -> 3 kafka nodes (re-apply)
:heavy_check_mark: `epicli apply:` 1 kafka node -> 3 kafka nodes -> 5 kafka nodes
[...]

PR #1581, opened 2020-08-21
Updated cluster documentation with current scaling state

atsikham:
Lots of components do not scale properly. These comments and references should be removed after issues are fixed, in the scope of #1574 , #1575, #1576, #1577, #1578, #1579, #1580.


mkyc (approved):
LGTM

rafzei (approved):

Issue #1296, opened 2020-05-26
Epicli does not interpret alternative yaml boolean values as true booleans.


sk4zuzu:
Epipany switched to the `ruamel.yaml` library. I believe this caused following regression problem:
```python
import yaml
import ruamel.yaml

DOC = """
a: true
b: yes
c: false
d: no
"""

print(yaml.load(DOC))
print(ruamel.yaml.YAML().load(DOC))
```
```
{'a': True, 'b': True, 'c': False, 'd': False}
ordereddict([('a', True), ('b', 'yes'), ('c', False), ('d', 'no')])
```
This means that "yes/no" (and others) are no longer converted to boolean values. Non-empty strings in python always evaluate to True, it means that user can by mistake put "no" inside yaml and expect something will not run, but the opposite will happen.

> To clarify, this has been changed with YAML 1.2:
> The primary objective of this revision is to bring YAML into compliance with JSON as an official subset. YAML 1.2 is compatible with 1.1 for most practical applications - this is a minor revision. An expected source of incompatibility with prior versions of YAML, especially the syck implementation, is the change in implicit typing rules. We have removed unique implicit typing rules and have updated these rules to align them with JSON's productions. In this version of YAML, boolean values may be serialized as “true” or “false”; the empty scalar as “null”.

Please check here https://yaml.org/spec/1.2/spec.html#tag/repository/bool

mkyc:
There is no unit tests for https://github.com/epiphany-platform/epiphany/blob/develop/core/src/epicli/cli/helpers/yaml_helpers.py This task should introduce some tests for regression of this  kind.

przemyslavic:
Tested different values for `yes` / `no` option:
```
YES_VALUES = ['y', 'Y', 'yes', 'Yes', 'YES']
NO_VALUES = ['n', 'N', 'no', 'No', 'NO']
```
:heavy_check_mark: All values were correctly converted to booleans: `true` / `false`. Additionally 2 unit tests have been added.

PR #1586, opened 2020-08-26
Haproxy refactor

sk4zuzu:


sk4zuzu:
This PR is already getting too big, so the `epicli upgrade` part will be handled in another PR. So far only the `eplici apply` part has been implemented.

Features included:
- docker based haproxy https://hub.docker.com/_/haproxy ✔️
- lightweight docker image extraction tool (replaces `docker export` command) ✔️
- runc / systemd managed reusable haproxy container (not using docker) ✔️
- reimplementation of kubernetes load balancer service ✔️
- reimplementation of the "load_balancer" feature ✔️
- identical code for all linux distros we support ✔️
- offline package ✔️
- fully working logging (kibana) and monitoring (grafana) code for the "load_balancer" feature ✔️

Sugestions for testing:
- offline / online modes for each distro Ubuntu/RHEL/CentOS (`epicli apply`)
- verify if `epicli apply` is idempotent and restarts / reloads systemd service in reaction to config changes
- deploy with monitoring and logging enabled and verify if metrics and logs are read
- backup / recovery scenarios for each distro Ubuntu/RHEL/CentOS (`epicli backup`, `eplicli recovery`)

Example config for "any" provider:
```
kind: epiphany-cluster
title: "Epiphany cluster Config"
name: any4
specification:
  name: any4
  admin_user:
    name: centos
    key_path: /workspaces/epiphany/core/src/epicli/clusters/id_rsa
  components:
    kubernetes_master:
      count: 1
      machines:
        - default-k8s-master1
    kubernetes_node:
      count: 1
      machines:
        - default-k8s-node1
    load_balancer:
      count: 1
      machines:
        - default-k8s-master3
    logging:
      count: 1
      machines:
        - default-k8s-node2
    monitoring:
      count: 1
      machines:
        - default-k8s-node3
provider: any
---
kind: configuration/shared-config
title: Shared configuration that will be visible to all roles
name: default
specification:
  use_ha_control_plane: true
  promote_to_ha: false
provider: any
---
kind: configuration/haproxy
title: "HAProxy"
name: default
specification:
  logs_max_days: 60
  self_signed_certificate_name: self-signed-fullchain.pem
  self_signed_private_key_name: self-signed-privkey.pem
  self_signed_concatenated_cert_name: self-signed-test.tld.pem
  haproxy_log_path: "/var/log/haproxy.log"
  stats:
    enable: true
    bind_address: 127.0.0.1:9000
    uri: "/haproxy?stats"
    user: operations
    password: your-haproxy-stats-pwd
  frontend:
    - name: https_front
      port: 443
      https: true
      backend:
      - http_back1
  backend: # example backend config below
    - name: http_back1
      server_groups:
      - kubernetes_node
      # servers: # Definition for server to that hosts the application.
      # - name: "node1"
      #   address: "epiphany-vm1.domain.com"
      port: 30104
provider: any
---
kind: infrastructure/machine
provider: any
name: default-k8s-master1
specification:
  hostname: z1a1
  ip: 10.40.2.10
---
kind: infrastructure/machine
provider: any
name: default-k8s-master2
specification:
  hostname: z1a2
  ip: 10.40.2.11
---
kind: infrastructure/machine
provider: any
name: default-k8s-master3
specification:
  hostname: z1a3
  ip: 10.40.2.12
---
kind: infrastructure/machine
provider: any
name: default-k8s-node1
specification:
  hostname: z1b1
  ip: 10.40.2.20
---
kind: infrastructure/machine
provider: any
name: default-k8s-node2
specification:
  hostname: z1b2
  ip: 10.40.2.21
---
kind: infrastructure/machine
provider: any
name: default-k8s-node3
specification:
  hostname: z1b3
  ip: 10.40.2.22
```

core/src/epicli/data/common/ansible/playbooks/roles/haproxy_exporter/tasks/main.yml
```diff
@@ -26,21 +25,19 @@
     file_name: "{{ exporter_file_name }}"
 
 - name: Create /opt/haproxy_exporter directories
-  become: yes
   file:
     path: "{{ item }}"
-    recurse: yes
+    recurse: true
```
atsikham, line 30: Is there a common practice of boolean values usage? Seems like not necessary change (this one and others)

core/src/epicli/data/common/ansible/playbooks/roles/haproxy_runc/tasks/install-packages-Debian.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: Install containerd.io package for Debian family
+  apt:
```
atsikham, line 3: Is it the correct place to install packages? Should it be in 'repository' role?

core/src/epicli/data/common/ansible/playbooks/roles/haproxy_runc/tasks/install-packages-RedHat.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: Install containerd.io package for RedHat family
```
atsikham, line 2: Is it the correct place to install packages? Should it be in 'repository' role?

core/src/epicli/data/common/ansible/playbooks/roles/haproxy_runc/tasks/main.yml
```diff
@@ -0,0 +1,31 @@
+---
+- name: Assert input parameters
+  assert:
+    that:
+      - haproxy_service is defined
+
+- name: Install required system packages
+  include_tasks: "install-packages-{{ ansible_os_family }}.yml"
```
atsikham, line 8: Is it the correct place to install packages? Should it be in 'repository' role?

core/src/epicli/data/common/ansible/playbooks/roles/haproxy_runc/files/extract-docker-image-V1.sh
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env bash
```
atsikham, line 1: I would use a docker daemon to control haproxy container lifecycle, but if there is such standard/requirement, this is ok.

atsikham (changes requested):

core/src/epicli/data/common/ansible/playbooks/roles/haproxy/tasks/generate-certificates.yml
```diff
@@ -11,7 +10,13 @@
...
 
 - name: Generate private key if certificate hasn't been found in directory
-  shell: openssl req -x509 -nodes -newkey rsa:4096 -keyout /tmp/{{ specification.self_signed_private_key_name }} -out /tmp/{{ specification.self_signed_certificate_name }} -days 365 -subj '/CN=test'
+  shell: |
+    openssl req \
+      -x509 -nodes -newkey rsa:4096 \
+      -keyout /tmp/{{ specification.self_signed_private_key_name }} \
+      -out /tmp/{{ specification.self_signed_certificate_name }} \
+      -days 365 \
+      -subj '/CN=test'
```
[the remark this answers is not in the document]
sk4zuzu: Not sure, this is some legacy setting. I don't know what was the idea behind the name I just made the line a little more readable.

core/src/epicli/data/common/ansible/playbooks/roles/haproxy_exporter/tasks/main.yml
sk4zuzu: Please take a look at this issue https://github.com/epiphany-platform/epiphany/issues/1296
Currently support for yes/no/on/off... is completely broken for the python part of Epiphany :-1:
YAML 1.2 dropped support of the alternative forms https://yaml.org/spec/1.2/spec.html#tag/repository/bool
Also IMHO the mere idea of having something like `on` for `true` and `no` for `false` is ridiculous :)

I don't know if these are good reasons, I like to think that they are ;)

core/src/epicli/data/common/ansible/playbooks/roles/haproxy_runc/tasks/install-packages-Debian.yml
sk4zuzu: So the "repository" is reponsible for downloading images, files etc (handles offline case as well). It hosts them in apache and registry services - so later each component / role can just use apt or yum do install them in the OS. Having said that I think the answer is yes, this looks to me as, a correct place to do it. 🤔 What do you suggest instead?

core/src/epicli/data/common/ansible/playbooks/roles/haproxy_runc/tasks/install-packages-RedHat.yml
sk4zuzu: Same as before.

core/src/epicli/data/common/ansible/playbooks/roles/haproxy_runc/tasks/main.yml
sk4zuzu: Same as before.

core/src/epicli/data/common/ansible/playbooks/roles/haproxy_runc/files/extract-docker-image-V1.sh
sk4zuzu: So my rationale for not using docker is that docker does not work well with systemd, because of the client-server architecture (here's some source of knowledge https://github.com/ibuildthecloud/systemd-docker on the topic). Instead in this PR I'm using runc which seems to be working fine with systemd. My rationale for not using just the docker itself (without systemd) is that we have never done that (as far as I know) and Epiphany seems to be systemd-exclusive. I think if there is a common approval for using just the docker I can live with that. Any more votes for it? 🤔

core/src/epicli/data/common/ansible/playbooks/roles/haproxy_exporter/tasks/main.yml
atsikham: The second one seems to be reasonable.

core/src/epicli/data/common/ansible/playbooks/roles/haproxy_runc/tasks/install-packages-Debian.yml
atsikham: Ok, now this is clear.

atsikham (approved):

mkyc (commented):
Why are you extracting haproxy from docker image instead of building it from sources?

seriva (approved):

sk4zuzu:
> Why are you extracting haproxy from docker image instead of building it from sources?

IMO taking in the account that we support oflfine mode it's acutally less harm if we use docker images (alpine based) than add another set of packages (needed for the build) to already existing 14GiB of stuff.

mkyc (approved):
Cool

PR #1594, opened 2020-08-28
Add Logstash to export data from Elasticsearch to csv format

rafzei:


docs/home/howto/LOGGING.md
```diff
@@ -46,3 +49,28 @@ specification:
...
 ```
+
+### How to export Elasticsearch data to csv format
+Since v0.8 Epiphany provide posibility to export data from Elasticsearch to CSV using Logstash *(logstash-oss v7.8.1*) along with *logstash-input-elasticsearch (v4.6.2)* and *logstash-output-csv (v3.0.8)* plugin.
+
+To install Logstash in your cluster add **logstash** to feature mapping for *logging, opendistro_for_elasticsearch* or *elasticsearch* group.
+
+`Note: Logstash is disabled by default after installation.`
+
+Epiphany provide a basic configuration file `(logstash-export.conf.temp)` as template for your data export.
```
atsikham, line 60: provideS

atsikham (approved):
LGTM, but small syntax fix is requested

core/src/epicli/data/common/defaults/configuration/logstash.yml
```diff
@@ -0,0 +1,7 @@
+kind: configuration/logstash
+title: "Logstash"
+name: default
+specification:
+  versions:
+    RedHat: "7.8.1"
```
sk4zuzu, line 6: This is kinda interesting what happens when an user uses the offline package and changes those versions here. 🤔 I don't think we should allow modifying versions in the config if we don't support it in other places in the same codebase. :-1: For me personally it makes no sense at all. :) I believe this should be set in "defaults" for the role and should not be exposed to the user.

core/src/epicli/data/common/validation/configuration/logstash.yml
```diff
@@ -0,0 +1 @@
+$ref: '#/definitions/unvalidated_specification'
```
sk4zuzu, line 1: I'd suggest adding some validation here (if you still want to keep version definitions in the doc exposed to users). :+1:

docs/home/howto/LOGGING.md
sk4zuzu, line 60: Does it have to be `.temp`? Is this some default way of naming it? If not then can we change it somehow? I'm asking because from the first look it's like a temporary file and this will probably trigger more people. :)

core/src/epicli/data/common/ansible/playbooks/roles/logstash/tasks/main.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Install logstash-oss {{ specification.version }}
+  include_tasks: "{{ ansible_os_family }}.yml"
+
+- name: Ensure Logstash is disabled
+  systemd:
+    name: logstash
+    enabled: no
```
sk4zuzu, line 8: Do I understand correctly that we use it only as a tool executed manually? 🤔

docs/home/howto/LOGGING.md
```diff
@@ -46,3 +49,28 @@ specification:
     repo: /var/lib/elasticsearch-snapshots
     logs: /var/log/elasticsearch
 ```
+
+### How to export Elasticsearch data to csv format
+Since v0.8 Epiphany provide posibility to export data from Elasticsearch to CSV using Logstash *(logstash-oss v7.8.1*) along with *logstash-input-elasticsearch (v4.6.2)* and *logstash-output-csv (v3.0.8)* plugin.
+
+To install Logstash in your cluster add **logstash** to feature mapping for *logging, opendistro_for_elasticsearch* or *elasticsearch* group.
+
+`Note: Logstash is disabled by default after installation.`
```
sk4zuzu, line 58: Can you explain here why it is disabled from the start? :)

sk4zuzu (changes requested):

core/src/epicli/data/common/ansible/playbooks/roles/logstash/tasks/main.yml
rafzei: Yes. FR was for non scheduled export so for now we only providing **possibility** to export

docs/home/howto/LOGGING.md
rafzei: @sk4zuzu Documentation stands clearly that it's a template but I understand this can cause a misunderstanding. Any suggestions? (maybe .draft,.template)

docs/home/howto/LOGGING.md
rafzei: Yes I can :)

core/src/epicli/data/common/defaults/configuration/logstash.yml
rafzei: That's fine, I will change it!

docs/home/howto/LOGGING.md
sk4zuzu: I'd go with `.template` - it's a little longer, but does not trigger certain people... 🤭

atsikham (approved):
LGTM

sk4zuzu (approved):
LGTM

Issue #1300, opened 2020-05-28
Configure OpenSSH according to Mozilla Infosec guidance.


erzetpe:
We need to configure OpenSSH to have more secure settings that can be shared by multiple versions of OpenSSH installed on different platforms. We need to create configuration in /etc/ssh/sshd_config similar to proposed below:

```
HostKey /etc/ssh/ssh_host_ed25519_key
HostKey /etc/ssh/ssh_host_rsa_key
SyslogFacility AUTHPRIV
AuthorizedKeysFile      .ssh/authorized_keys
KexAlgorithms curve25519-sha256@libssh.org,ecdh-sha2-nistp521,ecdh-sha2-nistp384,ecdh-sha2-nistp256,diffie-hellman-group-exchange-sha256
Ciphers chacha20-poly1305@openssh.com,aes256-gcm@openssh.com,aes128-gcm@openssh.com,aes256-ctr,aes192-ctr,aes128-ctr
MACs hmac-sha2-512-etm@openssh.com,hmac-sha2-256-etm@openssh.com,umac-128-etm@openssh.com,hmac-sha2-512,hmac-sha2-256,umac-128@openssh.com
AuthenticationMethods publickey
LogLevel VERBOSE
Subsystem sftp  /usr/lib/ssh/sftp-server -f AUTHPRIV -l INFO
PermitRootLogin No
PasswordAuthentication no
ChallengeResponseAuthentication no
UsePAM yes
PrintMotd no
```

with additional Diffie-Hellman moduli. More information on Mozilla Infosec guide site under  [Modern (OpenSSH 6.7+)](https://infosec.mozilla.org/guidelines/openssh.html#modern-openssh-67).

mum0l:
Everything looks reasonable from my point of view.

One thing which wonders me, is do we really need to enable verbose mode? Our logs might be super huge after that option and less readable.. I understood enabling this option for particular issue investigation, but setting it globally is a bit of over-productive and useless.

przemyslavic:
Tested configurations:

- Azure/AWS x Ubuntu/RedHat

- on-prem CentOS

No issues found.

PR #1606, opened 2020-09-01
Configure OpenSSH according to Mozilla Infosec guidance

atsikham:
This PR is for #1300. Need to be verified with all possible OS distributions.


erzetpe (approved):

rafzei (approved):
LGTM

PR #1613, opened 2020-09-02
Changed openssh config according to requirements

atsikham:
Made a small change according to request from Marcin Wilk.


WilkMarcin (approved):
There exist concerns with ECDSA, while ED25519 and RSA are fine.

seriva (approved):

erzetpe (approved):

PR #1619, opened 2020-09-03
Haproxy upgrade (epicli)

sk4zuzu:
#1586 should be merged first, then this one rebased and reviewed :+1:

### DIFF
--- a/core/src/epicli/cli/engine/ansible/AnsibleVarsGenerator.py
+++ b/core/src/epicli/cli/engine/ansible/AnsibleVarsGenerator.py
@@ -1,6 +1,7 @@
 import os
 import copy
 
+from cli.version import VERSION
 from cli.helpers.Step import Step
 from cli.helpers.build_saver import get_ansible_path, get_ansible_path_for_build, get_ansible_vault_path, MANIFEST_FILE_NAME
 from cli.helpers.doc_list_helpers import select_first, select_single, ExpectedSingleResultException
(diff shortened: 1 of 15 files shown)


core/src/epicli/cli/engine/ansible/AnsibleVarsGenerator.py
```diff
@@ -53,38 +55,64 @@ def generate(self):
...
 
         with open(vars_file_path, 'w') as stream:
             dump(document, stream)
 
-        self.roles_with_generated_vars.append(to_role_name(role))
+        if vars_file_name == 'main.yml':
+            self.roles_with_generated_vars.append(to_role_name(role))
+
+    def write_role_manifest_vars(self, ansible_dir, role, kind):
+        enabled_kinds = {"configuration/haproxy"}
```
atsikham, line 94: Why dict is used here and not tuple for example?

atsikham: Could you explain why do we need both `enabled_roles` and `enabled_kinds`?

atsikham (changes requested):
LGTM, but with a few questions, will be approved then

core/src/epicli/cli/engine/ansible/AnsibleVarsGenerator.py
sk4zuzu: > Why dict is used here and not tuple for example?

This is a set(), not a dict(). In python {} have double meaning, depending on syntax of the literal.

sk4zuzu: > Could you explain why do we need both `enabled_roles` and `enabled_kinds`?

This is just an optimization. This is a new thing never done before, so after some discussion with @to-bar we decided to narrow the scope of this operation. Default behavior is still to use default values.

atsikham: > > Why dict is used here and not tuple for example?
> 
> This is a set(), not a dict(). In python {} have double meaning, depending on syntax of the literal.

Agree, it's my inattention

atsikham (approved):

atsikham (approved):

rafzei (approved):
LGTM but I see no docs update (components,readme etc..)

sk4zuzu:
> LGTM but I see no docs update (components,changelog etc..)

@rafzei, please re-check now. I added components, changelog and some info how to expose k8s service (it was TODO previously).

rafzei (approved):
Nice, thanks.

atsikham (approved):

PR #1625, opened 2020-09-04
Add kubectl and Helm to epicli and devcontainer images

to-bar:


Dockerfile
```diff
@@ -21,7 +24,18 @@ COPY --from=build-epicli-wheel /src/core/src/epicli/dist/ /epicli/
 
 RUN apt-get update \
     && apt-get install --no-install-recommends -y \
-        gcc libffi-dev make musl-dev openssh-client ruby-full sudo tar unzip vim \
+        curl gcc gnupg2 libffi-dev make musl-dev openssh-client ruby-full sudo tar unzip vim \
+\
+    && echo "Adding APT repositories ..." \
+    && curl -s https://baltocdn.com/helm/signing.asc | APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=1 apt-key add - \
```
sk4zuzu, line 30: This is some wierd url :) Why not just download the binaries (statically linked) from Helm releases on github https://github.com/helm/helm/releases/tag/v3.3.1 ? :)
https://www.getbalto.com/about.html this looks like a cheezy startup company to me :D

Dockerfile
```diff
@@ -21,7 +24,19 @@ COPY --from=build-epicli-wheel /src/core/src/epicli/dist/ /epicli/
 
 RUN apt-get update \
     && apt-get install --no-install-recommends -y \
-        gcc libffi-dev make musl-dev openssh-client ruby-full sudo tar unzip vim \
+        curl gcc libffi-dev make musl-dev openssh-client ruby-full sudo tar unzip vim \
+\
+    && echo "Installing helm binary ..." \
+    && curl -LOsS https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz \
```
sk4zuzu, line 30: "-f" would be nice here, otherwise if there is some error and data cannot be downloaded curl will return 0 anyway and docker build will not fail.
From https://curl.haxx.se/docs/manpage.html we can read:
```
-f, --fail

(HTTP) Fail silently (no output at all) on server errors. This is mostly done to better enable scripts etc to better deal with failed attempts. In normal cases when an HTTP server fails to deliver a document, it returns an HTML document stating so (which often also describes why and more). This flag will prevent curl from outputting that and return error 22.

This method is not fail-safe and there are occasions where non-successful response codes will slip through, especially when authentication is involved (response codes 401 and 407).
```
```suggestion
    && curl -fLOsS https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz \
```

core/src/epicli/.devcontainer/Dockerfile
```diff
@@ -15,13 +16,25 @@ COPY requirements.txt /
...
     && apt-get install --no-install-recommends -y \
         apt-utils dialog 2>&1 \
     && apt-get install --no-install-recommends -y \
-        gcc libffi-dev make musl-dev openssh-client ruby-full sudo tar unzip vim \
+        curl gcc libffi-dev make musl-dev openssh-client ruby-full sudo tar unzip vim \
         \
         git git-lfs lsb-release procps \
+\
+    && echo "Installing helm binary ..." \
+    && curl -LOsS https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz \
```
sk4zuzu, line 29:
```suggestion
    && curl -fLOsS https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz \
```

core/src/epicli/.devcontainer/Dockerfile
```diff
@@ -15,13 +16,25 @@ COPY requirements.txt /
...
+    && tar -xzof ./helm-v${HELM_VERSION}-linux-amd64.tar.gz --strip=1 -C /usr/local/bin linux-amd64/helm \
+    && rm ./helm-v${HELM_VERSION}-linux-amd64.tar.gz \
+    && helm version \
+    && echo "Installing kubectl binary ..." \
+    && curl -LOsS https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl \
```
sk4zuzu, line 34:
```suggestion
    && curl -fLOsS https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl \
```

sk4zuzu (changes requested):

Dockerfile
```diff
@@ -21,7 +24,18 @@ COPY --from=build-epicli-wheel /src/core/src/epicli/dist/ /epicli/
 
 RUN apt-get update \
     && apt-get install --no-install-recommends -y \
-        gcc libffi-dev make musl-dev openssh-client ruby-full sudo tar unzip vim \
+        curl gcc gnupg2 libffi-dev make musl-dev openssh-client ruby-full sudo tar unzip vim \
+\
+    && echo "Adding APT repositories ..." \
+    && curl -s https://baltocdn.com/helm/signing.asc | APT_KEY_DONT_WARN_ON_DANGEROUS_USAGE=1 apt-key add - \
+    && echo "deb https://baltocdn.com/helm/stable/debian/ all main" | tee /etc/apt/sources.list.d/helm-stable-debian.list \
```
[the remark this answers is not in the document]
to-bar: Image size is almost the same (1.051GB vs 1.045GB) but agree about the build time 👍

Dockerfile
to-bar: Good point, thanks! 👍

sk4zuzu (approved):
o/\o

atsikham (approved):
LGTM

Issue #1520, opened 2020-08-03
[BUG] Kubectl client outside of HA/multi-master Epiphany cluster fails to connect to server with invalid certificate


ks4225:
**Describe the bug**
On a HA / multi-master, issuing `kubectl` commands from a machine outside the cluster (e.g. CI agent) will sometime fail with a certificate error. The thought is that the HAProxy on the k8s master machines ends up routing the `kubectl` in a way that mismatches with the config on the external machine.

**To Reproduce**
1. Build an Epiphany cluster with HA / multi-master (3 masters in this case)
2. Copy the kube config from one of the k8s master machines to an external machine (as part of this, `localhost` needs to be replaced in the kube config)
3. Issue `kubectl` commands from the external machine, which will fail periodically (depending on how traffic is routed)

**Expected behavior**
It should be possible to issue `kubectl` commands from the external machine that work consistently.

**Config files**
Key aspects of the config are:
```
components:
    kubernetes_master:
      count: 3
...
use_ha_control_plane: true
```

**OS (please complete the following information):**
 - OS: Ubuntu

**Cloud Environment (please complete the following information):**
 - Cloud Provider: MS Azure

**Additional context**

cc @jsmith085 @sunshine69

ks4225:
Example error message is:
`Unable to connect to the server: x509: certificate is valid for ###, ###, not ###` (where ### are IPs)

sk4zuzu:
Thank you for reporting the issue, @ks4225 !

I've checked that indeed in non-HA and HA clusters the kubeconfig handling differ.

I believe, two things need to be done to fix the problem:
1. In new clusters - during the `kubeadm init` run we need to provide addtional cert SANs in the kubeadm config file/map.
2. In exsting clusters that aleady have this issue - we need to modify the config map and regenerate certificates.

All this should be done during the `epicli apply` run.

As a temporary workaround some kind of tcp proxy can be used, for example:
```bash
$ ssh -L 3446:localhost:3446 ubuntu@10.20.2.10 -N
```
```bash
$ kubectl --kubeconfig admin.conf get nodes,pods -A
NAME        STATUS   ROLES    AGE     VERSION
node/x1a1   Ready    master   58m     v1.18.6
node/x1a2   Ready    master   10m     v1.18.6
node/x1a3   Ready    master   9m12s   v1.18.6
node/x1b1   Ready    <none>   56m     v1.18.6

[...]

PR #1626, opened 2020-09-08
Use pki config from manifest for apiserver certificates

atsikham:
These changes are related to #1556, but as I knew about this configuration parameter only during #1595 implementation, PR is created now.


mkyc (commented):
Is this related to some task?

atsikham:
@mkyc it could be #1520, but it's not there because of the reason written in description. So I decided to fix that without any task.

mkyc (approved):
LGTM

sk4zuzu (approved):
LGTM

PR #1627, opened 2020-09-08
Fixed autoscaler configuration

mum0l:
- account_id and autoscaler_name put into vars.
- resource name adjusted.


mum0l:
Wrong branch.

PR #1631, opened 2020-09-08
Open network ports for K8s api.

seriva:


rpudlowski93 (approved):
LGTM, in case of AKS - 443 (API) port is open by default.

Issue #1659, opened 2020-09-16
[BUG] Upgrade process fails on Ubuntu with package downgrade


atsikham:
**Describe the bug**
Upgrade fails on Ubuntu. Related log part:

```text
{"asctime": "15:54:02", "levelname": "INFO", "name": "cli.engine.ansible.AnsibleCommand", "message": "TASK [upgrade : k8s/install | Get information about installed packages as facts] ********************************************************"}
{"asctime": "15:54:05", "levelname": "INFO", "name": "cli.engine.ansible.AnsibleCommand", "message": "ok: [ec2-3-95-156-60.compute-1.amazonaws.com]"}
{"asctime": "15:54:05", "levelname": "INFO", "name": "cli.engine.ansible.AnsibleCommand", "message": ""}
{"asctime": "15:54:05", "levelname": "INFO", "name": "cli.engine.ansible.AnsibleCommand", "message": "TASK [upgrade : k8s/install | Include unhold packages task] *****************************************************************************"}
{"asctime": "15:54:05", "levelname": "INFO", "name": "cli.engine.ansible.AnsibleCommand", "message": "included: /workspaces/epiphany/core/src/epicli/clusters/rabbit/build/rabbit/ansible/roles/upgrade/tasks/kubernetes/Debian/unhold-packages.yml for ec2-3-95-156-60.compute-1.amazonaws.com"}
{"asctime": "15:54:05", "levelname": "INFO", "name": "cli.engine.ansible.AnsibleCommand", "message": ""}
{"asctime": "15:54:05", "levelname": "INFO", "name": "cli.engine.ansible.AnsibleCommand", "message": "TASK [upgrade : k8s/install | Unhold packages: kubelet, kubectl, kubernetes-cni] ********************************************************"}
{"asctime": "15:54:08", "levelname": "INFO", "name": "cli.engine.ansible.AnsibleCommand", "message": "changed: [ec2-3-95-156-60.compute-1.amazonaws.com]"}
{"asctime": "15:54:08", "levelname": "INFO", "name": "cli.engine.ansible.AnsibleCommand", "message": ""}
{"asctime": "15:54:08", "levelname": "INFO", "name": "cli.engine.ansible.AnsibleCommand", "message": "TASK [upgrade : k8s/install | Remove newer Debian packages installed as dependencies if they exist] *************************************"}
[...]

PR #1660, opened 2020-09-16
Used list value for apt module parameter

atsikham:
This PR is related to the bug #1659.
With value change, also added `kubeadm` package there, as it's a dependent one. Probably it's possible not to add `kubeadm`,  just to add force removal, but as Ansible [documentation](https://docs.ansible.com/ansible/latest/modules/apt_module.html) states, `This is a destructive operation with the potential to destroy your system, and it should almost never be used`.

### DIFF
--- a/core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/kubernetes/Debian/install-packages.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/kubernetes/Debian/install-packages.yml
@@ -10,14 +10,20 @@
   vars:
     packages: >-
       {%- if ansible_facts.packages['kubernetes-cni'] is defined -%}
-        ['kubelet', 'kubectl', 'kubernetes-cni']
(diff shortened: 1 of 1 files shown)


sk4zuzu (approved):
LGTM

to-bar:
In upgrade role we upgrade `kubeadm` package first (`install-kubeadm.yml`), run some tasks, then upgrade other K8s packages (`install-packages.yml`/`install-packages-cni-in-kubelet.yml`). We don't want to remove previously upgraded `kubeadm`.

I think the better solution is to not install the latest version of `kubernetes-cni` (here `0.8.7-00`) in order to not downgrade it later.

In kubernetes_common role we manage versions of:
- kubelet={{specification.version}}-00
- kubectl={{specification.version}}-00
- kubeadm={{specification.version}}-00

I would opt for adding to the above list also `kubernetes-cni` with a fixed version (in sync with `requirements.txt`).

to-bar:
Closing since more changes have been done in #1666.

PR #1665, opened 2020-09-16
Remove local kubeconfig in postflight

sk4zuzu:
Slight local kubeconfing handling change before refactoring "applications" role.

### DIFF
--- a/core/src/epicli/cli/engine/ansible/AnsibleRunner.py
+++ b/core/src/epicli/cli/engine/ansible/AnsibleRunner.py
@@ -78,11 +78,12 @@ class AnsibleRunner(Step):
         self.ansible_command.run_playbook(inventory=inventory_path,
                                           playbook_path=self.playbook_path('common'))
 
-
     def post_flight(self, inventory_path):
         self.ansible_command.run_playbook(inventory=inventory_path,
                                           playbook_path=self.playbook_path('repository_teardown'))
 
+        self.ansible_command.run_playbook(inventory=inventory_path,
+                                          playbook_path=self.playbook_path('postflight'))
 
     def apply(self):
         inventory_path = get_inventory_path(self.cluster_model.specification.name)
@@ -115,7 +116,6 @@ class AnsibleRunner(Step):
 
         #post-flight after we are done
         self.post_flight(inventory_path)
-
 
     def upgrade(self):
         inventory_path = get_inventory_path_for_build(self.build_dir)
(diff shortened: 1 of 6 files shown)


atsikham:
@sk4zuzu do we need to create `~/.kube/config` on remote? As I understand `/etc/kubernetes/admin.conf` is used everywhere.

core/src/epicli/cli/engine/ansible/AnsibleRunner.py
```diff
@@ -78,11 +78,12 @@ def pre_flight(self, inventory_path):
         self.ansible_command.run_playbook(inventory=inventory_path,
                                           playbook_path=self.playbook_path('common'))
 
-
     def post_flight(self, inventory_path):
         self.ansible_command.run_playbook(inventory=inventory_path,
                                           playbook_path=self.playbook_path('repository_teardown'))
 
+        self.ansible_command.run_playbook(inventory=inventory_path,
+                                          playbook_path=self.playbook_path('postflight'))
```
mkyc, line 86: I would suggest to rename this playbook to explicit "clean local kubeconf" or something similar. When I read line 83 it's quite clear what `repository_teardown` is doing, but it's not clear at all at this level what will `postflight` do. Other option in my opinion would be to move `repository_teardown` logic into `postflight` playbook.

mkyc (approved):
Small suggestion but generally LGTM.

core/src/epicli/cli/engine/ansible/AnsibleRunner.py
sk4zuzu: So the idea is to introduce similar task to preflight but just called postflight

> @sk4zuzu do we need to create `~/.kube/config` on remote? As I understand `/etc/kubernetes/admin.conf` is used everywhere.

Not really, but this is just a first PR in the chain. I will probably do what you suggest in the nex one.

sk4zuzu:
> @sk4zuzu do we need to create `~/.kube/config` on remote? As I understand `/etc/kubernetes/admin.conf` is used everywhere.

This is just a first PR in the chain. I will probably do what you suggest in the next one :+1:

atsikham (approved):
LGTM

core/src/epicli/cli/engine/ansible/AnsibleRunner.py
sk4zuzu: So we have a preflight task. but we dont have a postflight task. :D

I think instead of introducing more and more files in the main playbook directory I placed this in a special role that is supposed to run over all hosts in the inventory and do smaller cleanup-related tasks.

So if you look inside the role you'll see a clear distinction. I'd like to keep that. 🤗

mkyc: As you wish, mister 😝 👍

PR #1663, opened 2020-09-16
SSH tunnel for using kubectl from container for self-managed K8s clusters

seriva:
Creates an SSH tunnel so we can use Kubectl through localhost:port when we have a cluster that uses our own K8s cluster.

PR #1655, opened 2020-09-14
Use local kubectl for k8s deployment

atsikham:
#1615, #1616 and #1659 have to be implemented before merge
Verified by manual execution of @seriva [poc](https://github.com/epiphany-platform/epiphany/issues/1616#issuecomment-691019357) before applying configuration.

It's supposed that `kubeconfig` is located under `inventory_dir`. KUBECONFIG is specified on the play level and defined in `group_vars`.

**Untouched:**

- applications - separate task #1654 
- vault - separate task #1672 
- usage of `kubectl` with `kubelet.conf` as `kubelet.conf` is located only on remote hosts and tasks relates to these hosts:
:small_orange_diamond: core/src/epicli/data/common/ansible/playbooks/roles/kubernetes_common/tasks/gather-facts.yml
:small_orange_diamond: core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/kubernetes/verify-upgrade.yml
- bash completion setup on remote host
:small_orange_diamond: core/src/epicli/data/common/ansible/playbooks/roles/kubernetes_master/tasks/main.yml
- preflight
:small_orange_diamond: core/src/epicli/data/common/ansible/playbooks/roles/preflight_facts/tasks/kubernetes.yml. Explanation is [here](https://github.com/epiphany-platform/epiphany/pull/1663#discussion_r490091462)

**Verified locally with Ubuntu+AWS configuration:**

- apply:
:white_check_mark: single control plane node
:white_check_mark: promotion to HA
:white_check_mark: HA k8s cluster
:white_check_mark: k8s cluster with worker nodes
:white_check_mark: k8s + monitoring (prometheus) and rabbitmq
:white_check_mark: certificates renewal
- upgrade:
:white_check_mark: k8s with included #1659 fix
:white_check_mark: k8s + monitoring (prometheus) and rabbitmq
- backup
:x: backup with following configuration 

```yaml
---
kind: configuration/backup
title: Backup Config
name: default
specification:
  components:
    load_balancer:
      enabled: false
    logging:
      enabled: false
    monitoring:
      enabled: true
    postgresql:
      enabled: false
    rabbitmq:
      enabled: true
    kubernetes:
      enabled: true
```

Backup does not seem to work correctly itself. With HA cluster got such an error, at least etcd image name is not processed to have value only from one pod:

```text
16:43:33 INFO cli.engine.ansible.AnsibleCommand - TASK [backup : Save etcd snapshot] ******************************************************************************************************
[...]

PR #1689, opened 2020-09-22
Local kubectl for prometheus and kubectl usage refactoring

atsikham:
In #1655 it was decided not to delegate usage of `kubectl` to localhost for kubernetes roles as there are a lot of places where it can't be changed:

- usage of `kubectl` with `kubelet.conf` that is located only on remote hosts and tasks relate to these hosts:
:small_orange_diamond: core/src/epicli/data/common/ansible/playbooks/roles/kubernetes_common/tasks/gather-facts.yml
:small_orange_diamond: core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/kubernetes/verify-upgrade.yml
- bash completion setup on remote host
:small_orange_diamond: core/src/epicli/data/common/ansible/playbooks/roles/kubernetes_master/tasks/main.yml
- preflight
:small_orange_diamond: core/src/epicli/data/common/ansible/playbooks/roles/preflight_facts/tasks/kubernetes.yml. Explanation is [here](https://github.com/epiphany-platform/epiphany/pull/1663#discussion_r490091462)
- master-init 
:small_orange_diamond: most of tasks in `master-init.yml`, as `kubernetes-secrets.yml` is used in `copy-cubeconfig.yml`, where local kubeconig is generated, but `kubernetes-secrets.yml` is created almost in the end of `master-init.yml` during cluster deployment

So the size of previous PR was reduced, main changes are:
* Usage of local kubectl for prometheus
* Refactoring to use kubeconfig location value from `group_vars`, mostly set on the play level

### DIFF
--- a/core/src/epicli/data/common/ansible/playbooks/roles/prometheus/tasks/main.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/prometheus/tasks/main.yml
@@ -11,11 +11,14 @@
   set_fact:
     does_k8s_exists: "(groups['kubernetes_master'] | default([])) | length > 0"
 
-- include: configure-k8s-apps-monitoring.yml
+- name: Configure kubernetes monitoring
+  when: does_k8s_exists
   become: true
+  include: configure-k8s-apps-monitoring.yml
+  environment:
+    KUBECONFIG: "{{ kubeconfig.local }}"
   tags:
     - kubernetes
-  when: does_k8s_exists
 
 - name: Gather variables for each operating system
   include_vars: "{{ item }}"
--- a/core/src/epicli/data/common/ansible/playbooks/roles/prometheus/tasks/main.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/prometheus/tasks/main.yml
@@ -13,10 +13,13 @@
 
 - name: Configure kubernetes monitoring
   when: does_k8s_exists
-  become: true
-  include: configure-k8s-apps-monitoring.yml
-  environment:
-    KUBECONFIG: "{{ kubeconfig.local }}"
+  include_tasks:
+    file: configure-k8s-apps-monitoring.yml
+    apply:
+      environment:
+        KUBECONFIG: "{{ kubeconfig.local }}"
+      tags:
+        - kubernetes
   tags:
     - kubernetes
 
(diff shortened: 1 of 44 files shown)


core/src/epicli/data/common/ansible/playbooks/roles/prometheus/tasks/main.yml
```diff
@@ -11,11 +11,14 @@
...
 
-- include: configure-k8s-apps-monitoring.yml
+- name: Configure kubernetes monitoring
+  when: does_k8s_exists
   become: true
+  include: configure-k8s-apps-monitoring.yml
+  environment:
+    KUBECONFIG: "{{ kubeconfig.local }}"
   tags:
     - kubernetes
```
to-bar, line 21:
```suggestion
- name: Configure Kubernetes monitoring
   when: does_k8s_exists
   include_tasks: configure-k8s-apps-monitoring.yml
   apply:
     environment:
       KUBECONFIG: "{{ kubeconfig.local }}"
     delegate_to: localhost
     become: false
     run_once: true
     tags:
        - kubernetes
  tags:
    - kubernetes
```
`delegate_to`, `become` & `run_once` not checked in this context, just idea to test.

atsikham: There is not only `kubectl` inside, but also setting facts for example, so I suggest don't apply these changes

to-bar: Updated my suggestion, kept only  `environment` and `include` -> `include_tasks`.  The `include` module is still not deprecated but it's a good idea to avoid using it, see [here](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/include_module.html#notes).

atsikham: May be

```yaml
- name: Configure kubernetes monitoring
  when: does_k8s_exists
  include_tasks:
    file: configure-k8s-apps-monitoring.yml
    apply: 
      environment:
        KUBECONFIG: "{{ kubeconfig.local }}"
      tags:
        - kubernetes
  tags:
    - kubernetes
```

to-bar (approved):

seriva (approved):

Issue #1691, opened 2020-09-23
[FEATURE REQUEST] Run HAProxy using dedicated user instead of root


to-bar:
**Is your feature request related to a problem? Please describe.**
Before changes from #1586 we had HAProxy installed as OS package (deb/rpm). The package automatically created OS user and group `haproxy`.

We had automated tests to check the following:

- User "haproxy" is expected to exist
- User "haproxy" is expected to have login shell "/usr/sbin/nologin" or have login shell "/sbin/nologin"
- Group "haproxy" is expected to exist
- User "haproxy" is expected to belong to group "haproxy"
- User "haproxy" is expected to have home directory "/var/lib/haproxy"

Now there is no `haproxy` user.

```
# grep haproxy /etc/shadow
haproxy_exporter:!:18528::::::
```

```
# grep -A3 user /opt/runc/haproxy/config.json
"user": {
    "gid": 0,
    "uid": 0
}
```

**Describe the solution you'd like**
Add `haproxy` user and run the `haproxy` service as this user.

**Describe alternatives you've considered**
No alternatives.

**Additional context**
As a part of this task we should restore automated tests removed in #1692.

---

**DoD checklist**

- Changelog
  - [ ] updated
  - [ ] not needed
- COMPONENTS.md
  - [ ] updated
  - [ ] not needed
- Schema
  - [ ] updated
  - [ ] not needed
- Backport tasks
  - [ ] created
  - [ ] not needed
- Documentation
  - [ ] added
  - [ ] updated
  - [ ] not needed
- [ ] Feature has automated tests
- [ ] Automated tests passed (QA pipelines)
  - [ ] apply
  - [ ] upgrade
  - [ ] backup/restore
- [ ] Idempotency tested
- [ ] All conversations in PR resolved
- [ ] Solution meets requirements and is done according to design doc
- [ ] Usage compliant with license

seriva:
Not going todo anymore feature requests for Epiphany.

Issue #1316, opened 2020-06-03
Upgrade HAProxy to v2.2


plirglo:
**Is your feature request related to a problem?** 
Currently HAProxy is in version 1.8 which is quite old (released 26.11.2017) with some bugs.
**Describe the solution you'd like**
HAProxy in the latest stable version at least : 2.2
Source: http://www.haproxy.org/
No major issues reported regarding this version: http://www.haproxy.org/bugs/bugs-2.2.2.html

**Describe alternatives you've considered** 

**Additional context** 
It [haproxy] is provided as a single executable file which doesn't
require any installation process. Multiple versions can easily coexist, meaning
that it's possible (and recommended) to upgrade instances progressively by
order of importance instead of migrating all of them at once. Configuration
files are easily versioned. Configuration checking is done off-line so it
doesn't require to restart a service that will possibly fail. During
configuration checks, a number of advanced mistakes may be detected (e.g. a rule
hiding another one, or stickiness that will not work) and detailed warnings and
configuration hints are proposed to fix them. Backwards configuration file
compatibility goes very far away in time, with version 1.5 still fully
supporting configurations for versions 1.1 written 13 years before, and 1.6
only dropping support for almost unused, obsolete keywords that can be done
differently. The configuration and software upgrade mechanism is smooth and non
disruptive in that it allows old and new processes to coexist on the system,
each handling its own connections. System status, build options, and library
compatibility are reported on startup.

to-bar:
Created related issue #1691.

ar3ndt:
LGTM

PR #1692, opened 2020-09-24
Remove tests for haproxy user

to-bar:
Removes tests for `haproxy` user since this feature is not implemented yet (after upgrading HAProxy to v2.2 #1316), see #1691.


atsikham (approved):
LGTM

przemyslavic (approved):

PR #1693, opened 2020-09-24
Removed examples as its moved to a new repository.

seriva:
Removed examples folder as its moved to a new repository here:

https://github.com/epiphany-platform/examples

Reasons for moving:

- To simplify blackduck opensource scan. The examples are not part of anything we officially release but do cause a lot of issues during scans of all the thirdparty things included for the different languages and tech stacks.
- So we don`t have any unnecessary security warnings on the Epiphany repo as they are often related to dependencies from the examples.


mkyc (approved):
👍

atsikham (approved):
LGTM

atsikham (changes requested):
I suggest to add a link somewhere in the doc to the new repository.

PR #1703, opened 2020-09-25
Get rid of truncates single quotes in yaml file

rafzei:


atsikham (approved):
LGTM

core/src/epicli/data/common/defaults/configuration/filebeat.yml
```diff
@@ -5,6 +5,7 @@ specification:
   filebeat_version: "7.8.1"
   postgresql_input:
     multiline:
-      pattern: '^\d{4}-\d{2}-\d{2} '
+      pattern: >-
```
sk4zuzu, line 8: This is not the cleanest solution, but I guess it's better than nothing. :+1:

sk4zuzu (approved):

Issue #1702, opened 2020-09-25
[BUG] Reruning epicli may fail for clustered Postgres


sunshine69:
**Describe the bug**
 Re run the deployment causing fail in postgres roles

See it in version 0.6.0. and expect same behaviour in 0.7.0 because that postgres role does not change.

**To Reproduce**

Build a cluster with postgres with two nodes

Observe that vm-1 is set as primary and vm-0 is hot standby

Re run the deployment with nothing changes 

epcli run failed when applying postgres role.

**Expected behavior**
It should be no errors

**Config files**

**OS (please complete the following information):**
 - OS: Ubuntu 18.04

**Cloud Environment (please complete the following information):**
 - Cloud Provider MS Azure

**Additional context**

The reason is the that the role uses the condition `groups['postgresql'][0] == inventory_hostname` to decide which host is primary. The first run the condition is resolved to vm-1.

However the second run it resolved to vm-0 and because vm-0 is already setup as standby the task failed.

Below is the log case

First run 

https://abb-jenkins.duckdns.org:8080/view/Development/job/DEPLOY-de-cluster/433/console

it picks master is vm-1 

```
02:23:07 INFO cli.engine.ansible.AnsibleCommand - TASK [postgresql : Check if master is already registered in repmgr] ************
02:23:07 INFO cli.engine.ansible.AnsibleCommand - skipping: [de-stdbase-postgresql-vm-0]
02:23:07 INFO cli.engine.ansible.AnsibleCommand - ok: [de-stdbase-postgresql-vm-1]
```

epicli postgres role in replication-repmgr-Debian.yml

```
# Master:
- name: Check if master is already registered in repmgr
  become_user: postgres
  shell: >-
    set -o pipefail &&
    {{ repmgr_bindir[ansible_os_family] }}/repmgr cluster show -f {{ repmgr_config_dir[ansible_os_family] }}/repmgr.conf | grep -i {{ inventory_hostname }} | grep -v standby
  changed_when: false
  register: is_master_already_registered
  failed_when: is_master_already_registered.rc == 2
  args:
    executable: /bin/bash
  when:
    - groups['postgresql'][0] == inventory_hostname
```

Now re-run it.

https://abb-jenkins.duckdns.org:8080/view/Development/job/DEPLOY-de-cluster/434/console

```
06:15:30 INFO cli.engine.ansible.AnsibleCommand - TASK [postgresql : Check if master is already registered in repmgr] ************
[...]

PR #1706, opened 2020-09-27
Deterministic hosts order in hostgroup

sunshine69:
Tested by me. The behavior is now expected order of hosts in host group. The order is sorted by hostname as string sort and does not change in multiple re runs.


przemyslavic:
@mkyc @seriva @to-bar 
I performed some testing on this PR. It works fine for `applying` and `re-applying` cluster configuration (tested on Azure and AWS).
Additionally it solves the issue #1702 on Azure. However, there may still be issues on AWS scaling components up, depending on which position the new host goes to.
I would introduce the change for Azure provider, and not necessarily for AWS.

mkyc:
@przemyslavic can you please check upgrade scenario?

mkyc:
We are closing it since its implemented in #2733

PR #1712, opened 2020-09-29
Fix kubeconfig generation logic

sk4zuzu:
The issue here was the invalid order of operations:
1. try creating local kubeconfig (depends on 3.)
2. deploy first master (kubeadm init)
3. generate secrets (1. depends on it)
4. if kubernetes_node group is not empty then retry 1.

:DDD

When single_machine config is used here 4. never happens. :-1:

I also removed legacy `.kube/config` generation as it had been removed from upgrades already and caused us so much pain and sorrow. :sob:

### DIFF
--- a/core/src/epicli/data/common/ansible/playbooks/kubernetes_master.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/kubernetes_master.yml
@@ -8,6 +8,9 @@
   become_method: sudo
   pre_tasks:
     - import_role:
+        name: kubernetes_master
+        tasks_from: noop  # just to include role vars
+    - import_role:
         name: docker
         tasks_from: main
     - import_role:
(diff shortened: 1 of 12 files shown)


core/src/epicli/data/common/ansible/playbooks/kubernetes_master.yml
```diff
@@ -7,15 +7,15 @@
   become: true
   become_method: sudo
   pre_tasks:
+    - import_role:
+        name: kubernetes_master
+        tasks_from: noop  # just to include role vars
```
przemyslavic, line 12: Wouldn't it be better to use the include_vars module?

sk4zuzu: Please take a look a latest commits :+1:

to-bar (approved):

przemyslavic (approved):

Issue #1322, opened 2020-06-03
 Add proper documentation how to define additional security rules for AWS/Azure


seriva:
Currently there is no documentation how to manage security rules for both AWS and Azure. In most cases teams are adding them manually through the clouds web interfaces or APIs resulting in Terraform wiping them out after re-running Epiphany.

Documentation (with example) should be written for both AWS and Azure showing:

- Change existing default rules
- How to add/extend rules

And should show how to do it so it does not break Epiphany features and support (e.g. still allows use of Epiphany tools to scale out the cluster, update it. etc.

mkyc:
looks like related to #896

TolikT:
Defaults:

- https://github.com/epiphany-platform/epiphany/blob/develop/core/src/epicli/data/aws/defaults/infrastructure/virtual-machine.yml
- https://github.com/epiphany-platform/epiphany/blob/develop/core/src/epicli/data/azure/defaults/infrastructure/virtual-machine.yml

Documentation should display with examples how to override/change these settings.

mkyc:
We should add task (or check if there is one already) to adjust current defaults.

mum0l:
Please note, that I stuck with this topic. 

Cannot set the additional security rules in AWS, as none of configs provided by colleagues is working. If I need to document the procedure, first I need to have working solution (working yaml deployment file). Without that, it might be really hard to document, something what is not working

seriva:
Reviewed provided configuration: Azure was fine, AWS had port assigned to wrong field. Outstanding pull request to simplify the defaults and make them less confusing: #1714

mkyc:
Is there any PR to be connected to this issue?

mum0l:
@mkyc yes it is: https://github.com/epiphany-platform/epiphany/pull/1717

seriva:
Tested with the configurations provided and everything works as expected.

PR #1714, opened 2020-09-30
Removed unused security rules fields.

seriva:
Cleanup for the documentation task #1322 regarding changing/adding security rules:

- Removes unused fields from the defaults for both Azure and AWS
- Fixes issue with duplicate security rules in AWS.

AWS security rule definition uses the fieldnames as for Azure. Ideally in the future we should match them to the origninal Terraform, templates. For now I left it as is to make it backwards compatible with older version of Epiphany.


pyrkamarcin (dismissed):

pyrkamarcin (dismissed):

core/src/epicli/cli/engine/providers/aws/InfrastructureBuilder.py
```diff
@@ -84,7 +84,9 @@ def run(self):
             autoscaling_group = self.get_autoscaling_group(component_key, component_value, subnets_to_create, asg_index)
 
             for security_group in security_groups_to_create:
-                security_group.specification.rules += autoscaling_group.specification.security.rules
+                for rule in autoscaling_group.specification.security.rules:
+                    if self.rule_exists_in_list(security_group.specification.rules, rule) == False:
```
atsikham, line 88: May be `if not`?

core/src/epicli/data/aws/terraform/infrastructure/security-group.j2
```diff
@@ -15,33 +15,30 @@ resource "aws_security_group" "{{ specification.name }}" {
...
+    cidr_blocks = ["{{ rule.source_address_prefix }}"]
+  {%- else %}
+    cidr_blocks = ["{{ rule.destination_address_prefix }}"]
+  {%- endif %}
+  }
 
-}
 {%- endfor %}
-
 }
```
atsikham, line 44: Newline

atsikham (changes requested):
In general some day I would like to see code formatted according to PEP8. Not sure if vscode can do this, but it's for sure possible in pycharm.

LGTM, but a few changes requested.

core/src/epicli/cli/engine/providers/aws/InfrastructureBuilder.py
seriva: Fixed

core/src/epicli/data/aws/terraform/infrastructure/security-group.j2
seriva: fixed

seriva:
Yes we can set up PEP8 in VSCode easily, I am pretty sure its even in the backlog somewhere.

atsikham (approved):

erzetpe (approved):

PR #1717, opened 2020-10-01
Security groups documentation added

mum0l:


docs/design-docs/security-groups/SECURITY-GROUPS.md
```diff
@@ -0,0 +1,354 @@
...
+As we see, on this example, we are seting new rule name "prometheus", with priority 302, which is allowing accesses from local network "10.1.0.0/20" into Prometheus application host on port 9090. Note, that here we set dest "0.0.0.0/0" address and source 10.1.0.0/20, as those are the default configurations in Epiphany used if noone specify others. This might be adjusted in case you're using difrent addresses. 
+
+
+### Setting own security groups
+Sometimes, there is a need to set additional security rules for some application which we're deploying in epiphany kubernetes cluster. Than, we need to stick into following rules:
+- Whenever we want to add new rule - for example open port "222", we should *COPY* all current roles into our deployment .yaml file, and at the end, add the rule which we want.
+- Each component has his own rule-set, so we need to be very carefull where we're putting them.
+- After adding new rules, and infra part is done (terraform), we can go into terraform build directory and check if fiiles contain our port definition.
+
+###### Example
```
rafzei, line 52: \*\*Example:**

docs/design-docs/security-groups/SECURITY-GROUPS.md
```diff
@@ -0,0 +1,354 @@
...
+  destination_port_range: "9090"
+  source_address_prefix: "10.1.0.0/20"
+  destination_address_prefix: "0.0.0.0/0"
+```
+As we see, on this example, we are seting new rule name "prometheus", with priority 302, which is allowing accesses from local network "10.1.0.0/20" into Prometheus application host on port 9090. Note, that here we set dest "0.0.0.0/0" address and source 10.1.0.0/20, as those are the default configurations in Epiphany used if noone specify others. This might be adjusted in case you're using difrent addresses. 
+
+
+### Setting own security groups
+Sometimes, there is a need to set additional security rules for some application which we're deploying in epiphany kubernetes cluster. Than, we need to stick into following rules:
+- Whenever we want to add new rule - for example open port "222", we should *COPY* all current roles into our deployment .yaml file, and at the end, add the rule which we want.
```
rafzei, line 48: /n after line 47

docs/design-docs/security-groups/SECURITY-GROUPS.md
```diff
@@ -0,0 +1,354 @@
...
+- Each component has his own rule-set, so we need to be very carefull where we're putting them.
+- After adding new rules, and infra part is done (terraform), we can go into terraform build directory and check if fiiles contain our port definition.
+
+###### Example
+Please check bellow example, how to setup basic epiphany cluster in AWS with 1 master, 2 nodes, and open accesses to all hosts on port 10051 from monitoring network.
+
+```
```
rafzei, line 55: code blocks should have a language specified
like \```yaml

rafzei (changes requested):
I've put down some comments from markdownlint, please use it in your vsc.

atsikham:
Shouldn't the target branch be `develop`?

mum0l:
> Shouldn't the target branch be `develop`?

Spoke with Pawel, and agreed. Branch changed to develop now.

seriva (changes requested):

docs/design-docs/security-groups/SECURITY-GROUPS.md
```diff
@@ -0,0 +1,360 @@
...
+  direction: Inbound
+  access: Allow
+  protocol: Tcp
+  source_port_range: "*"
+  destination_port_range: "9090"
+  source_address_prefix: "10.1.0.0/20"
+  destination_address_prefix: "0.0.0.0/0"
+```
+
+As we see, on this example, we are seting new rule name "prometheus", with priority 302, which is allowing accesses from local network "10.1.0.0/20" into Prometheus application host on port 9090. Note, that here we set dest "0.0.0.0/0" address and source 10.1.0.0/20, as those are the default configurations in Epiphany used if noone specify others. This might be adjusted in case you're using difrent addresses.
```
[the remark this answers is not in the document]
rafzei: I would propose "In this example" or "In the above example".

docs/design-docs/security-groups/SECURITY-GROUPS.md
```diff
@@ -18,11 +18,9 @@ Rules description:
 ```yaml
 - name:                       "Name of the rule"
   description:                "Short rule description"
-  priority:                   "Rule priority; which rules should be considered as first-used basing on lowest number"
   direction:                  "Inbound || Egress" - which direction are you allowing rule"
-  access:                     "Allow || Deny" - whenever rule should allow connections or block"
   protocol:                   "TCP || UDP" - which protocol should be used for connections"
-  destination_port_range:     "Destination prot ranges"
+  destination_port_range:     "Destination prot"
```
atsikham:
```suggestion
  destination_port_range:     "Destination port"
```

atsikham (changes requested):

docs/design-docs/security-groups/SECURITY-GROUPS.md
```diff
@@ -0,0 +1,364 @@
...
+name: default
+provider: azure
+title: Epiphany cluster Config
+specification:
+  name: azure
+  prefix: azure
+  admin_user:
+    name: operations
+    key_path:  /workspaces/epiphany/core/src/epicli/clusters/keys/abb_rsa
+  cloud:
```
rpudlowski93, line 76: It could be nice to add the latest change in epiphany config file for v0.8.0
```
      cloud:
        k8s_as_cloud_service: false # <----- make sure that flag is set, as it indicates usage of AKS service
        .... 
```

mum0l:
Everything fixed. @atsikham @seriva @rpudlowski93 @rafzei - could you take a look one more time?

docs/design-docs/security-groups/SECURITY-GROUPS.md
seriva: Default is false so don't think its needed for the this.

seriva (approved):

rafzei (approved):

atsikham (approved):

PR #1757, opened 2020-10-08
Prepare release 0.8.0-rc.0

to-bar:


seriva (dismissed):

core/src/epicli/cli/version.txt.py
```diff
@@ -1 +1 @@
-0.7.1
\ No newline at end of file
+0.8.0-rc.0
```
mkyc, line 1: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_206

mkyc (changes requested):

mkyc (approved):

PR #1769, opened 2020-10-14
Do not fail on Kibana upgrade when legacy version is present

to-bar:
Kibana is upgraded only when there is no 'kibana-oss' package (replaced by 'opendistroforelasticsearch-kibana' since v0.5).
This condition has been added to not fail when 'epicli upgrade' is run for Epiphany v0.4 cluster.
We cannot upgrade Kibana to v7 having Elasticsearch in legacy version (v6).


rafzei (approved):
LGTM

seriva (approved):

PR #1807, opened 2020-10-28
Fix postgres hot_standby setting

to-bar:
- Set hot_standby via postgresql-epiphany.conf
- Adjust tests for hot_standby setting
- Do not run PGAudit tests when there is no logging host
- Fix Ruby constants deprecation warnings
- Move section with legacy replication down
- Fix setting paths for legacy replication in docs
- Treat PostgreSQL native replication as deprecated

### DIFF
--- a/CHANGELOG-0.9.md
+++ b/CHANGELOG-0.9.md
@@ -6,6 +6,8 @@
 
 ### Fixed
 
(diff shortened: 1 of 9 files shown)


atsikham (approved):
LGTM

przemyslavic (approved):

Issue #1833, opened 2020-11-05
[BUG] DaemonSets of Node Exporter and Filebeat deploy in default namespace


rpudlowski93:
**Describe the bug**
DaemonSets of Node Exporter and FileBeat in case of flag k8s_as_cloud_service: true, will deploy in default namespace which should be clear for user usages.

**To Reproduce**
Create Epiphany cluster with flag k8s_as_cloud_service: true

**Expected behavior**
Deploy DaemonSets of Node Exporter and FileBeat in custom namespaces (epi-monitoring and epi-logging)

**OS (please complete the following information):**
 - Ubuntu/Centos/RHEL

**Cloud Environment (please complete the following information):**
 - AWS/Azure

rpudlowski93:
Added draft of PR [link](https://github.com/epiphany-platform/epiphany/pull/1839)
- We need to discuss how the upgrade process should works for the patch

rpudlowski93:
- Changes done in PR #1839
- Now, the daemonsets will be installed in custom namespaces

to-bar:
Checked:
❌ `epicli upgrade` (for cluster in v0.8.0):
```
15:41:51 INFO cli.engine.ansible.AnsibleCommand - TASK [upgrade : Check if filebeat release exist in default namespace and set fact] *********************************************************************
15:41:51 ERROR cli.engine.ansible.AnsibleCommand - fatal: [tb-module-tests-all-1]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: 'specification' is undefined\n\nThe error appears to be in '/workspaces/epiphany/core/src/epicli/clusters/build/tmp/azepi/ansible/roles/upgrade/tasks/filebeat.yml': line 11, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - name: Check if filebeat release exist in default namespace and set fact\n      ^ here\n"}
```

rpudlowski93:
Additional changes done in PR #[1878](https://github.com/epiphany-platform/epiphany/pull/1878). Please repeat tests

przemyslavic:
Checked:
✔️ epicli upgrade (from 0.8.0 to develop):
```
kubectl get pods -A -o wide
NAMESPACE        NAME                                                  READY   STATUS    RESTARTS   AGE    IP           NODE                              NOMINATED NODE   READINESS GATES
[...]

Issue #1756, opened 2020-10-08
[FEATURE REQUEST] Separate role vars and manifest vars generation during upgrades


sk4zuzu:
**Is your feature request related to a problem? Please describe.**
In this PR https://github.com/epiphany-platform/epiphany/pull/1619 a helper mechanism was introduced to make it possible during upgrades to re-use role config docs taken from manifest.

Currenly it's not enabled for all roles. Also currently enabling manifest vars implies generation of default vars as well. This may cause some issues for certain roles during upgrades.

This FR is about removing this implication and separate those two functionalities to gain fine control over them. Also this separation will surely cause that the whole feature will become more attractive to developers as it may have some potential to simplify the upgrade process. :+1::innocent:

**Describe the solution you'd like**
Split this list https://github.com/epiphany-platform/epiphany/blob/develop/core/src/epicli/cli/engine/ansible/AnsibleVarsGenerator.py#L63 into two. :+1: 

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

toszo:
Currently, upgrade overwrites role vars (as far as I remember) and this FR will fix it? Do I understand that correctly?

sk4zuzu:
Yes, upgrade still overwrites role vars with defaults for specific roles, but we've also added an optional file `vars/manifest.yml` that is placed next to `vars/main.yml` (not instead of it). This extra file contains settings from the real manifest and is not loaded by default. Later (during upgrade) we can call a role with manifest vars instead of default vars (or just load them to a dictionary) if needed, like this:
```yaml
- name: load_balancer | Upgrade haproxy service (runc)
  include_role:
    name: haproxy
    vars_from: manifest
```
Haproxy component is a good example where this is handy as all the important config is provided by the user.

This FR is to improve this mechnism as currently if we want "manifest vars" we generate "default vars' as well. We'd like to separate those two things. :thinking:

sk4zuzu:
I'm sorry I got maybe a little ahead of myself, but the implementation was extremely short and straightforward #1942. I believe this could help with the grooming... :hand_over_mouth:

przemyslavic:
Looks good. Roles tested:
✅repository, image_registry, node_exporter, haproxy

PR #1839, opened 2020-11-09
Installing node exporter and filebeat as daemonsets in custom namespaces

rpudlowski93:
Fix related to bug : #1833 

- Installing Node Exporter and FileBeat as DaemonSets in custom namespaces.
- DaemonSets for k8s as cloud service

### DIFF
--- a/core/src/epicli/data/common/ansible/playbooks/roles/filebeat/defaults/main.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/filebeat/defaults/main.yml
@@ -1,3 +1,5 @@
 ---
 filebeat_helm_chart_file_name: filebeat-7.9.2.tgz
 filebeat_version: "7.9.2"
+# Use custom namespace for logging charts such as filebeat in case of k8s as cloud service.
+logging_chart_namespace: epi-logging
--- a/core/src/epicli/data/common/ansible/playbooks/roles/filebeat/tasks/install-filebeat-as-daemonset.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/filebeat/tasks/install-filebeat-as-daemonset.yml
@@ -26,7 +26,7 @@
 
     - name: Install Filebeat using custom Helm chart (custom-chart-values.yml)
       command: |
-        helm upgrade --install \
+        helm -n {{ logging_chart_namespace }} upgrade --install \
           -f {{ download_directory }}/custom-chart-values.yml \
           {{ specification.helm_chart_name }} \
-          {{ download_directory }}/{{ filebeat_helm_chart_file_name }}
+          {{ download_directory }}/{{ filebeat_helm_chart_file_name }} --create-namespace
--- a/core/src/epicli/data/common/ansible/playbooks/roles/node_exporter/tasks/uninstall-node-exporter-as-daemonset.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/node_exporter/tasks/uninstall-node-exporter-as-daemonset.yml
@@ -6,6 +6,4 @@
   become: false
   run_once: true
 
-  environment: { KUBECONFIG: "{{ vault_location }}/../kubeconfig" }
-
   command: helm uninstall {{ specification.helm_chart_name }}
--- a/core/src/epicli/data/common/ansible/playbooks/roles/node_exporter/tasks/uninstall-node-exporter-as-daemonset.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/node_exporter/tasks/uninstall-node-exporter-as-daemonset.yml
@@ -1,6 +1,6 @@
 ---
 
-- name: Uninstall Helm release "{{ filebeat_helm_chart_file_name }}"
+- name: Uninstall Helm release "{{ specification.helm_chart_name }}"
 
   delegate_to: localhost
   become: false
@@ -8,4 +8,4 @@
 
   environment: { KUBECONFIG: "{{ vault_location }}/../kubeconfig" }
 
-  command: helm uninstall {{ filebeat_helm_chart_file_name }}
+  command: helm uninstall {{ specification.helm_chart_name }}
--- a/core/src/epicli/data/common/ansible/playbooks/roles/node_exporter/tasks/uninstall-node-exporter-as-daemonset.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/node_exporter/tasks/uninstall-node-exporter-as-daemonset.yml
@@ -1,0 +1,11 @@
+---
+
+- name: Uninstall Helm release "{{ filebeat_helm_chart_file_name }}"
+
+  delegate_to: localhost
+  become: false
+  run_once: true
+
+  environment: { KUBECONFIG: "{{ vault_location }}/../kubeconfig" }
+
+  command: helm uninstall {{ filebeat_helm_chart_file_name }}
--- a/core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/filebeat.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/filebeat.yml
@@ -5,17 +5,21 @@
   run_once: true
   environment: { KUBECONFIG: "{{ vault_location }}/../kubeconfig" }
   command: helm list --output json
-  register: helm_list_releases
+  register: helm_list
 
-- name: Check if exist filebeat release in default namespace, if exist set the filebeat release name
+- name: Check if filebeat release exist in default namespace and set facts
   set_fact:
-    filebeat_release_name: "{{ item.name }}"
-  when: item.name == specification.helm_chart_name
-  with_items: "{{ (helm_list_releases.stdout|from_json) }}"
+    filebeat_release_exists: >-
+      {{ _names | ternary(true, false) }}
+  vars:
+    _names: >-
+      {{ helm_list.stdout|from_json | map(attribute='name')
+                  | select('eq', specification.helm_chart_name)
+                  | list }}
 
 - name: Reinstall Filebeat as DaemonSet from default namespace to custom namespace for "k8s as cloud service"
   when:
-    - filebeat_release_name is defined
+    - filebeat_release_exists
     - k8s_as_cloud_service is defined
     - k8s_as_cloud_service
   block: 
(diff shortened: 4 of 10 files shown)


core/src/epicli/data/common/ansible/playbooks/roles/filebeat/defaults/main.yml
```diff
@@ -1,3 +1,5 @@
 ---
 filebeat_helm_chart_file_name: filebeat-7.9.2.tgz
 filebeat_version: "7.9.2"
+# Use custom namespace for logging charts such as filebeat in case of k8s as cloud service.
+logging_chart_namespace: epi-logging
```
atsikham, line 5: Shouldn't it be configurable?

core/src/epicli/data/common/ansible/playbooks/roles/filebeat/tasks/install-filebeat-as-daemonset.yml
```diff
@@ -26,7 +26,7 @@
 
     - name: Install Filebeat using custom Helm chart (custom-chart-values.yml)
       command: |
-        helm upgrade --install \
+        helm -n {{ logging_chart_namespace }} upgrade --install \
           -f {{ download_directory }}/custom-chart-values.yml \
           {{ specification.helm_chart_name }} \
-          {{ download_directory }}/{{ filebeat_helm_chart_file_name }}
+          {{ download_directory }}/{{ filebeat_helm_chart_file_name }} --create-namespace
```
sk4zuzu, line 32: Just a remark. We have to be careful with the `--create-namespace` flag I believe, but for now it's in the official documentation so it should be good. :+1::innocent:

core/src/epicli/data/common/ansible/playbooks/roles/node_exporter/tasks/uninstall-node-exporter-as-daemonset.yml
```diff
@@ -0,0 +1,11 @@
+---
+
+- name: Uninstall Helm release "{{ specification.helm_chart_name }}"
+
+  delegate_to: localhost
+  become: false
+  run_once: true
+
+  environment: { KUBECONFIG: "{{ vault_location }}/../kubeconfig" }
```
atsikham, line 9: Recently we reduced the number of tasks where we use `KUBECONFIG` variable. Can't the value be taken from `group_vars/all.yml` and be specified on a playbook level? The same question for the left part of PR.

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/filebeat.yml
```diff
@@ -1,26 +1,50 @@
...
+  delegate_to: localhost
+  become: false
+  run_once: true
+  environment: { KUBECONFIG: "{{ vault_location }}/../kubeconfig" }
+  command: helm list
+  register: check_filebeat_release
+
+- name: Reinstall Filebeat as DaemonSet from default namespace to custom namespace for "k8s as cloud service"
+  when:
+    - specification.helm_chart_name in check_filebeat_release.stdout
```
sk4zuzu, line 12: Isn't `specification.helm_chart_name in check_filebeat_release.stdout` too wide of a condition? :thinking: What if user deploys a chart with a name that completely contains `specification.helm_chart_name`? :monkey:

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/filebeat.yml
```diff
@@ -1,26 +1,50 @@
...
+  block: 
+    - name: Include uninstall task for Filebeat as DaemonSet in default namespace for "k8s as cloud service"
+      include_role:
+        name: filebeat
+        tasks_from: uninstall-filebeat-as-daemonset.yml
+
+    - name: Include install task for Filebeat as DaemonSet in custom namespace for "k8s as cloud service"
+      include_role:
+        name: filebeat
+        tasks_from: install-filebeat-as-daemonset.yml
```
atsikham, line 24: Am I correct that if we run `upgrade` a few times with the same version it will be reinstalled anyway? The same question for the left part of PR.

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/node-exporter.yml
```diff
@@ -11,32 +7,60 @@
...
+  delegate_to: localhost
+  become: false
+  run_once: true
+  environment: { KUBECONFIG: "{{ vault_location }}/../kubeconfig" }
+  command: helm list
+  register: check_node_exporter_release
+
+- name: Reinstall Node Exporter as DaemonSet from default namespace to custom namespace for "k8s as cloud service"
+  when:
+    - specification.helm_chart_name in check_node_exporter_release.stdout
```
sk4zuzu, line 20: Isn't `specification.helm_chart_name in check_node_exporter_release.stdout` too wide of a condition? :thinking: What if user deploys a chart with a name that completely contains `specification.helm_chart_name`? :cat2:

sk4zuzu (changes requested):

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/filebeat.yml
rpudlowski93: No, it is just a fix in order to reinstall from default namespace to custom namespace the daemonset. If we run the upgrade once, the daemonset will reinstall to custom namespace and if we run upgrade once again, the upgrade process will skip since there is the conditions which check only the default namespace

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/filebeat.yml
rpudlowski93: True, to be honest i wasn't sure of the condition and I was waiting for this kind of review just to be sure. I will do it more strictly

sk4zuzu: I suspect of you'd use `stdout_lines` that should make it work. :thinking:

rpudlowski93: Changes added, now it looks good in my opinion. Please check.

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/node-exporter.yml
rpudlowski93: Changes added, now it looks good in my opinion. Please check.

sk4zuzu (commented):
What about something like this? In this case it's not a big performance improvement, but this is a better way to do these things in general, it's faster and we get rid of useless log entries. :thinking: 

```yaml
---
- hosts: localhost
  tasks:
    - set_fact:
        specification:
          helm_chart_name: asd2
        helm_ls:
          - {"name":"asd1","namespace":"default","revision":"20","updated":"2020-11-20 03:14:29.269115248 +0100 CET","status":"failed","chart":"asd-0.0.1","app_version":"0.0.1"}
          - {"name":"asd2","namespace":"default","revision":"20","updated":"2020-11-20 03:14:29.269115248 +0100 CET","status":"failed","chart":"asd-0.0.1","app_version":"0.0.1"}

    - set_fact:
        helm_release_exists: >-
          {{ _names | ternary(true, false) }}
      vars:
        _names: >-
          {{ helm_ls | map(attribute='name')
                     | select('eq', specification.helm_chart_name)
                     | list }}

    - debug: msg=OK
      when: helm_release_exists
```

core/src/epicli/data/common/ansible/playbooks/roles/node_exporter/tasks/uninstall-node-exporter-as-daemonset.yml
sk4zuzu: I think in case of just the helm feature we have to keep such construction, but for filebeat and node_exporter we have properly defined non-localhost ansible groups, @atsikham I think you're right. :thinking: Actually in both cases lines
```
environment: { KUBECONFIG: "{{ vault_location }}/../kubeconfig" }
```
can be removed because it's already defined at playbook level. :+1:

core/src/epicli/data/common/ansible/playbooks/roles/filebeat/defaults/main.yml
sk4zuzu: We've decided to keep it in defaults for now. After this refactor #1756 it will be easier to properly implement upgrades for this case.

core/src/epicli/data/common/ansible/playbooks/roles/node_exporter/tasks/uninstall-node-exporter-as-daemonset.yml
sk4zuzu: @rpudlowski93 any thoughts? :hugs:

rpudlowski93: I've just added the kubeconfig env to upgrade playbook level and removed the env from role. Tested, works fine. Small changes also added in order to optimize code

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/node-exporter.yml
```diff
@@ -7,24 +7,25 @@
...
-  set_fact:
-    node_exporter_release_exists: >-
-      {{ _names | ternary(true, false) }}
-  vars:
-    _names: >-
-      {{ helm_list.stdout | from_json 
-                          | map(attribute='name')
-                          | select('==', specification.helm_chart_name)
-                          | list }}
+  block:
```
sk4zuzu, line 14: :heart:

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/filebeat.yml
```diff
@@ -1,22 +1,23 @@
 ---
-- name: Filebeat as DaemonSet | Get information about helm releases
+- name: Filebeat as DaemonSet | Prepare fact about helm release
   delegate_to: localhost
   become: false
   run_once: true
-  environment: { KUBECONFIG: "{{ vault_location }}/../kubeconfig" }
-  command: helm list --output json
-  register: helm_list
+  block:
```
sk4zuzu, line 6: :heart:

sk4zuzu (approved):
LGTM :+1::heart_eyes:

atsikham (approved):

sk4zuzu (approved):
:monkey:

atsikham (approved):
:partying_face:

PR #1845, opened 2020-11-12
Pgpool not showing Replication State

to-bar:
* Correct backend_application_name in pgpool.conf
* Group tasks and change order
* Remove unneeded tasks
* Group tasks using blocks
* Improve code
* Create symlink /etc/repmgr.conf
* Shorten repmgr commands due to symlink to config file
* Ensure postgresql instance is stopped
* Update changelog
* Update documentation

### DIFF
--- a/core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/replication-repmgr-Debian.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/replication-repmgr-Debian.yml
@@ -84,9 +84,7 @@
         {{ repmgr_bindir[ansible_os_family] }}/repmgr cluster show -f {{ repmgr_config_dir[ansible_os_family] }}/repmgr.conf | grep -i {{ inventory_hostname }} | grep -v standby
       changed_when: false
       register: is_primary_already_registered
-      failed_when: is_primary_already_registered.rc == 2
-      args:
-        executable: /bin/bash
+      failed_when: is_primary_already_registered.rc not in [0, 1]
 
     - name: Register primary node
       when: is_primary_already_registered.stdout == ""
@@ -113,9 +111,7 @@
         {{ repmgr_bindir[ansible_os_family] }}/repmgr cluster show -f {{ repmgr_config_dir[ansible_os_family] }}/repmgr.conf | grep -i {{ inventory_hostname }}
       changed_when: false
       register: is_node_already_registered
-      failed_when: is_node_already_registered.rc == 2
-      args:
-        executable: /bin/bash
+      failed_when: is_node_already_registered.rc not in [0, 1]
 
     - name: Clone node from primary and register
       when: is_node_already_registered.stdout == ""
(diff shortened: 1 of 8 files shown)


rafzei (approved):
LGTM

plirglo (approved):

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/replication-repmgr-Debian.yml
```diff
@@ -1,185 +1,181 @@
...
-- name: Restart repmgrd service
-  service:
-    name: "{{ repmgr_service_name[ansible_os_family] }}"
-    state: restarted
-  when:
-    - groups['postgresql'][1] == inventory_hostname
-    - is_node_already_registered.stdout == ""
+# For repmgr installed from Ubuntu package additional configuration is required before repmgrd is started as daemon
+- name: Set repmgr.conf file in /etc/default/repmgrd
+  replace:
```
sk4zuzu, line 77: Same argument as in `I don't really like that we're using lineinfile, for me it's ugly..` :hugs:

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/replication-repmgr-Debian.yml
```diff
@@ -1,185 +1,181 @@
...
-
-# All
-- name: Restart repmgrd service
-  service:
-    name: "{{ repmgr_service_name[ansible_os_family] }}"
-    state: restarted
+# --- Configure primary node ---
+
+- name: Tasks to run only on primary node
+  when: groups.postgresql[0] == inventory_hostname
```
sk4zuzu, line 91: Again this is not in scope of this PR, but I think we talked about this somewhere before. This is a naive asumption that we can rely on ordering of inventory groups. Especially in AWS where we miss-use autocaling groups and in the end make Epiphany `un-scallable` in the process. :(

sk4zuzu (approved):
Code in this PR looks ok to me, but I'm not impressed after reading whole that role for the first time. :( I guess my suggestions are more for refactoring the role which is out of the scope of this PR. :hugs:

My recommendations for the future would be:
1. Make important settings (like archive_mode that we care about) mapped one-to-one with the target config files, which would elliminate having them seeded with random values coming from the package creator in case of issues with merging.
2. Add schema validation! So an user has any kind of feedback if after any upgrade current config is not disabling/enabling some important settings. 
3. Make all configs dict-based instead of being list-based (that will enable merging) or we should add proper merging of list-based dicts with `name:` as a key. 
4. Get rid of lineinfile/replace statements, instead use full config templates and include statements.
5. Probably good idea would be keeping Postgres' data dir in a custom place (identical on all Linux distros), it would be handy with upgrades I think.
6. Relying on ordering of postgresql group is naive and simply incorrect if you take into account autoscaling groups in AWS. Is there any upgrade procedure to switch from 1 master to 1 master + 1 slave and back - is it supported?
7. Please don't use Software Collections aver again anywhere in the code! :hugs:

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/replication-repmgr-Debian.yml
to-bar: I created issue #1846 to solve that.

Issue #1865, opened 2020-11-18
[BUG] Epicli fails on helm installation when using a custom image registry and local repository


przemyslavic:
**Describe the bug**
Epicli fails on helm installation when using a custom image registry and local repository.

**To Reproduce**
1. Execute `epicli apply -f testcustomrepo.yml` using the configuration given below.

**Expected behavior**
The cluster has been deployed successfully.

**Config files**
```yaml
...
specification:
  components:
    kubernetes_master:
      count: 1
    kubernetes_node:
      count: 1
    repository:
      count: 0
...
---
kind: configuration/shared-config
title: Shared configuration that will be visible to all roles
provider: aws
name: default
specification:
  custom_image_registry_address: "x.x.x.x:5000"
  custom_repository_url: ""
---
kind: configuration/feature-mapping
title: "Feature mapping to roles"
provider: aws
name: default
specification:
  roles_mapping:
    kubernetes_master:
      - repository
      - image-registry
      - kubernetes-master
      - helm
      - applications
      - node-exporter
      - filebeat
      - firewall
      - vault
```
**OS (please complete the following information):**
 - OS: [all]

**Cloud Environment (please complete the following information):**
 - Cloud Provider [all]

**Additional context**
Log:
```
11:05:58 INFO cli.engine.ansible.AnsibleCommand - Running: "/ s h a r e d / b u i l d / t e s t - c u s t o m r e p o / a n s i b l e / h e l m . y m l"
11:05:59 WARNING cli.engine.ansible.AnsibleCommand - [WARNING]: Found variable using reserved name: name
11:05:59 INFO cli.engine.ansible.AnsibleCommand -
11:05:59 INFO cli.engine.ansible.AnsibleCommand - PLAY [127.0.0.1] *****************************************************************************************************************************************************************************************************************************
11:05:59 INFO cli.engine.ansible.AnsibleCommand -
11:05:59 INFO cli.engine.ansible.AnsibleCommand - TASK [preflight_facts : Load preflight facts] ************************************************************************************************************************************************************************************************
11:05:59 INFO cli.engine.ansible.AnsibleCommand - ok: [127.0.0.1]
11:05:59 INFO cli.engine.ansible.AnsibleCommand -
11:05:59 INFO cli.engine.ansible.AnsibleCommand - TASK [helm : Check if index file exists for system Helm repo] ********************************************************************************************************************************************************************************
[...]

PR #1867, opened 2020-11-19
sshtunnel/helm: use correct custom flag to enable/disable helm tunnel (fix)

sk4zuzu:


atsikham (approved):

to-bar (approved):

to-bar (changes requested):
Please update changelog.

atsikham:
> Please update changelog.

Do we need to do that if there is no task related to PR? Or it isn't so?

to-bar:
> 
> 
> > Please update changelog.
> 
> Do we need to do that if there is no task related to PR? Or it isn't so?

I've noticed that there is linked task #1865.

sk4zuzu:
@to-bar there is no need for adding changelog because this is the first time we release that feature. :+1::innocent:

to-bar (approved):

to-bar:
> 
> 
> @to-bar there is no need for adding changelog because this is the first time we release that feature. 👍😇

OK, so then I approve.

PR #1869, opened 2020-11-24
BDS scan improvements.

seriva:
Updated gen-licences.py script to create dependencies.xml to avoid scanning all python dependencies during the BDS scan.


core/src/epicli/gen-licenses.py
```diff
@@ -114,5 +114,34 @@ def _main() -> None:
...
+		<source></source>
+	</component>"""
+        count=count+1
+
+    dependencies_content = dependencies_content + """
+</components>
+"""
+    path = os.path.join(os.path.dirname(__file__), 'dependencies.xml')
+    with open(path, 'w') as file:
+        file.write(dependencies_content)
```
to-bar, line 144: Using `textwrap.dedent` we could indent multiline strings, for example:
```
    dependencies_content = """\
    <?xml version="1.0" encoding="UTF-8" standalone="yes"?>
    <components xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">"""

    count = 1
    for dep in all_deps_data:
        dep_name = dep['Name']
        dep_version = dep['Version']
        dep_website = dep['Home-page']
        dep_license = dep['License']
        dependencies_content = dependencies_content + f"""
        <component>
            <id>{count}</id>
            <name>{dep_name}</name>
            <version>{dep_version}</version>
            <license>{dep_license}</license>
            <url>{dep_website}</url>
            <source></source>
        </component>"""
        count=count+1

    dependencies_content = dependencies_content + """
    </components>
    """
    path = os.path.join(os.path.dirname(__file__), 'dependencies.xml')
    with open(path, 'w') as file:
        file.write(textwrap.dedent(dependencies_content).replace("    ", "\t"))
```

seriva: Good suggestion:)

core/src/epicli/gen-dependency-info.py
```diff
@@ -83,36 +84,63 @@ def _main() -> None:
...
+        dependencies_content = dependencies_content + f"""
+        <component>
+            <id>{count}</id>
+            <name>{dep_name}</name>
+            <version>{dep_version}</version>
+            <license>{dep_license}</license>
+            <url>{dep_website}</url>
+            <source></source>
+        </component>"""
+        count=count+1
```
to-bar, line 132:
```suggestion
        count = count + 1
```

to-bar (approved):

atsikham (approved):
There is no newline, but it's not related to changes.

PR #1876, opened 2020-11-25
epicli upgrade fails when there is no kubernetes_master group in inventory

to-bar:


sk4zuzu (dismissed):
:mechanical_arm:

core/src/epicli/cli/engine/ansible/AnsibleInventoryUpgrade.py
```diff
@@ -101,26 +101,19 @@ def upgrade(self):
...
         self.cluster_model = default_cluster_model
 
-        # Check if repo roles are present and if not add them
-        master = self.get_role(new_inventory, 'kubernetes_master')
-        if master == None:
-            raise Exception('No kubernetes_master to use as repository')
-        master_node = master.hosts[0]
+        # repository & image_registry roles added in v0.4.0
+        repository = self.get_role(new_inventory, 'repository')
+        if repository == None:
```
atsikham, line 106: https://www.python.org/dev/peps/pep-0008/#programming-recommendations
`Comparisons to singletons like None should always be done with is or is not, never the equality operators`

atsikham (approved):

sk4zuzu (approved):

Issue #1875, opened 2020-11-25
[BUG] epicli upgrade fails when there is no kubernetes_master group in inventory


to-bar:
**Describe the bug**
`epicli upgrade` requires `kubernetes_master` group. For AKS/EKS clusters there is no such group.

```
INFO cli.engine.UpgradeEngine - Run done in 20887ms
ERROR epicli - No kubernetes_master to use as repository
```

**How to reproduce**
1. Deploy Epiphany cluster without k8s or with k8s as cloud service (`k8s_as_cloud_service: true`). 
2. Run `epicli upgrade`.

**Expected behavior**
Upgrade succeeded.

**Environment**
- Cloud provider: All
- OS: All

**Additional context**
The error is from `core\src\epicli\cli\engine\ansible\AnsibleInventoryUpgrade.py`.

przemyslavic:
✅ The installation of the cluster without K8s has been checked and the upgrade now works fine.

PR #1878, opened 2020-11-26
Include default vars to filebeat's upgrade role

rpudlowski93:
- PR releated to failed test in task #1833 
- Default vars included to filebeat upgrade role from filebeat role
- Filebeat chart name moved from specification to defaults

### DIFF
--- a/core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/filebeat.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/filebeat.yml
@@ -4,10 +4,13 @@
     file: roles/filebeat/defaults/main.yml
     name: filebeat_defaults
 
-- name: Filebeat as DaemonSet | Prepare fact about helm release
+- name: Filebeat as DaemonSet | Prepare fact about helm release and reinstall daemonset if release exists
   delegate_to: localhost
   become: false
   run_once: true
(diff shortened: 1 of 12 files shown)


core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/filebeat.yml
```diff
@@ -1,4 +1,9 @@
 ---
+- name: Filebeat | Load defaults from filebat role
```
to-bar, line 2:
```suggestion
- name: Filebeat | Load defaults from filebeat role
```

to-bar (changes requested):

to-bar (approved):

przemyslavic:
@rpudlowski93 I found a bug while testing fix for failing upgrades [#1875](https://github.com/epiphany-platform/epiphany/issues/1875) (PR [#1876](https://github.com/epiphany-platform/epiphany/pull/1876)). I deployed a cluster without kubernetes and helm. During upgrade it fails on task `Get information about helm releases`. The entire block `Filebeat as DaemonSet | Prepare fact about helm release` should be run conditionally (when k8s_as_cloud_service). Same for node exporter.

to-bar (approved):

atsikham (approved):

PR #1879, opened 2020-11-30
Increase yum lock timeout (workaround)

sk4zuzu:


core/src/epicli/data/common/ansible/playbooks/roles/grafana/tasks/Debian.yml
```diff
@@ -0,0 +1,22 @@
...
+  apt:
+    name: grafana-data
+    update_cache: true
+    state: absent
+
+- name: Install dependencies for Debian family
+  apt:
+    name:
+      - apt-transport-https
+      - adduser
```
atsikham: Do we need this package? If no, it can be also removed from repository part.

atsikham:
I would use `|lower` for `ansible_os_family` everywhere, but in general LGTM.

sk4zuzu:
> I would use `|lower` for `ansible_os_family` everywhere, but in general LGTM.

What we really would like to do is upgrade ansible to 2.10.3 or 2.10.4 and then use bugfixed "package" and remove chunks of code from this PR. That's why I'm not so enthusiastic about merging it :hugs:

sk4zuzu:
I can confirm that in ansible 2.10.5, passing module defaults to `yum` module through the `package` module finally works :+1::innocent:, so I was able to simplify this PR! :hugs:

core/src/epicli/data/common/ansible/playbooks/roles/kubernetes_common/tasks/install-packages.yml
```diff
@@ -1,12 +1,38 @@
...
+      Debian:
+        - cifs-utils
+        - samba-client
+        - samba-common
+        - nfs-common
+      RedHat:
+        - cifs-utils
+        - samba-client
+        - samba-common
+        - nfs-utils
```
atsikham, line 17: Haven't seen it yet, but what do you think about how it should look like when there is a long common list of packages?

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/docker.yml
```diff
@@ -41,12 +43,12 @@
...
+          yum:
             name: docker-ce-{{ ansible_facts.packages['docker-ce'][0].version }}
             state: absent
 
         - name: Remove docker-ce-selinux
-          package:
+          yum:
             name: docker-ce-selinux-{{ ansible_facts.packages['docker-ce-selinux'][0].version }}
             state: absent
           when: "'docker-ce-selinux' in ansible_facts.packages"
```
atsikham, line 54: Do we need the last condition? If no, can it be combined to the one task?

core/src/epicli/data/common/ansible/playbooks/roles/kubernetes_common/tasks/install-packages.yml
sk4zuzu: If it's really long we could just do set arithmetic (union) in jinja2, but here I think it's not worth it :thinking:

atsikham: Agree.

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/docker.yml
sk4zuzu: I don't really know at this point :thinking: @to-bar can you shed some light here? :+1: :innocent:

to-bar: Today I would write this like:

```
- name: Remove Docker in version < 18.09 (RedHat)
  when: ansible_os_family == "RedHat"
  module_defaults:
    yum: { lock_timeout: 300 }
  block:
    - name: Get information about installed packages as facts
      package_facts:
        manager: auto

    - name: Remove docker-ce and docker-ce-selinux
      yum:
        name:
          - docker-ce
          - docker-ce-selinux
        state: absent
      when:
        - "'docker-ce' in ansible_facts.packages"
        - ansible_facts.packages['docker-ce'][0].version is version ('18.09', '<')
```
but not tested :confused:

atsikham (approved):
LGTM

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/docker.yml
sk4zuzu: I think another PR would be a better place for this :)

to-bar (approved):
:ant:

atsikham (approved):

Issue #1854, opened 2020-11-16
[UPGRADE] Upgrade component to the latest version: RabbitMQ


rafzei:
**Is your feature request related to a problem? Please describe.**
- [CVE-2020-5419 & CVE-2018-11087](https://nvd.nist.gov/vuln/search/results?form_type=Advanced&results_type=overview&search_type=all&cpe_vendor=cpe%3A%2F%3Avoxpupuli&cpe_product=cpe%3A%2F%3A%3Arabbitmq&cpe_version=cpe%3A%2F%3Apivotal_software%3Arabbitmq%3A3.8.3) - Epiphany v0.7, v0.8
- End of support for version 3.7.10 (Epiphany v0.5, v0.6)
- [CVE-2019-11287 & CVE-2020-5419 & CVE-2018-11087 & CVE-2019-11281 & CVE-2019-11291](https://nvd.nist.gov/vuln/search/results?form_type=Advanced&results_type=overview&search_type=all&cpe_vendor=cpe%3A%2F%3Apivotal_software&cpe_product=cpe%3A%2F%3A%3Arabbitmq&cpe_version=cpe%3A%2F%3Apivotal_software%3Arabbitmq%3A3.7.10) - Epiphany v0.5, v0.6

**Describe the solution you'd like**
- upgrade to latest version if possible
- backport to supported Epiphany versions

przemyslavic:
Due to the vulnerabilities mentioned above, we need to update RabbitMQ to the latest versions (at the moment it is RabbitMQ 3.8.9 release). This applies to both standalone installation and deployment on Kubernetes.

przemyslavic:
Checked:
✅`epicli apply`: vm installation + k8s deployment
✅`epicli upgrade`: vm installation
It was agreed (@atsikham @sk4zuzu) not to update the k8s deployment during the upgrade process.

PR #1889, opened 2020-12-04
Upgrade rabbitmq to v3.8.9

atsikham:
Task #1854: implemented upgrade for RabbitMQ. Latest version is compatible with previous 3.8.3 and 3.7.x.
See https://www.rabbitmq.com/upgrade.html for more details.


core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/kubernetes/reconfigure-rabbitmq-app.yml
```diff
@@ -7,8 +7,9 @@
   block:
     - name: after-patching | Get rabbitmq namespace
       shell: |-
-        kubectl get statefulsets.apps --all-namespaces -o=jsonpath='{range .items[*]}{.metadata.namespace}{"\t"}{.spec.template.spec.containers[].image}{"\n"}{end}'|
-        awk '/rabbitmq/ {print $1}'
+        kubectl get statefulsets.apps --all-namespaces \
```
sk4zuzu, line 10: @atsikham how about replacing all that ugly code with something like this (:thinking:):
```yaml
- command: kubectl get statefulsets --all-namespaces --output json
  register: shell_get_statefulsets
  changed_when: false

- command: kubectl delete pods --namespace {{ item }} --all
  loop: "{{ _namespaces }}"

  when: _namespaces | length > 0

  vars:
    _namespaces: >-
      {%- set output = [] -%}
      {%- for item in (shell_get_statefulsets.stdout | from_json)["items"] -%}
        {%- for container in item.spec.template.spec.containers -%}
          {%- if '/rabbitmq:' in container.image -%}
            {{- output.append(item.metadata.namespace) -}}
          {%- endif -%}
        {%- endfor -%}
      {%- endfor -%}
      {{- output -}}
```

atsikham: @sk4zuzu I suggest to create another PR for that as testing scope is different. I changed that line and it's showed here as there was a linter warning.

sk4zuzu: That's fine! If it gets there somehow I'm :+1::heart_eyes:

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/rabbitmq.yml
```diff
@@ -0,0 +1,71 @@
+---
+- name: RabbitMQ | Set rabbitmq_master fact
+  set_fact:
+    rabbitmq_master: "{{ groups['rabbitmq'][0] }}"
+
+- name: RabbitMQ | Check version
+  command: rabbitmqctl version
+  register: rabbitmq_version_current
+
```
sk4zuzu, line 9:
```suggestion
  changed_when: false

```

sk4zuzu (changes requested):

sk4zuzu (approved):
LGTM

toszo:
I like the upgrade.md update - it sets a context and makes clear how upgrade of RMQ will look like. 
LGTM

toszo (approved):

PR #1790, opened 2020-10-23
Upgrade kafka 2.6.0

sunshine69:

PR #1684, opened 2020-09-18
Feature/Postgres-exporter

plirglo:
Create postgres-exporter Ansible role

PR #1890, opened 2020-12-06
Kafka update to 2.6.0

toszo:
Added small changes to PR made by @sunshine69 (#1790)
Tested clean install and upgrade:
- on prem Ubuntu 1804
- AWS Ubuntu 18.04

Kafka updated in: 
- defaults file,
- requirements.txt (centos, rhel, ubuntu)
- changelog for 0.9
- components.md

Additionally:
Added socket_settings section in Kafka yaml that allows to modify msg sizes, io threads number, network threads number - those values are often manually modified after epicli run.

### DIFF
--- a/core/src/epicli/data/common/ansible/playbooks/roles/kafka/defaults/main.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/kafka/defaults/main.yml
@@ -1,5 +1,3 @@
-specification:
-  kafka_var:
-    version: 2.6.0
-    scala:
-      version: 2.12
+kafka_version: 2.6.0
+scala_version: 2.12
+kafka_bin_filename: "kafka_2.12-2.6.0.tgz"
--- a/core/src/epicli/data/common/ansible/playbooks/roles/kafka/defaults/main.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/kafka/defaults/main.yml
@@ -1,5 +1,5 @@
 specification:
   kafka_var:
-    version: 2.3.1
+    version: 2.6.0
     scala:
       version: 2.12
--- a/core/src/epicli/data/common/defaults/configuration/kafka.yml
+++ b/core/src/epicli/data/common/defaults/configuration/kafka.yml
@@ -4,10 +4,10 @@ name: default
 specification:
 
   kafka_var:
-    version: 2.3.1
+    version: 2.6.0
     scala:
       version: 2.12
-    file_name: "kafka_2.12-2.3.1.tgz"
+    file_name: "kafka_2.12-2.6.0.tgz"
     enabled: True
     admin: kafka
     admin_pwd: epiphany
@@ -73,6 +73,12 @@ specification:
     conf_dir: /opt/kafka/config
     data_dir: /var/lib/kafka
     log_dir: /var/log/kafka
+    socket_settings:
+      network_threads: 3 # The number of threads handling network requests
+      io_threads: 8 # The number of threads doing disk I/O
+      send_buffer_bytes: 102400 # The send buffer (SO_SNDBUF) used by the socket server
+      receive_buffer_bytes: 102400 # The receive buffer (SO_RCVBUF) used by the socket server      
+      request_max_bytes: 104857600 # The maximum size of a request that the socket server will accept (protection against OOM)
 
   zookeeper_set_acl: false
   zookeeper_hosts: "{{ groups['zookeeper']|join(':2181,') }}:2181"
--- a/core/src/epicli/data/common/defaults/configuration/kafka.yml
+++ b/core/src/epicli/data/common/defaults/configuration/kafka.yml
@@ -2,16 +2,10 @@ kind: configuration/kafka
 title: "Kafka"
 name: default
 specification:
-
   kafka_var:
-    version: 2.6.0
-    scala:
-      version: 2.12
-    file_name: "kafka_2.12-2.6.0.tgz"
     enabled: True
     admin: kafka
     admin_pwd: epiphany
-
     security:
       ssl:
         enabled: False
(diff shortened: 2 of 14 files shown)


core/src/epicli/data/common/defaults/configuration/kafka.yml
```diff
@@ -73,6 +73,12 @@ specification:
     conf_dir: /opt/kafka/config
     data_dir: /var/lib/kafka
     log_dir: /var/log/kafka
+    socket_settings:
+      network_threads: 3 # The number of threads handling network requests
+      io_threads: 8 # The number of threads doing disk I/O
+      send_buffer_bytes: 102400 # The send buffer (SO_SNDBUF) used by the socket server
+      receive_buffer_bytes: 102400 # The receive buffer (SO_RCVBUF) used by the socket server      
+      request_max_bytes: 104857600 # The maximum size of a request that the socket server will accept (protection against OOM)
```
plirglo, line 81: Shouldn't those parameters be placed in file roles/kafka/defaults/main.yml?

core/src/epicli/data/common/ansible/playbooks/roles/kafka/defaults/main.yml
```diff
@@ -1,5 +1,5 @@
 specification:
```
atsikham, line 1: `specification` can be defined by user and used in `configuration` part, for role defaults it's not so and I think it should be changed here.

core/src/epicli/data/common/defaults/configuration/kafka.yml
```diff
@@ -4,10 +4,10 @@ name: default
 specification:
 
   kafka_var:
```
atsikham, line 6: :pensive:

core/src/epicli/data/common/defaults/configuration/kafka.yml
```diff
@@ -4,10 +4,10 @@ name: default
...
-    version: 2.3.1
+    version: 2.6.0
     scala:
       version: 2.12
-    file_name: "kafka_2.12-2.3.1.tgz"
+    file_name: "kafka_2.12-2.6.0.tgz"
```
atsikham, line 10: If we don't allow user to change/choose the component versions, It should be moved to the role defaults. The same for any other parameters that are not allowed/supposed to be changed by user.

core/src/epicli/data/common/defaults/configuration/kafka.yml
atsikham: Agree with @plirglo if it shouldn't be changed by user.

atsikham (changes requested):
In addition to comments please check that `upgrade` document is up to date and inspect the whole role with Ansible linter. The last point is not about this task, but otherwise It will be never done and improved.

core/src/epicli/data/common/defaults/configuration/kafka.yml
toszo: User should be able to change those parameters. There are users that are running Kafka and having bigger messages than allowed by default - that is why it was added.

core/src/epicli/data/common/ansible/playbooks/roles/kafka/defaults/main.yml
toszo: Could you elaborate more? And show some example on ` like we started to do for other roles.`?

core/src/epicli/data/common/ansible/playbooks/roles/kafka/defaults/main.yml
```diff
@@ -1,5 +1,5 @@
...
   kafka_var:
```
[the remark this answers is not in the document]
toszo: I agree, but changing it will cause a breaking change for existing cluster configuration. It will require to add validation to yaml if those are updated. 
I propose to create a "Kafka role refactoring" task to prevent "wild" refactorings across the solution.

toszo:
> In addition to comments please check that `upgrade` document is up to date and inspect the whole role with Ansible linter. The last point is not about this task, but otherwise It will be never done and improved.

My proposition is to create a new issue "Kafka role refactoring" or "Create Kafka Module" where major changes can be covered.

core/src/epicli/data/common/ansible/playbooks/roles/kafka/defaults/main.yml
atsikham: Here is the [thread](https://github.com/epiphany-platform/epiphany/pull/1684#discussion_r491903720) to follow. As I remember all necessary changes were done there, so it can be used as an example.

atsikham:
> My proposition is to create a new issue "Kafka role refactoring" or "Create Kafka Module" where major changes can be covered.

Separate task is ok for me, but let's be honest, it will be taken to be done not soon/never.

plirglo:
> > My proposition is to create a new issue "Kafka role refactoring" or "Create Kafka Module" where major changes can be covered.
> 
> Separate task is ok for me, but let's be honest, it will be taken to be done not soon/never.

There is a lot of changes to do in kafka tasks (eg. versions in two separate files with the same value), but i don't think it's part of this one. We should create addition change with refactoring of this role

atsikham (dismissed):
Approved with a note that "refactoring" task should be created.

core/src/epicli/data/common/ansible/playbooks/roles/kafka/defaults/main.yml
toszo: Modified kafka defaults to have values that will not be editable by the user.

core/src/epicli/data/common/defaults/configuration/kafka.yml
toszo: Moved to defaults :)

toszo:
I created another issue for the creation of Kafka Module #1894. If the direction of this component will be changed, this component should be refactored.

atsikham (approved):

plirglo (approved):

core/src/epicli/data/common/defaults/configuration/kafka.yml
```diff
@@ -2,16 +2,10 @@ kind: configuration/kafka
 title: "Kafka"
 name: default
 specification:
-
   kafka_var:
-    version: 2.3.1
-    scala:
-      version: 2.12
-    file_name: "kafka_2.12-2.3.1.tgz"
```
sk4zuzu, line 10: It's good that we removed it from here :+1::innocent:

sk4zuzu (approved):
LGTM

Issue #1853, opened 2020-11-16
[UPGRADE] Upgrade component to the latest version: Zookeeper


rafzei:
**Is your feature request related to a problem? Please describe.**
- End of support for 3.4.12 version (06.2020).
- [CVE-2019-0201](https://nvd.nist.gov/vuln/search/results?form_type=Advanced&results_type=overview&search_type=all&cpe_vendor=cpe%3A%2F%3Aapache&cpe_product=cpe%3A%2F%3Aapache%3Azookeeper&cpe_version=cpe%3A%2F%3Aapache%3Azookeeper%3A3.4.12)

**Describe the solution you'd like**
- upgrade to latest version if possible
- backport to supported Epiphany versions

**Additional context**
If possible, do an improvements in Zookeeper role. :+1:

rafzei:
Please verify if:
- latest Kafka version contains embedded Zookeeper
- we need a standalone Zookeeper (ask other teams)

plirglo:
Please see task #1137 - Kafka Upgrade. 
We need upgrade Zookeeper to version supported by Kafka latest version:
Kafka 2.6.0 -> Zookeeper 3.5.8

After Kafka 3.0 release Zookeeper probably will not be needed for Kafka.

mkyc:
Lets upgrade to 3.5.8.

toszo:
> Lets upgrade to 3.5.8.

@mkyc why 3.5.8 and not 3.6.2 - the latest one?

mkyc:
Hmmmm ... there was some reason @plirglo gave that version in discussion, but I do not remember that reason.

rpudlowski93:
Upgraded to 3.5.8 done in PR #1897

rpudlowski93:
@plirglo Can You comment why to version 3.5.8 not to latest one? I think that the information in kafka release was leading the decision : [link](https://downloads.apache.org/kafka/2.6.0/RELEASE_NOTES.html)kafka in latest version 3.6.2 could be nice (some workaround implemented in upgrading zookeeper to 3.5.8 version was needed). @plirglo Do You have any significant knowledge about it? If not I can make upgrade to latest version in mentioned PR #1897

rafzei:
After the discussion about the target version, can we backport this change to previous Epi versions?

plirglo:
[...]

PR #1897, opened 2020-12-09
Upgrade Zookeeper (to version 3.5.8)

rpudlowski93:
PR related to task #1853 

- New role added for zookeeper
- Upgrade tested with kafka upgrade. Cluster from 0.7.2 to feature/zookeeper-upgrade branch upgraded with success (kafka upgraded to 2.6.0, zookeeper to 3.5.8) 
- Zookeeper versions 3.5.8 and 3.4.12 are compatible with each other
- Zookeeper doesn't have any interface for output with version
- Zookeeper needs just for upgrading process existing snapshot in data dir: [link](https://cwiki.apache.org/confluence/display/ZOOKEEPER/Upgrade+FAQ) and adding flag snapshot.trust.empty=true is not recommened , we need to ensure that the system is in good shape
- Zookeeper upgrade playbook works on zookeeper hosts, zookeeper and kafka hosts must independent so we don't want to make some relations between them (possible use case with using only zookeeper) 
- Zookeeper needs only for upgrading any snapshot file
- Added small fix in jmx-exporter.yml default configuration. There should be version 0.14 instead of 0.12. Cluster creation crash with 0.12.


core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/zookeeper.yml
```diff
@@ -0,0 +1,28 @@
...
+- name: Include defaults from zookeeper role
+  include_vars:
+    file: roles/zookeeper/defaults/main.yml
+    name: zookeeper_defaults
+
+- name: Include pre-flight checks
+  include_tasks: zookeeper/preflight-check.yml
+
+- name: Get installed Zookeeper's version
+  shell: ls -la /opt/zookeeper | grep -oP '\d+\.\d+\.\d+'
```
sk4zuzu, line 11: This is so ugly :disappointed:. The thing with `grep -o` is that there are different versions that behave differently on various distributions! I'd avoid using it. Please don't do `ls` in scripts, use `find` or even better here use the anisble `find:` module.

sk4zuzu (changes requested):

core/src/epicli/data/common/defaults/configuration/zookeeper.yml
```diff
@@ -1,6 +1,4 @@
 kind: configuration/zookeeper
 title: "Zookeeper"
 name: default
-specification:
```
sk4zuzu, line 4: Fantastic! :+1::heart_eyes:

sk4zuzu (changes requested):
empty file -> core/src/epicli/data/common/ansible/playbooks/roles/zookeeper/vars/main.yml
what is this file ->  core/src/epicli/data/common/ansible/playbooks/roles/zookeeper/files/snapshot.0 ?
:thinking:

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/zookeeper.yml
sk4zuzu: Yes there is a difference between BSD grep and GNU grep in regards to the `-o` parameter, please avoid using it in production code! :scream:

rpudlowski93: You are right, improvement done for that. "ls" and "grep" removed forever :)

rpudlowski93:
> empty file -> core/src/epicli/data/common/ansible/playbooks/roles/zookeeper/vars/main.yml
> what is this file -> core/src/epicli/data/common/ansible/playbooks/roles/zookeeper/files/snapshot.0 ?
> 🤔

Empty file was already removed and additional description about the snapshot file added :)

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/zookeeper.yml
sk4zuzu: Fantastic improvement! :+1::heart_eyes:

sk4zuzu (approved):
LGTM

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/zookeeper/install-upgrade.yml
```diff
@@ -0,0 +1,78 @@
...
+  lineinfile:
+    path: /lib/systemd/system/zookeeper.service
+    state: present
+    regexp: '^ExecStart=/opt/zookeeper-.*'
+    line: "ExecStart=/opt/zookeeper/bin/zkServer.sh start-foreground"
+
+- name: Check if any snapshots exists in data dir what is necessary in order to run zookeeper after upgrade
+  find:
+    paths: "{{ zookeeper_defaults.zookeeper_data_dir }}/version-2"
+    patterns: 'snapshot*.'
```
to-bar, line 60: This pattern looks suspiciously. You have `use_regex: yes` so I would expect rather `snapshot.*` if you want to match files such as `snapshot` or `snapshot.something`. For such simple matching you can use the default (shell) pattern (without `use_regex`).

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/zookeeper/install-upgrade.yml
```diff
@@ -0,0 +1,78 @@
...
+    use_regex: yes
+  register: snapshot_exists
+
+# From 3.5.5 version, ZooKeeper is not able to start when no snapshot files present, what is valid scenario in 3.4.X version. Empty snapshot downloaded from Zookeeper's Jira ticket. 
+- name: Copy empty snapshot if not exists
+  copy:
+    dest: "{{ zookeeper_defaults.zookeeper_data_dir }}/version-2" 
+    src: roles/zookeeper/files/snapshot.0
```
to-bar, line 68: Since it seems this file is used only for upgrade, maybe it's better to put it somewhere inside upgrade role?

Alternative approach (not tested) is to force Zookeeper to generate a snapshot file when needed:
1. Set `snapshot.trust.empty` property in config file or use `-Dzookeeper.snapshot.trust.empty=true` property when there is no any snapshot.
2. Start ZK to generate a snapshot.
3. Remove the property.
4. Restart ZK.

Based on: https://issues.apache.org/jira/browse/ZOOKEEPER-3056?focusedCommentId=16974905&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16974905

to-bar (changes requested):

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/zookeeper/install-upgrade.yml
rpudlowski93: I was already testing this scenario and it didn't solve the problem. ZK will run with the property but we don't want to keep the property forever. if the ZK doesn't generate snapshot after startup and we will remove the property and we will restart zk, the problem will still exist. We can't force to generate snapshot.

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/zookeeper.yml
```diff
@@ -0,0 +1,29 @@
...
+  block:
+    - name: Stop Zookeeper service
+      service:
+        name: zookeeper
+        state: stopped
+
+    - name: Include upgrade Zookeeper task
+      include_tasks: zookeeper/install-upgrade.yml
+  when:
+    - before_upgrade_zookeeper_version is version( zookeeper_defaults.zookeeper_version, '<' )
```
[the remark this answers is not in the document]
rpudlowski93: Additional checks and lock files added. Good point 👍

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/zookeeper/install-upgrade.yml
```diff
@@ -0,0 +1,80 @@
+---
+
+# This task can be reused from zookeeper role. Rememeber about it in near future. We should refactor zookeeper role and split some code into smaller pieces.
```
to-bar, line 3:
```suggestion
# Some tasks are the same as in zookeeper role. It should be refactored (with splitting code into separate files) in order to reuse common tasks in upgrade role.
```

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/zookeeper/install-upgrade.yml
```diff
@@ -0,0 +1,80 @@
...
+    state: link
+    src: /opt/zookeeper-{{ zookeeper_defaults.zookeeper_version }}
+    force: yes
+  register: new_symbolic_link
+
+- name: Remove previous version binaries
+  file:
+    path: /opt/zookeeper-{{ before_upgrade_zookeeper_version }}
+    state: absent
+  when: new_symbolic_link.changed
```
to-bar, line 51:
```suggestion
  when: before_upgrade_zookeeper_version != zookeeper_defaults.zookeeper_version
```
This is clean up task so I would move it somewhere after starting ZK service. I think it's better to remove old version only after the new one is working 😉

to-bar (changes requested):

to-bar (approved):

sk4zuzu (approved):
LGTM

Issue #1884, opened 2020-12-02
[BUG] Prometheus is not able to scrape metrics from AKS/EKS nodes.


przemyslavic:
**Describe the bug**
Prometheus is not able to scrape metrics from AKS/EKS nodes .

**How to reproduce**
1. Deploy the AKS / EKS cluster with monitoring enabled.

**Expected behavior**
Prometheus configuration allows you to scrape metrics from node exporters installed as DaemonSets.

**Config files**
You can use the `m-generic-epiphany-on-aks` module with the following config:
```yaml
kind: azepi-config
azepi:
  vault_password: "test"
  config: |
    kind: epiphany-cluster
    title: Epiphany cluster Config
    name: azepi
    provider: any
    specification:
      name: azepi
      admin_user:
        name: operations
        key_path: ~/vms_rsa
      cloud:
        k8s_as_cloud_service: true
      components:
        repository:
          count: 1
          machines:
            - default-qa-monitoring-tests-all-0
        kubernetes_master:
          count: 0
        kubernetes_node:
          count: 0
        logging:
          count: 1
          machines:
            - default-qa-monitoring-tests-all-1
        monitoring:
          count: 1
          machines:
            - default-qa-monitoring-tests-all-0
        kafka:
          count: 0
        postgresql:
          count: 0
        load_balancer:
          count: 0
        rabbitmq:
          count: 0
    ---
    kind: configuration/feature-mapping
    title: Feature mapping to roles
    name: azepi
    provider: any
    specification:
      roles_mapping:
        repository:
          - repository
          - image-registry
          - firewall
          - filebeat
          - node-exporter
          - applications
    ---
    kind: infrastructure/machine
    name: default-qa-monitoring-tests-all-0
    provider: any
    specification:
      hostname: qa-monitoring-tests-all-0
      ip: x.x.x.x
    ---
    kind: infrastructure/machine
    name: default-qa-monitoring-tests-all-1
    provider: any
    specification:
      hostname: qa-monitoring-tests-all-1
      ip: y.y.y.y
```

**Environment**
- Cloud provider: [all]
- OS: [all]

**Additional context**
There are missing entries for node exporters in `k8s as cloud service` in the Prometheus configuration file found on the monitoring machine:
```
root@qa-monitoring-tests-all-0:~# cat /etc/prometheus/prometheus.yml
# Ansible managed
# http://prometheus.io/docs/operating/configuration/

global:
  scrape_interval: 15s #15s
[...]

PR #1898, opened 2020-12-09
Prometheus on managed k8s

atsikham:
Task #1884: Finally I was able to install Epiphany on top of modules and checked that Prometheus configuration template is rendered as expected and there are related metrics in Prometheus.

### DIFF
--- a/core/src/epicli/data/common/ansible/playbooks/roles/prometheus/tasks/main.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/prometheus/tasks/main.yml
@@ -31,7 +31,7 @@
 
 - name: "Check if Alertmanager is enabled"
   set_fact:
-    is_alertmanager_enabled: "{{ specification.alerts.alertmanager.enable == True and (specification.alerts.alertmanager.config.receivers | length>0 ) }}"
+    is_alertmanager_enabled: "{{ specification.alerts.alertmanager.enable|bool and (specification.alerts.alertmanager.config.receivers | length>0 ) }}"
 
 - include: install.yml
   become: true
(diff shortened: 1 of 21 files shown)


core/src/epicli/data/common/ansible/playbooks/roles/prometheus/tasks/main.yml
```diff
@@ -9,10 +9,10 @@
 
 - name: Set master exists flag
   set_fact:
-    does_k8s_exists: "(groups['kubernetes_master'] | default([])) | length > 0"
+    does_k8s_exist: "{{ (groups['kubernetes_master'] | default([]) | length > 0) or k8s_as_cloud_service }}"
```
atsikham, line 12: There are places where `k8s_as_cloud_service` is checked to be defined, but it's not necessary as it's located in default settings with `false` value.

sk4zuzu (dismissed):
LGTM

sk4zuzu (approved):

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/rabbitmq.yml
```diff
@@ -8,7 +8,7 @@
   register: rabbitmq_version_current
   changed_when: false
 
-- name: RabbitMQ | Include defaults from node_exporter role
+- name: RabbitMQ | Include defaults from rabbitmq role
```
atsikham, line 11: This is a typo from previous PR that has been already merged, imo we don't need a separate PR for that.

przemyslavic (approved):

PR #1903, opened 2020-12-10
Fix rabbitmq installation

atsikham:
Task #1854: installation bug fix. Tested locally:
* Installation:
1. AWS + RHEL, Ubuntu
2. Azure + RHEL, Ubuntu
* Upgrade

Long names usage is not supported by our Azure infrastructure as FQDNs are not used there, so default value has been changed.

### DIFF
--- a/core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/main.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/main.yml
@@ -1,8 +1,7 @@
 ---
-- name: Set facts
+- name: Set rabbitmq_master fact
   set_fact:
     rabbitmq_master: "{{ groups['rabbitmq'][0] }}"
-    long_names: "{% if specification.rabbitmq_use_longname|bool %}--longnames{% else %}{% endif %}"
 
 - include_tasks: "install-packages-{{ ansible_os_family | lower }}.yml"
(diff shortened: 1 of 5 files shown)


core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/main.yml
```diff
@@ -1,7 +1,8 @@
 ---
-- name: Set rabbitmq_master fact
+- name: Set facts
   set_fact:
     rabbitmq_master: "{{ groups['rabbitmq'][0] }}"
+    long_names: "{% if specification.rabbitmq_use_longname|bool %}--longnames{% else %}{% endif %}"
```
erzetpe, line 5: Maybe can be done different way? Maybe with using when, not to write such weird template?

erzetpe (changes requested):

core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/main.yml
atsikham: Removed it at all as variable should be used.

erzetpe (approved):

mkyc (approved):

rafzei (approved):
LGTM

Issue #1908, opened 2020-12-11
[SPIKE] Research why Epiphany nodes hang when memory is overcommited


sk4zuzu:
**Is your feature request related to a problem? Please describe.**
It seems that when k8s resource limits are incorrectly defined and memory is overcommited by applications user loses all ssh access to Epiphany VMs. This indicates that operating system is not protected from memory and cpu overusage. Usually it's not possible to debug such machines, restart is required and cluster becomes unstable. Since OS is not protected there is also security risk of possible denial-of-service attacks. :scream: 

I suppose it's likely systemd slices are not correctly configured on Epiphany VMs in 0.8 release (at least).

**Describe the solution you'd like**
What we need to do in this spike is decide what is the culprit here and research ways to prevent such situations in the future. We need confirmation also that this situation happens.

**Describe alternatives you've considered**
None.

**Additional context**
Logs found on affected nodes:
```
Killed process 13428 (java) total-vm:14218144kB, anon-rss:2096560kB, file-rss:3684kB, shmem-rss:0kB, UID:0 pgtables:25228kB oom_score_adj:998
```
```
kernel: [200403.323264] memory: usage 3145696kB, limit 3145728kB, failcnt 17191
```
```
dockerd[14005]: time="2020-11-25T21:40:40.609327299Z" level=warning msg="Your kernel does not support swap limit capabilities,or the cgroup is not mounted. Memory limited without swap."
```
```
 kernel: [ 0.289359] PM: Registered nosave memory: [mem 0x00000000-0x00000fff]
```

sk4zuzu:
I was basically unable to reproduce the issue in a generic way. I tried Ubuntu and JVM versions exactly as on the faulty environment:
```
  storage_image_reference:
    publisher: Canonical
    offer: UbuntuServer
    sku: 18.04-LTS
    version: 18.04.202006101
```
```
openjdk version "11.0.5" 2019-10-15
OpenJDK Runtime Environment (build 11.0.5+10-alpine-r0)
OpenJDK 64-Bit Server VM (build 11.0.5+10-alpine-r0, mixed mode)
```
Epiphany's Kubernetes behaved well in every attempt to crash node VMs. I think the next step would be to work closely with the team that experienced the issue and carefuly observe what their software does. :thinking:

@to-bar thank you for very nice memory stress testing toolset! :+1::heart_eyes:

mkyc:
@sk4zuzu what is next step here in this task? Why is it blocked?

[...]

PR #1909, opened 2020-12-11
Add k8s memory dev utils

to-bar:
Adds utils that may be useful for #1908.

### DIFF
--- a/tools/development/k8s/memory/memory-stress-deployments.yml
+++ b/tools/development/k8s/memory/memory-stress-deployments.yml
@@ -1,0 +1,216 @@
+# Deployments with memory request
+
+kind: Deployment
+apiVersion: apps/v1
+metadata:
+  name: memory-stress-1g-with-request
+spec:
+  replicas: 0
+  selector:
+    matchLabels:
+      app: memory-stress-1G
+  template:
+    metadata:
+      labels:
+        app: memory-stress-1G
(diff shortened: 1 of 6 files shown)


sk4zuzu (dismissed):
:heart_eyes:

tools/development/k8s/memory/memory-stress-deployments.yml
```diff
@@ -0,0 +1,216 @@
...
+        - stress
+        args:
+        - --vm
+        - "1"
+        - --vm-bytes
+        - 1M
+        - --vm-keep
+        - --vm-hang
+        - "1"
+---
```
atsikham, line 216: Is that line correct or in the correct place?

atsikham:
It would be nice to have a note in some place how to use that.

tools/development/k8s/memory/memory-stress-deployments.yml
to-bar: For k8s it is correct but it doesn't comply with yaml standard (i.e. three periods) so I've removed it 👍

to-bar:
> 
> 
> It would be nice to have a note in some place how to use that.

Readme files added 👍

rafzei (approved):
Nice, I would like to give it a try ASAP 🤟

atsikham (approved):

PR #1930, opened 2020-12-29
Epiphany support matrix

rafzei:


seriva:
Might I suggest using mermaidjs to make the lifecycle diagram to make it easier to upgrade over time. This is slowly becoming the new standard for doing diagrams inside markup. Lots of platforms like GitLab already support it and there is an outstanding feature request to for GitHub to add it as well.

https://mermaid-js.github.io/mermaid/#/gantt

You can add it in a separate MD file and inspect it with this extension:

https://chrome.google.com/webstore/detail/github-%2B-mermaid/goiiopgdnkogdbjmncgedmgpoajilohe/related

Until GitHub adds support we can include the PNG of a screenshot we make with that extension like you have now.

docs/home/LIFECYCLE.md
```diff
@@ -0,0 +1,14 @@
+# Epiphany platform lifecycle
```
seriva, line 1: Maybe also add a note that we current support the quarterly minor releases for a year with new patch versions but that the patch versions itself fall under the minor releases.

seriva:
Also maybe make an entry for this from the main readme: https://github.com/epiphany-platform/epiphany/blob/develop/README.md#documentation

seriva (approved):

sk4zuzu (approved):

PR #1932, opened 2021-01-04
rabbitmq: ensure rmq service is running before rabbitmqctl is executed (fix)

sk4zuzu:
Tested on Ubuntu and CentOS.

### DIFF
--- a/core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/main.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/main.yml
@@ -5,6 +5,14 @@
 
 - include_tasks: "install-packages-{{ ansible_os_family | lower }}.yml"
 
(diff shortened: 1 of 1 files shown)


core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/main.yml
```diff
@@ -5,6 +5,14 @@
 
 - include_tasks: "install-packages-{{ ansible_os_family | lower }}.yml"
 
+# NOTE: rabbitmqctl start_app, stop_app, join_cluster commands do NOT work when rabbitmq-server instance is down
+- name: Ensure rabbitmq-server service is running
+  systemd:
```
atsikham, line 10: The `service` module is used in all other places of the role. I would do it the same way.

atsikham (changes requested):

atsikham (approved):

przemyslavic (approved):

PR #1939, opened 2021-01-11
Upgrade Ignite to v2.9.1

sk4zuzu:
This does not take into account `ignite-stateless` k8s app from the `applications` role. That's because it's "stateless" and users can modify versions, IMO we should not upgrade it automatically, because of possible data loss.


mkyc (dismissed):
LGTM

erzetpe (approved):

mkyc (approved):

PR #1962, opened 2021-01-14
Upgrade Vault to latest version

ar3ndt:


plirglo:
What about upgrade procedure for existing clusters? Have we got one for Vault?

ar3ndt:
> What about upgrade procedure for existing clusters? Have we got one for Vault?

not really but it was not used yet I think

erzetpe (dismissed):

atsikham (approved):
If we don't plan to have changes in the upgrade role at the moment, LGTM.

erzetpe (approved):

atsikham (approved):

Issue #1959, opened 2021-01-14
[BUG] No traffic allowed from load balancer subnet to Kubernetes subnet in AWS


przemyslavic:
**Describe the bug**
[AWS] Traffic from the load balancer subnet to the Kubernetes subnet is not allowed, resulting in the HAProxy not being able to forward requests to the backend server.

**How to reproduce**
1. Deploy a cluster with:
- Kubernetes master 
- Kubernetes node
- load balancer
- sample application that will be load balanced

2. Try to reach the application endpoint from the load balancer vm.

**Expected behavior**
The application endpoint is reachable from the load balancer vm.

**Environment**
- Cloud provider: [AWS]
- OS: [all]

**Additional context**
I had to manually modify the network security group by adding a rule to allow traffic coming from subnet `10.1.7.0/24` (load balancer subnet).

przemyslavic:
✅ Load balancing checked. The application deployed on Kubernetes is now accessible from lb machine.
Tested on `AWS`: `Ubuntu` + `RHEL`
Azure was not affected by the issue.

PR #1980, opened 2021-01-18
[BUG] Allow traffic from loadbalancer to kubernetes subnet

rpudlowski93:
PR related to BUG reported in #1959 

- Problem exist only on AWS. On Azure, there is by default rules which allow all traffic in vnet.
- Additional rule added and tested using connection from Loadbalancer instance (HAProxy) to application in kubernetes cluster.

### DIFF
--- a/core/src/epicli/data/aws/defaults/infrastructure/virtual-machine.yml
+++ b/core/src/epicli/data/aws/defaults/infrastructure/virtual-machine.yml
@@ -253,7 +253,7 @@ specification:
        destination_port_range: "0"
        source_address_prefix: "0.0.0.0/0"
        destination_address_prefix: "0.0.0.0/0"
-     - name: load-balacner-subnet-traffic
+     - name: load-balancer-subnet-traffic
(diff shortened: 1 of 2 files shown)


core/src/epicli/data/aws/defaults/infrastructure/virtual-machine.yml
```diff
@@ -253,6 +253,13 @@ specification:
        destination_port_range: "0"
        source_address_prefix: "0.0.0.0/0"
        destination_address_prefix: "0.0.0.0/0"
+     - name: load-balacner-subnet-traffic
```
atsikham, line 256:
```suggestion
     - name: load-balancer-subnet-traffic
```

atsikham (changes requested):

atsikham (changes requested):
Please update changelog

atsikham (approved):

erzetpe (approved):

PR #1988, opened 2021-01-20
Patching cgroup drivers (switch to systemd)

sk4zuzu:


rafzei (dismissed):
The fix was tested by the internal team, after merge we should perform similar stress tests.
Kudos MO!

atsikham (dismissed):
LGTM

sk4zuzu:
I found fatal problem during k8s upgrades so I had to reconsider my solution :) I believe it's correct now!

to-bar (dismissed):
:kimono:

atsikham (approved):
🥇

rafzei (approved):

sk4zuzu:
Please note this PR after testing should be backported at least to 0.9 and 0.8 and would be good also to do it for 0.7 and 0.6. :+1:

PR #1993, opened 2021-01-22
Upgrade ansible to v2.10.5

sk4zuzu:


core/src/epicli/data/common/ansible/playbooks/roles/kafka/tasks/setup-kafka.yml
```diff
@@ -119,9 +119,6 @@
     owner: "{{ specification.kafka_var.user }}"
     group: "{{ specification.kafka_var.group }}"
     mode: 0644
-    src: log4j.properties
-  notify:
-    - restart kafka
```
mkyc, line 124: why removing notify?

sk4zuzu: This task was previously a "template" ansible task, somebody converted it to a "file" ansible task, but chnaged only the "template" string to a "file" string. It's completely incorrect, and this notify has no meaning, it is just a bug tbh :(
If you provide (for "file" module) "dest" and "src" and dont set "state: link" it is big warning in 2.8, but fatal error in 2.10.

mkyc (approved):
Code LGTM, but this change requires extended testing.

core/src/epicli/data/common/ansible/playbooks/kubernetes_node.yml
```diff
@@ -3,6 +3,18 @@
...
+  become_method: sudo
+  pre_tasks:
+    - import_role:
+        name: kubernetes_common
+        tasks_from: gather-facts
+  vars_files:
+    - roles/kubernetes_master/vars/main.yml
+  environment:
+    KUBECONFIG: "{{ kubeconfig.remote }}"
+
```
rpudlowski93, line 17: Why we need to add the common role on kubenetes_master hosts in this playbook? This task is already defined in kubernetes_master playbook so what is the case (just curious :) )?

sk4zuzu: This is kinda weird. It is a bug. In 2.8 it was ignored for some reason, but 2.10 throws fatal error later, because of undefined "kubernetes_common" fact. So this is needed here to populate facts on kubernetes_master inventory hosts inside the hostvars dictionary.

sk4zuzu: Basically without it "kubernetes_node" provisioning fails. :slightly_frowning_face:

core/src/epicli/data/common/ansible/playbooks/applications.yml
```diff
@@ -7,7 +7,6 @@
 
 - hosts: 127.0.0.1
   gather_facts: false
-  connection: local
```
atsikham, line 10: Is that default setting?

core/src/epicli/data/common/ansible/playbooks/kubernetes_node.yml
atsikham: Is it required to run for all hosts in `kubernetes_master` group?

sk4zuzu: Yes I believe so, because it tries to determine if masters are actually a part of the cluster and we cannot assume that any of them is working correctly. We have to understand it. :)

core/src/epicli/data/common/ansible/playbooks/applications.yml
sk4zuzu: So the problem/feature in 2.10 I found is that "delegate_to" does not work correctly when "connection: local" is set. I had to remove it here, because later in the playbook when we try creating posrgtres db, that delegation falls flat on its face.. :)

atsikham (approved):
LGTM

rpudlowski93 (approved):

PR #2000, opened 2021-01-28
Dealing with slash in subscription name for terraform and MS quirkiness

sunshine69:
Azure allow / in the sub name however the API endpoint return wrongly even we
encode the / into %2F. Get
https://management.azure.com/subscriptions/CALM-WFM-Development-Dev%2FTest/providers?api-version=2016-02-01
into your browser and see, azure will return sub not found and the sub name is
the first part of the slash only.

However ID seems to work.

This changes put some more debugging and try to parse and use the ID if found a slash.

### DIFF
--- a/core/src/epicli/cli/engine/terraform/TerraformRunner.py
+++ b/core/src/epicli/cli/engine/terraform/TerraformRunner.py
@@ -16,7 +16,7 @@ class TerraformRunner(Step):
         self.new_env = os.environ.copy()
         self.terraform.init(env=self.new_env)
         if self.cluster_model.provider == 'azure':
-            self.azure_login()        
+            self.azure_login()
 
     def __enter__(self):
         super().__enter__()
@@ -30,9 +30,9 @@ class TerraformRunner(Step):
 
     def delete(self):
         self.terraform.destroy(auto_approve=True, env=self.new_env)
-        
+
     def azure_login(self):
-        # From the 4 methods terraform provides to login to 
+        # From the 4 methods terraform provides to login to
         # Azure we support (https://www.terraform.io/docs/providers/azurerm/auth/azure_cli.html):
         # - Authenticating to Azure using the Azure CLI
         # - Authenticating to Azure using a Service Principal and a Client Secret
@@ -52,22 +52,29 @@ class TerraformRunner(Step):
                 # Create the service principal, for now we use the default subscription
                 self.logger.info('Creating service principal')
                 cluster_name = self.cluster_model.specification.name.lower()
-                cluster_prefix = self.cluster_model.specification.prefix.lower()               
+                cluster_prefix = self.cluster_model.specification.prefix.lower()
                 resource_group_name = resource_name(cluster_prefix, cluster_name, 'rg')
                 sp = apiproxy.create_sp(resource_group_name, subscription['id'])
                 sp['subscriptionId'] = subscription['id']
                 save_sp(sp, self.cluster_model.specification.name)
             else:
-                self.logger.info('Using service principal from file')
+                self.logger.info(f'Using service principal from file {sp_file}')
                 sp = load_yaml_file(sp_file)
 
             # Login as SP and get the default subscription.
             subscription = apiproxy.login_sp(sp)
 
             if 'subscriptionId' in sp:
-                # Set active subscription if sp contains it. 
+                # Set active subscription if sp contains it.
+                self.logger.info(f"subscriptionId from sp file: {sp['subscriptionId']}")
                 apiproxy.set_active_subscribtion(sp['subscriptionId'])
-                self.new_env['ARM_SUBSCRIPTION_ID'] = sp['subscriptionId']
+                arm_subscription_id = ""
+                if "/" in sp['subscriptionId']:
+                    self.logger.info(f"WARN Slash detected in the subscription name {sp['subscriptionId']}. Will parse the ID and use it instead")
+                    arm_subscription_id = [x['id'] for x in subscription if x['name'] == sp['subscriptionId']][0]
+                else:
+                    arm_subscription_id = sp['subscriptionId']
+                self.new_env['ARM_SUBSCRIPTION_ID'] = arm_subscription_id
             else:
                 # No subscriptionId in sp.yml so use the default one from Azure SP login.
                 self.new_env['ARM_SUBSCRIPTION_ID'] = subscription[0]['id']
--- a/core/src/epicli/cli/engine/terraform/TerraformRunner.py
+++ b/core/src/epicli/cli/engine/terraform/TerraformRunner.py
@@ -66,11 +66,11 @@ class TerraformRunner(Step):
 
             if 'subscriptionId' in sp:
                 # Set active subscription if sp contains it.
-                self.logger.info(f"subscriptionId from sp file: {sp['subscriptionId']}")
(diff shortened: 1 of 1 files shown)


seriva (changes requested):
Dont think we should be logging any kind of SP related data.

sunshine69:
> Dont think we should be logging any kind of SP related data.

Ouch! Will remove it :)

sunshine69:
> Dont think we should be logging any kind of SP related data.

Hey it is just subscriptions! Is it regarded as secrets though?

seriva (dismissed):

core/src/epicli/cli/engine/terraform/TerraformRunner.py
```diff
@@ -52,22 +52,29 @@ def azure_login(self):
...
             if 'subscriptionId' in sp:
-                # Set active subscription if sp contains it. 
+                # Set active subscription if sp contains it.
+                self.logger.debug(f"subscriptionId from sp file: {sp['subscriptionId']}")
                 apiproxy.set_active_subscribtion(sp['subscriptionId'])
-                self.new_env['ARM_SUBSCRIPTION_ID'] = sp['subscriptionId']
+                arm_subscription_id = ""
+                if "/" in sp['subscriptionId']:
+                    self.logger.debug(f"WARN Slash detected in the subscription name {sp['subscriptionId']}. Will parse the ID and use it instead")
+                    arm_subscription_id = [x['id'] for x in subscription if x['name'] == sp['subscriptionId']][0]
```
atsikham: What's about usage of [generator expressions](https://www.python.org/dev/peps/pep-0289/) here? Something like this (not tested):

```python
arm_subscription_id = next((x['id'] for x in subscription if x['name'] == sp['subscriptionId']), None)
```

Please change the name of "x" to something valuable.

The number of lines could be reduced:

```python
arm_subscription_id = sp['subscriptionId']
if "/" in arm_subscription_id:
  self.logger.debug(f"WARN Slash detected in the subscription name {sp['subscriptionId']}. Will parse the ID and use it instead")
  arm_subscription_id = next(x['id'] for x in subscription if x['name'] == sp['subscriptionId'])
self.new_env['ARM_SUBSCRIPTION_ID'] = arm_subscription_id
```

atsikham (changes requested):

core/src/epicli/cli/engine/terraform/TerraformRunner.py
sunshine69: Interesting new feature (at least new to me), I start looking into it and see if it work. brb.

In terms of readability and simplicity I still prefer a normal list comprehension though.

sunshine69: Seems to work
```
In [1]: l0 = [1,2,3,4,5]                                                                                                               

In [2]: y = next(x for x in l0 if x == 4)                                                                                              

In [3]: y                                                                                                                              
Out[3]: 4

In [4]: y = next((x for x in l0 if x == 4), None)                                                                                      

In [5]: y                                                                                                                              
Out[5]: 4
```

atsikham (approved):

seriva (approved):

core/src/epicli/cli/engine/terraform/TerraformRunner.py
```diff
@@ -52,22 +52,27 @@ def azure_login(self):
...
             subscription = apiproxy.login_sp(sp)
 
             if 'subscriptionId' in sp:
-                # Set active subscription if sp contains it. 
+                # Set active subscription if sp contains it.
+                self.logger.debug(f"subscriptionId from sp file: {sp['subscriptionId']}")
                 apiproxy.set_active_subscribtion(sp['subscriptionId'])
-                self.new_env['ARM_SUBSCRIPTION_ID'] = sp['subscriptionId']
+                arm_subscription_id = sp['subscriptionId']
+                if "/" in sp['subscriptionId']:
```
mkyc, line 72: `subscriptionId` is always UUID. Why would you have slash in it?

core/src/epicli/cli/engine/terraform/TerraformRunner.py
```diff
@@ -52,22 +52,27 @@ def azure_login(self):
...
+                    self.logger.debug(f"WARN Slash detected in the subscription name {sp['subscriptionId']}. Will parse the ID and use it instead")
+                    arm_subscription_id = next((_sub['id'] for _sub in subscription if _sub['name'] == sp['subscriptionId']), None)
```
mkyc, line 74: AFAIU you are fixing incorrectly created service principal file. I assume that service principal file you are using has `subscriptionId` and `name` fields filled with name.

mkyc (changes requested):
I do not understand that PR. It looks that service principal used is incorrect.

core/src/epicli/cli/engine/terraform/TerraformRunner.py
sunshine69: Please review the existing code as I don't have much time on this, but you will see we accept two types of ID, one UUID and the other format which is the subscription name.

It is just takes it too long a simple change like this with all test and uses cases presented!

mkyc: Sorry for inconvenience and late notice but this PR just looks incorrect. Putting `name` into `subscriptionId` field looks like misuse. Service principal file is documented [here](https://github.com/epiphany-platform/epiphany/blob/88cfb9e559f2cdd20c39d2d072d6436f4bed5c39/docs/home/howto/CLUSTER.md#how-to-create-an-epiphany-cluster-on-a-cloud-provider) in point 3: 

> Where the clustername is the name you specified under specification.name in your cluster yaml. Then in the terraform folder add the file named sp.yml and fill it with the service priciple information like so:
> 
> ```yaml
> appId: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxx"
> displayName: "app-name"
> name: "http://app-name"
> password: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxx"
> tenant: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxx"
> subscriptionId: "xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxx"
> ```

From PR correctness point of view, we tested this change and if we put name without `/` in it, this change will not work, so even if I didn't mind misusing field `if "/" in sp['subscriptionId']` is incorrect and some "validate if is UUID" should be placed here instead.

sunshine69: It is up to the team. At the time in ABB RND team, we hit the problems, our sp.yaml is correct (in terms of correctness) and the problems is describe in this pr, and has been fixed by using this change PR.

That is the fact. If you don't ACKed it then when the RND team move to new version they will hit it again (if they do not use my custom build image/branch which has the fix)

I no longer work for ABB thus I do not have time to debug anymore. From my memory by tracing the code, epiphany accept two format, the UUID or the name - and when the name has slash the azure behavior makes it errors. Read again about my description.

I do it with best of considerations and judgement at the time and I retain. If problem goes away then it might be Azure has fixed their issues (that is if they allow slash in name then in API request they would interpret the encode of slash %F02 to be a string rather than a API path separator).

mkyc:
Putting `name` into `subscriptionId` field is incorrect.

PR #2009, opened 2021-02-01
Update kafka to version 2.6.1

sunshine69:
Upstream update


atsikham (changes requested):
Changelog and COMPONENTS.md must be updated

seriva:
@sunshine69 Can you look at the requested changes?

mkyc:
Since it is quite old and doesn't include upgrade procedure we probably want to close it without merge, but we should discuss it before we do.

mkyc:
We agreed together that it needs to be closed.

PR #2040, opened 2021-02-09
Support Azure RHEL LVM images

to-bar:
Rebootless solution which requires cloud-init.
It merges all logical volumes except usrlv (mounted on /usr), which requires reboot (or `systemctl switch-root`). 
So after merging there are two LVs:
```
$ df -h | grep lv
/dev/mapper/rootvg-rootlv   54G  264M   53G   1% /
/dev/mapper/rootvg-usrlv    10G  1.4G  8.7G  14% /usr
```

Feature can be disabled using `infrastructure/custom-data` doc. 

Tested with images:
- RedHat:RHEL:7lvm-gen2:latest
- RedHat:RHEL:8-LVM:latest


core/src/epicli/cli/helpers/build_saver.py
```diff
@@ -51,13 +52,9 @@ def save_ansible_config_file(ansible_config_file_settings, ansible_config_file_p
...
 
 
-# method cleans generated .tf files (not tfstate)
-def clear_terraform_templates(cluster_name):
-    terraform_dir = get_terraform_path(cluster_name)
-    files = os.listdir(terraform_dir)
-    for file in files:
-        if file.endswith(".tf"):
-            os.remove(os.path.join(terraform_dir, file))
+def remove_files_matching_glob(dir_path, pattern):
```
seriva, line 55: 👏🏻

core/src/epicli/data/azure/defaults/infrastructure/custom-data.yml
```diff
@@ -0,0 +1,8 @@
+kind: infrastructure/custom-data
```
seriva: Really wonder if we should go through the trouble of having this custom-data template as its only used to set the ```use_custom_data```  value in the ```virtual-machine.yml```.  

IMO just having a flag inside the ```virtual-machine.yml``` would suffice, which can also have a understandable name for the task it would be used for. This makes configuration confusing imo.

core/src/epicli/data/azure/validation/infrastructure/custom-data.yml
```diff
@@ -0,0 +1,4 @@
+properties:
```
seriva: See comment on ```core/src/epicli/data/azure/defaults/infrastructure/custom-data.yml```. I think this might not be needed.

core/src/epicli/data/azure/defaults/infrastructure/virtual-machine.yml
```diff
@@ -12,6 +12,7 @@ specification:
   tags: []
   os_type: linux
   size: Standard_DS1_v2
+  use_custom_data: SET_BY_AUTOMATION
```
seriva: See comment on core/src/epicli/data/azure/defaults/infrastructure/custom-data.yml. I think we can get by with only this flag with a more relevant name (redhat-lvm-merge or something), set to true.

core/src/epicli/cli/helpers/build_saver.py
```diff
@@ -51,13 +52,9 @@ def save_ansible_config_file(ansible_config_file_settings, ansible_config_file_p
...
+    for file in Path(dir_path).glob(pattern):
+        file.unlink()
```
erzetpe, line 57: why not using remove here?

core/src/epicli/data/common/tests/Rakefile
```diff
@@ -39,16 +40,24 @@ File::open(ENV['inventory']) do |f|
     if line =~ /^\[([^\]]+)\]/
       current_group = $1
     elsif line =~ /(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})/ # regex for IP address
+      host_ip = $1
+      host_name = line.split.first
       if current_group
         groups[current_group] ||= {}
-        groups[current_group][$1] = line.split.first
+        groups[current_group][host_ip] = host_name  # e.g. "zookeeper"=>{"18.212.26.47"=>"ec2-18-212-26-47.compute-1.amazonaws.com"}
```
erzetpe: Please remove this reference to some possible real address, use some fake as an example.

seriva (changes requested):

erzetpe (dismissed):

core/src/epicli/data/common/tests/Rakefile
erzetpe: Done.

core/src/epicli/data/azure/defaults/infrastructure/custom-data.yml
to-bar: Let me explain why I've chosen such approach, then we can decide together whether it is overkill or not 😉 

1. I wanted to be able to disable cloud-init (custom-data) globally - i.e. just by changing a single setting (one line) in config file. For this reason the setting is populated from one place `infrastructure/custom-data` to all `infrastructure/virtual-machine` docs. With the alternative approach it seems it would require to change the value of `use_custom_data` for each component (VM type) and if you forget about any, you will get inconsistent environment. I run a test trying to change OS distro globally using the following docs but it was not possible - i.e. I got VM with Ubuntu.

```
kind: infrastructure/virtual-machine
name: default
provider: azure
specification:
  storage_image_reference:
    publisher: RedHat
    offer: RHEL
    sku: 7lvm-gen2
    version: latest
---
kind: infrastructure/virtual-machine
name: repository-machine-rhel
provider: azure
based_on: repository-machine
specification:
  size: Standard_DS2_v2
```
  
2. The current solution is not intuitive but quite generic (easily extendable). For example, you can add another `cloud-config` for Ubuntu, which can do something different than merging LVM. Also, it is possible to use not only `cloud-config` type but any [User-Data format](https://cloudinit.readthedocs.io/en/latest/topics/format.html) including `MIME multipart` file, which allows to specify more than one type of data. However, I don't know whether we will use cloud-init for something else than merging LVM in the future, since we prefer Ansible, so maybe this point is not meaningful.

If the current approach is not intuitive enough but we will decide to keep it, I can document how it works.

core/src/epicli/data/azure/validation/infrastructure/custom-data.yml
to-bar: See my answer for the first comment.

core/src/epicli/data/azure/defaults/infrastructure/custom-data.yml
seriva: Both fair points, so lets start with adding some documentation for sure. 

And maybe a different name for ```infrastructure/custom-data``` as for me this seems to generic?  ```infrastructure/cloud-init``` or something like that? At least that will return something usefull in google if people are looking at that and are to lazy to read the documentation:)

image

to-bar: Suggestion applied, kind renamed to `infrastructure/cloud-init-custom-data` :relaxed:

erzetpe (approved):

seriva (approved):

PR #2047, opened 2021-02-10
Update test checking ownership of /etc/prometheus

to-bar:


rafzei (dismissed):
I see no difference from a practical point of view. In my opinion, it's always good to have more strict rights if they work. If our tests are failing and we want to stick with the same approach as in Grafana role, let's merge it :)

przemyslavic (approved):

plirglo (approved):

rafzei:
There were few things we discover checking available options. I will create PR (Prometheus part3) and will verify and compare them. For now, we decided to adjust our tests and leave the owner as it is.

Issue #1901, opened 2020-12-09
[SPIKE] Make Epiphany upgrades selective (Kafka)


sk4zuzu:
**Description**

In this spike we want to consider an option with selective Kafka upgrades by Epicli. 

It should be relatively simple (likely requires no changes in upgrade procedures themselves) to refactor Epiphany upgrades and make it possible for the user to control what components will be upgraded. That may bring value not only to users with legacy clusters but also to the process of testing.

*  It needs to be checked if Kafka supports upgrades from any version to any other version. If it requires incremental upgrades between versions, it have to be taken into account.
* The output of this task is a PoC with selective Kafka upgrade by epicli. 
* If it's possible, design document can be created.

All the topics related to the Kafka module will be done separately and covered by #1976 task and tasks that will be created as a result of #1976.

**Additional information**

The original task contained additional options related to Kafka module, but to avoid dependencies and clarify a single topic per task, these options are not covered here:

```text
2. In the near future we will start introducing modules that will replace classic Epiphany components (like Kafka clusters etc). If we'd do that soon enough then the question would be if there is a possiblity to reuse modules with exising legacy on-prem clusters and what needs to be done to achieve that.

3. Support both 1. and 2.

We'd like to understand how much work do these options require to be implemented.

We want a design-doc out of his research. 
```

sk4zuzu:
To clarify this one is about preparing some Kafka solution for very old releases of Epiphany :). For customers that do not want or it's not easy for them to upgrade to newer versions.

sk4zuzu:
It looks like this is the same really as #1894 :hand_over_mouth:

DaOsie:
In my opinion there are 3 separate spikes "secretly" coded in the description:
 - checking whether selective upgrades of components are possible in the legacy versions of epiphany
 - checking whether modules will be able to be used in previous epiphany versions that do not use modules yet
 - find the best way to introduce kafka for older releases of Epiphany

At first sight, they seem to be connected with each other, however I think they aim for 3 different things. Shouldn't we just check the easiest way for introducing kafka on legacy clusters?

mkyc:
This one probably depends on #1976

atsikham:
As I described in [this](https://github.com/epiphany-platform/epiphany/issues/1976#issuecomment-771525428) comment, we still have to investigate a lot related to the Kafka module. All things related to modules including comparison with Kafka component of epicli will be done in the scope of #1976 and other tasks that will be created after #1976 is done.

[...]

PR #2064, opened 2021-02-16
initial version of selective components upgrade option [#1901]

ar3ndt:


docs/home/howto/UPGRADE.md
```diff
@@ -107,7 +107,13 @@ To upgrade the cluster components run the following steps:
 
 ### Additional parameters
 
-The `epicli upgrade` command had an additional flag `--wait-for-pods`. When this flag is added, the Kubernetes upgrade will wait until all pods are in the **ready** state before proceding. This can be usefull when a zero downtime upgrade is required. **Note: that this can also cause the upgrade to hang indefinitely.**
+The `epicli upgrade` command had an additional flags:
+-  `--wait-for-pods`. When this flag is added, the Kubernetes upgrade will wait until all pods are in the **ready** state before proceding. This can be usefull when a zero downtime upgrade is required. **Note: that this can also cause the upgrade to hang indefinitely.**
+- `--upgrade-components`. Specify comma separated component names so the upragde procedure will only process specific ones. If this flag is not provided then upgrade procedure will process all the components.
```
erzetpe: What if the flag is set and no components are specified?

erzetpe (changes requested):

erzetpe (approved):

core/src/epicli/cli/epicli.py
```diff
@@ -252,6 +258,10 @@ def upgrade_parser(subparsers):
                             help="Waits for all pods to be in the 'Ready' state before proceeding to the next step of the K8s upgrade.")
     optional.add_argument('--profile-ansible-tasks', dest='profile_ansible_tasks', action="store_true",
                             help='Enable Ansible profile_tasks plugin for timing tasks. (developer/debug option)')
+    optional.add_argument('--upgrade-components', dest='upgrade_components', type=str, required=False,
```
seriva: Might I suggest to only have the ```--upgrade-components``` which should be default to something like ```"all"``` to specify to just upgrade all components? if its not all its a list of specific components you want to upgrade and there will be no need for ```--upgrade-all```.

Finally check the ```choices``` argument for ```add_argument``` if you can add some validation for ```--upgrade-components``` argument.

seriva (changes requested):

core/src/epicli/cli/epicli.py
erzetpe: I agree with Luuk, probably this should work like this:
1. not pass --upgrade-components - this will work like upgrade all
2. --upgrade-components - this will fail if you won't specify components and upgrade only specified components

erzetpe (dismissed):
@seriva made sound comment about selective upgrade behavior.

core/src/epicli/cli/epicli.py
ar3ndt: it was the idea initially but changed. Wil amend it.

atsikham:
@ar3ndt @rpudlowski93 @erzetpe @seriva what do you think about making componet names case-insensitive for epicli? I would like to have no error if `Kafka` specified instead of `kafka`.

rpudlowski93:
> @ar3ndt @rpudlowski93 @erzetpe @seriva what do you think about making componet names case-insensitive for epicli? I would like to have no error if `Kafka` specified instead of `kafka`.

In my opinion It is definitely a good idea, I would like to have no errors in this scenario.

ar3ndt:
> > @ar3ndt @rpudlowski93 @erzetpe @seriva what do you think about making componet names case-insensitive for epicli? I would like to have no error if `Kafka` specified instead of `kafka`.
> 
> In my opinion It is definitely a good idea, I would like to have no errors in this scenario.

sure will do

core/src/epicli/cli/epicli.py
seriva: > > > @ar3ndt @rpudlowski93 @erzetpe @seriva what do you think about making componet names case-insensitive for epicli? I would like to have no error if `Kafka` specified instead of `kafka`.
> > 
> > 
> > In my opinion It is definitely a good idea, I would like to have no errors in this scenario.
> 
> sure will do

IMO I don't mind either way. Kafka/kafka is the same for me. More important would be to add proper validation to the Epicli upgrade command what to accept and **validate** a , or space separated array. Argparse does not support that but a custom type would do the trick. Something along the lines of (not tested):

```python
def upgrade_parser(subparsers):
    sub_parser = subparsers.add_parser('upgrade',
                                       description='Upgrades common and K8s components of an existing Epiphany Platform cluster.')
    optional = sub_parser._action_groups.pop()
    required = sub_parser.add_argument_group('required arguments')

    component_list = ['all', 'kubernetes', 'kafka']
    def comma_seperated_type(choices):
        """Return a function that splits and checks comma-separated values."""
        def splitarg(arg):
            values = arg.split(',')
            for value in values:
                if value not in choices:
                    raise argparse.ArgumentTypeError(
                        'invalid choice: {!r} (choose from {})'
                        .format(value, ', '.join(map(repr, choices))))
            return values
        return splitarg

    #required
    required.add_argument('-b', '--build', dest='build_directory', type=str, required=True,
                            help='Absolute path to directory with build artifacts.')

    #optional
    optional.add_argument('--offline-requirements', dest='offline_requirements', type=str, required=False,
                            help='Path to the folder with pre-prepared offline requirements.')
    optional.add_argument('--wait-for-pods', dest='wait_for_pods', action="store_true",
                            help="Waits for all pods to be in the 'Ready' state before proceeding to the next step of the K8s upgrade.")
    optional.add_argument('--profile-ansible-tasks', dest='profile_ansible_tasks', action="store_true",
                            help='Enable Ansible profile_tasks plugin for timing tasks. (developer/debug option)')
    optional.add_argument('--upgrade-components', dest='upgrade_components', type=comma_seperated_type(component_list), required=False,
                            help='Provides comma seperated list of components for upgrade selected from the following: [' + ','.join(map(str, component_list)) + ']')
    
    def run_upgrade(args):
        adjust_paths_from_build(args)
        with UpgradeEngine(args) as engine:
            return engine.upgrade()

    sub_parser.set_defaults(func=run_upgrade)
```

Also while you are making these changes maybe also straighten out the _- usage in the task file names:
image

seriva (dismissed):

seriva (dismissed):

core/src/epicli/data/common/ansible/playbooks/upgrade.yml
```diff
@@ -14,17 +22,10 @@
         name: upgrade
         tasks_from: kubernetes/gather-static-facts
       run_once: true
+      when: ('"kubernetes_master" in upgrade_components') or upgrade_components|length == 0
```
atsikham: Why `kubernetes_master` is used here when in `component_list` there is only `kubernetes` value?

core/src/epicli/data/common/ansible/playbooks/upgrade.yml
```diff
@@ -81,6 +86,7 @@
         name: upgrade
         tasks_from: kubernetes
       vars: { ver: "1.18.6", cni_ver: "0.8.6", upgrade_to_final_version: true }
+      when: ('"kubernetes_master" in upgrade_components') or upgrade_components|length == 0
```
atsikham: Here and above upgrade is performed when `kubernetes_master` is in components list, but upgrade is also done for `kubernetes_nodes`.

docs/home/howto/UPGRADE.md
```diff
@@ -107,7 +107,13 @@ To upgrade the cluster components run the following steps:
 
 ### Additional parameters
 
-The `epicli upgrade` command had an additional flag `--wait-for-pods`. When this flag is added, the Kubernetes upgrade will wait until all pods are in the **ready** state before proceding. This can be usefull when a zero downtime upgrade is required. **Note: that this can also cause the upgrade to hang indefinitely.**
+The `epicli upgrade` command had an additional flags:
```
atsikham:
```suggestion
The `epicli upgrade` command has additional flags:
```

docs/home/howto/UPGRADE.md
```diff
@@ -107,7 +107,13 @@ To upgrade the cluster components run the following steps:
...
+-  `--wait-for-pods`. When this flag is added, the Kubernetes upgrade will wait until all pods are in the **ready** state before proceding. This can be usefull when a zero downtime upgrade is required. **Note: that this can also cause the upgrade to hang indefinitely.**
+- `--upgrade-components`. Specify comma separated component names so the upgrade procedure will only process specific ones. List cannot be empty, otherwise execution will fail. By default upgrade will process all components if this parameter is not provided
```
atsikham, line 112: Was the empty or not specified value tested?

atsikham (changes requested):
In addition to the left comments could you update `launch.json` file with an example?

atsikham (approved):

rpudlowski93 (approved):
LGTM, all the mentioned cases applied

Issue #2032, opened 2021-02-09
Ensure that there are packages for all required components available for ARM architecture


mkyc:
- Analyze all packages required during Epiphany deployment process and confirm that ARM architecture is supported by them.
- Find replacement for not supported packages (if any)
- Use requirements.txt files as base

mkyc:
Here is script to setup ARM machine in AWS: https://github.com/mkyc/aws-arm-simple-setup

ar3ndt:
Not all packages are available. These are missing:

elasticsearch-curator-5.8.3
opendistroforelasticsearch-kibana-1.10.1*
pgbouncer-1.10.*
rh-haproxy18
rh-haproxy18-haproxy-syspaths
repmgr10-4.0.6

epel that works for aarch64:

`[epel]
name=Epel for arm
baseurl=http://dl.fedoraproject.org/pub/epel/7/aarch64
enabled=1
gpgcheck=0`

ar3ndt:
other [PR](https://github.com/epiphany-platform/epiphany/pull/2079) includes changes from this task

przemyslavic:
It looks like a spike result. Nothing to test at this point.

PR #2079, opened 2021-02-24
ARM support design doc

seriva:
- Added design doc for adding ARM support for CentOS and certain components.
- Added partial research for Centos requirements for ARM #2032 (Not complete but useful in the context.).


docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,86 @@
...
+
+1. Add “architecture” flag when using Epicli
+2. Add new OS (CentosARM64 fe.)
+
+Have 2 big disadvanges from the start:
+
+1. Will require an additional input which makes things more confusing as they will need supply not only the OS but also Architecture for (offline) install. This should not be needed as we can detect the architecture we are working on, on all required levels.
+2. Does not require additional input but this will lead to code duplication in the ```requirements``` role as we need to maintain ```download-requirements.sh``` for each OS and architecture then.
+
+That is why I opt for an approach where we don`t add any architecture flag or new additional OS. The architecture we can handle on the code level and on the OS level only the ```requirements.txt``` might be different for each as indicated by initial research [here.](./centos-arm-analysis.md).
```
mkyc:
```suggestion
That is why I opt for an approach where we don't add any architecture flag or new additional OS. The architecture we can handle on the code level and on the OS level only the ```requirements.txt``` might be different for each as indicated by initial research [here.](./centos-arm-analysis.md).

```

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,86 @@
...
+## Changes required
+
+### Repostitory role
+
+In the repository role we need to change the download of the requirements to support additional architectures as download requirements might be different as:
+
+- Some components/roles might not have packages/binaries/containers that support ARM
+- Some filenames for binaries will be different per architecture
+
+Hense we should make a requirements.txt for each architecture we want to support, for example:
```
mkyc:
```suggestion
Hence we should make a requirements.txt for each architecture we want to support, for example:

```

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,86 @@
...
+
+- requirements_x86_64.txt (Should be the default and present)
+- requirements_aarch64.txt
```
mkyc:
```suggestion
- requirements_arm64.txt

```

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,86 @@
...
+
+As fallback we could just use ```requirements.txt``` and asume its ```x86_64```.
+
+### Download role
+
+In the download role, which is used to download plain files from the repository, we should add support for filenames with aliases for different architectures:
+
+For example select between:
+
+- haproxy_exporter-0.12.0.linux-```amd64```.tar.gz
```
mkyc:
```suggestion
- haproxy_exporter-0.12.0.linux-```x86_64```.tar.gz
```

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,86 @@
...
+- haproxy_exporter-0.12.0.linux-```arm64```.tar.gz
+
+Based on ```ansible_architecture```.
+
+**Note that this should be optional as some packages dont require the use of an alias like Java bases ones for example.**
+
+### Artitecture support for each component/role
+
+As per current requirements not every Epiphany component is required to support ARM and there might be cases that a component/role can`t support ARM as indicated by initial research [here.](./centos-arm-analysis.md).
```
mkyc:
```suggestion
As per current requirements not every Epiphany component is required to support ARM and there might be cases that a component/role can't support ARM as indicated by initial research [here.](./centos-arm-analysis.md).

```

mkyc (commented):
Just little formatting comments while I was reading this.

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,86 @@
...
+
+Affected version: 0.10.x/0.11.x
+
+## Goals
+
+This document outlines an aproach to add (partial) ARM support to Epicli. Not finalized requirements so far:
+
+- ARMv8/ARM64 architecture
+- Centos 7 as operating system
+- "any" provider as we do not want to provide ARM infrastructure on any cloud providers jet through Terraform
```
erzetpe:
```suggestion
- "any" provider as we do not want to provide ARM infrastructure on any cloud providers yet through Terraform

```

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,86 @@
...
+## Approach
+
+The 2 high level approaches that have been opted so far:
+
+1. Add “architecture” flag when using Epicli
+2. Add new OS (CentosARM64 fe.)
+
+Have 2 big disadvanges from the start:
+
+1. Will require an additional input which makes things more confusing as they will need supply not only the OS but also Architecture for (offline) install. This should not be needed as we can detect the architecture we are working on, on all required levels.
```
erzetpe, line 24: Are you assuming that the server where we will be downloading packages for install will have the same architecture as the target one?

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,86 @@
...
+
+- Some components/roles might not have packages/binaries/containers that support ARM
+- Some filenames for binaries will be different per architecture
+
+Hense we should make a requirements.txt for each architecture we want to support, for example:
+
+- requirements_x86_64.txt (Should be the default and present)
+- requirements_aarch64.txt
+
+The ```download-requirements.sh``` script should be able to figure out which one to select based on the output of:
```
erzetpe, line 45: What if the server where we will be downloading packages will have different architecture than targeted one?

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,86 @@
...
+In the download role, which is used to download plain files from the repository, we should add support for filenames with aliases for different architectures:
+
+For example select between:
+
+- haproxy_exporter-0.12.0.linux-```amd64```.tar.gz
+- haproxy_exporter-0.12.0.linux-```arm64```.tar.gz
+
+Based on ```ansible_architecture```.
+
+**Note that this should be optional as some packages dont require the use of an alias like Java bases ones for example.**
```
erzetpe:
```suggestion
**Note that this should be optional as some packages don't require the use of an alias like Java bases ones for example.**

```

erzetpe (changes requested):

docs/design-docs/arm/arm.md
seriva: Yes, the download of requirements is already dependent on OS and for architecture the requirement will be the same. Packages you need to pull are architecture dependent.

docs/design-docs/arm/arm.md
seriva: We are not going to support combinations of architectures in a cluster. The same that currently all machines in a cluster need to run the same OS.

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,118 @@
+# Epiphany Platform backup design document
```
to-bar: backup? :smiley:

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Goals
+
+This document outlines an aproach to add (partial) ARM support to Epicli. Not finalized requirements so far:
+
+- ARMv8/ARM64 architecture
+- Centos 7 as operating system
+- "any" provider as we do not want to provide ARM infrastructure on any cloud providers yet through Terraform
+
+- Epiphany components: K8s, Kafka, ELK, Grafana, Prometheus, Kibana, HAProxy, Keycloak, PostgreSQL
```
to-bar:
```suggestion
- Epiphany components: K8s, Kafka, ELK, Grafana, Prometheus, Kibana, HAProxy, Keycloak, PostgreSQL, Docker (local image registry)

```

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,118 @@
...
+
+The 2 high level approaches that have been opted so far:
+
+1. Add “architecture” flag when using Epicli
+2. Add new OS (CentosARM64 fe.)
+
+Have 2 big disadvanges from the start:
+
+1. Will require an additional input which makes things more confusing as they will need supply not only the OS but also Architecture for (offline) install. This should not be needed as we can detect the architecture we are working on, on all required levels.
+2. Does not require additional input but this will lead to code duplication in the ```requirements``` role as we need to maintain ```download-requirements.sh``` for each OS and architecture then.
```
to-bar:
```suggestion
2. Does not require additional input but this will lead to code duplication in the ```repository``` role as we need to maintain ```download-requirements.sh``` for each OS and architecture then.

```

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+## Changes required
+
+### Repostitory role
+
+In the repository role we need to change the download of the requirements to support additional architectures as download requirements might be different as:
+
+- Some components/roles might not have packages/binaries/containers that support ARM
+- Some filenames for binaries will be different per architecture
```
to-bar, line 37:
```suggestion
- Some filenames for binaries will be different per architecture
- Some package repositories will have different URLs per architecture
```

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+- requirements_x86_64.txt (Should be the default and present)
+- requirements_arm64.txt
+
+
+The ```download-requirements.sh``` script should be able to figure out which one to select based on the output of:
+
+```shell
+uname -a
```
to-bar:
```suggestion
uname -i

```

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,118 @@
...
+
+```shell
+arch
+```
+
+As fallback we could just use ```requirements.txt``` and asume its ```x86_64```.
+
+### Download role
+
+In the download role, which is used to download plain files from the repository, we should add support for filenames with aliases for different architectures:
```
to-bar:
```suggestion
In the download role, which is used to download plain files from the repository, we should add support for filename patterns and automatically look for current architecture (optionally with regex based suffix like `linux[_-]amd64\.(tar\.gz|tar|zip)`):

```

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,118 @@
...
+
+For example select between:
+
+- haproxy_exporter-0.12.0.linux-```x86_64```.tar.gz
+- haproxy_exporter-0.12.0.linux-```arm64```.tar.gz
+
+Based on ```ansible_architecture```.
```
to-bar:
```suggestion
Based on ```ansible_architecture``` fact.

```

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,118 @@
...
+
+**Note that this should be optional as some packages don't require the use of an alias like Java bases ones for example.**
```
to-bar:
```suggestion
**Note that this should be optional as some filenames do not contain architecture like Java based packages for example.**

```

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,118 @@
...
+Thats why every component/role should be marked which architecture it supports. Maybe something in ```<rolename>/defaults/main.yml``` like:
+
+```yml
+supported_architectures:
+  - all ?
+  - x86_64
+  - arm64
+```
+
+We can assume the role/component will support everyting if ```all``` is defined or if ```supported_architectures``` is not present.
```
to-bar:
```suggestion
We can assume the role/component will support everything if ```all``` is defined or if ```supported_architectures``` is not present.

```

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Pre-flight check
+
+The ```preflight``` should be expended to check if all the components/roles we want to install from the inventory actually support the architecture we want to use. We should be able todo this with the definition from the above point. This way we will make sure people can only install components on ARM which we actually support.
```
to-bar:
```suggestion
The ```preflight``` should be expanded to check if all the components/roles we want to install from the inventory actually support the architecture we want to use. We should be able to do this with the definition from the above point. This way we will make sure people can only install components on ARM which we actually support.

```

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Pre-flight check
+
+The ```preflight``` should be expended to check if all the components/roles we want to install from the inventory actually support the architecture we want to use. We should be able todo this with the definition from the above point. This way we will make sure people can only install components on ARM which we actually support.
+
```
to-bar:
```suggestion
```

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,118 @@
...
+1. This tool can do the same as Skopeo:
+
+```bash
+./skopeo --insecure-policy copy docker://kubernetesui/dashboard:v2.0.3 docker-archive:skopeodashboard:v2.0.3
+./crane export --insecure kubernetesui/dashboard:v2.0.3 "cranedashboard:v2.0.3"
+```
+
+The above will produce the same Docker image package.
+
+2. Supports the universal cross distro binairy.
```
to-bar:
```suggestion
2. Supports the universal cross distro binary.

```

to-bar (changes requested):

docs/design-docs/arm/arm.md
seriva: Fixed copy paste error.

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,124 @@
...
+1. Add “architecture” flag when using Epicli
+2. Add new OS (CentosARM64 fe.)
+
+Have 2 big disadvanges from the start:
+
+1. Will require an additional input which makes things more confusing as they will need supply not only the OS but also Architecture for (offline) install. This should not be needed as we can detect the architecture we are working on, on all required levels.
+2. Does not require additional input but this will lead to code duplication in the ```repository``` role as we need to maintain ```download-requirements.sh``` for each OS and architecture then.
+
+
+That is why I opt for an approach where we don't add any architecture flag or new additional OS. The architecture we can handle on the code level and on the OS level only the ```requirements.txt``` might be different for each as indicated by initial research [here.](./centos-arm-analysis.md).
```
to-bar:
```suggestion
That is why I opt for an approach where we don't add any architecture flag or new additional OS. The architecture we can handle on the code level and on the OS level only the ```requirements.txt``` might be different for each as indicated by initial research [here](./centos-arm-analysis.md).

```

docs/design-docs/arm/arm.md
```diff
@@ -0,0 +1,124 @@
...
+Based on ```ansible_architecture``` fact.
+
+
+**Note that this should be optional as some filenames do not contain architecture like Java based packages for example.**
+
+
+
+### Artitecture support for each component/role
+
+As per current requirements not every Epiphany component is required to support ARM and there might be cases that a component/role can't support ARM as indicated by initial research [here.](./centos-arm-analysis.md).
```
to-bar:
```suggestion
As per current requirements not every Epiphany component is required to support ARM and there might be cases that a component/role can't support ARM as indicated by initial research [here](./centos-arm-analysis.md).

```

to-bar (approved):

erzetpe (approved):

erzetpe (changes requested):
Please do research on possibility of downloading packages for other architecture than used on download machine - e.g. https://askubuntu.com/questions/430705/how-to-use-apt-get-to-download-multi-arch-library

erzetpe (approved):

Issue #1927, opened 2020-12-22
[Backport to v0.7] Upgrade Zookeeper to 3.5.8


rpudlowski93:
Backport #1853 to v0.7 branch.

It should be synchronised, done and tested together with #1921

rpudlowski93:
Done in PR : https://github.com/epiphany-platform/epiphany/pull/2083

przemyslavic:
Tested:
✔ `epicli apply` with 1 Kafka (+ZooKeeper) node
✔ `epicli apply` with 2 Kafka (+ZooKeeper) nodes
✔ `epicli upgrade` with 1 Kafka node: upgraded ZooKeeper from v3.4.12 (epicli v0.6) to v3.5.8 (epicli v0.7)
✔ `epicli upgrade` with 2 Kafka nodes: upgraded ZooKeeper from v3.4.12 (epicli v0.6) to v3.5.8 (epicli v0.7)

PR #2083, opened 2021-02-26
[BACKPORT] Upgrade Zookeeper to v3.5.8

rpudlowski93:
Related to task: #1927


atsikham (changes requested):
Is the file core/src/epicli/data/common/ansible/playbooks/roles/zookeeper/files/snapshot.0 necessary here?

erzetpe (approved):

atsikham (approved):

rpudlowski93:
> Is the file core/src/epicli/data/common/ansible/playbooks/roles/zookeeper/files/snapshot.0 necessary here?

Yes, it is! This is the same solution which I used in 0.8 and 0.9 version of Epiphany. It is necessary to run zookeeper with any snapshot file and from time to time there can be situation when snapshot file is not present. There is no better solution, there was already some discussions about it. :) ... but good question :)

Issue #1934, opened 2021-01-05
[Backport to v0.7] Upgrade rabbitmq to v3.8.9


atsikham:
Following changes should be backported and tested:
* https://github.com/epiphany-platform/epiphany/pull/1889
* https://github.com/epiphany-platform/epiphany/pull/1903
* https://github.com/epiphany-platform/epiphany/pull/1924
* https://github.com/epiphany-platform/epiphany/pull/1932
* https://github.com/epiphany-platform/epiphany/pull/1938
* https://github.com/epiphany-platform/epiphany/pull/1957
* https://github.com/epiphany-platform/epiphany/pull/1977

przemyslavic:
✅ `epicli apply` v0.7 (1 node)
✅ `epicli` re-`apply` v0.7 (2 clustered nodes)
✅ `epicli upgrade` from v0.7 to dev (2 clustered nodes)

PR #2085, opened 2021-02-26
[Backport] Rabbitmq v3.8.9

atsikham:
Task #1934


erzetpe (dismissed):

CHANGELOG-0.7.md
```diff
@@ -5,6 +5,7 @@
 ### Added
 
 - [#1302](https://github.com/epiphany-platform/epiphany/issues/1302) - Ability to update control plane certificates expiration date
+- [#1934](https://github.com/epiphany-platform/epiphany/issues/1934) - Backported Rabbitmq upgrade to v3.8.9
```
to-bar: I would move this to the `Updated` section since this is not something new.

to-bar (changes requested):
Please update version in COMPONENTS.md.

CHANGELOG-0.7.md
```diff
@@ -10,6 +10,7 @@
 
 - [#1965](https://github.com/epiphany-platform/epiphany/issues/1965) - Upgrade Elasticsearch Curator to v5.8.3
 - [#1927](https://github.com/epiphany-platform/epiphany/issues/1927) - Upgrade Zookeeper to v3.5.8
+- [#1934](https://github.com/epiphany-platform/epiphany/issues/1934) - Backported Rabbitmq upgrade to v3.8.9
```
rpudlowski93: I would write just "Upgrade Rabbitmq to v3.8.9" as it is for others (there were also backported)

erzetpe (dismissed):

core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+cookie_file: /var/lib/rabbitmq/.erlang.cookie
+# check RabbitMQ/Erlang version compatibility matrix before modification
+versions:
+  general: 3.8.9 # required for upgrade
```
rpudlowski93, line 5: Great! :) It's always good to have versions in defaults

core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
...
+  # packages
+  debian:
+    erlang: 1:23.1.*
+    rabbitmq: 3.8.9*
+  redhat:
+    erlang: 23.1.*
+    rabbitmq: 3.8.9
```
rpudlowski93, line 12: Do we need the rabbitmq version specified also for debian and redhat section? Maybe let's use somehow the general version variable? I think that we won't have different version of rabbitmq for redhat and debian.

core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/set-erlang-cookie.yml
```diff
@@ -1,8 +1,43 @@
...
+          set_fact:
+            erlang_cookie: "{{ cookie_file_data.content | b64decode | trim }}"
+          delegate_to: "{{ item }}"
+          delegate_facts: true
+          loop: "{{ ansible_play_hosts }}"
+
+    - name: Generate cookie if it's not defined by user and cookie file doesn't exist
+      when: not cookie_file_stat.stat.exists
+      set_fact:
+        erlang_cookie: "{{ lookup('password', '/dev/null length=15 chars=ascii_letters') }}"
```
rpudlowski93, line 42: Does it lookup any sensitive data? If yes, I would add "no_log: true"

core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/configure-rabbitmq.yml
```diff
@@ -1,52 +1,49 @@
...
+- name: Copy rabbitmq.conf file
   template:
     src: rabbitmq-server.j2
     dest: /etc/rabbitmq/rabbitmq.conf
     owner: root
     group: root
-    mode: 0644
-  register: config_changed
+    mode: u=rw,go=r
+  register: config_file_stat
```
rpudlowski93, line 9: Do I understand it correctly that if we want to have SSL/TLS we need to specify all necessary configuration for it in custom_configurations? If yes, it is possible to set "ssl_options.password" there, so we need to be careful here about sensitive data and use "no_log: true"? WDYT

core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/configure-rabbitmq.yml
```diff
@@ -1,52 +1,49 @@
...
+- name: Restart service
   when:
-    - config_changed.changed or env_settings_changed.changed
-    - not is_ssl_enabled
+    - config_file_stat.changed or env_file_stat.changed or rabbitmq_plugins.changed
+    - not (specification.stop_service | bool)
+  service:
+    name: rabbitmq-server
+    enabled: true
+    state: restarted
```
rpudlowski93, line 42: WDYT about creating handlers for starting/stopping/restarting rabbitmq-server and use just "notify" as it is for example in node exporter role : .../roles/node_exporter/handlers/main.yml ?

core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/update-erlang-cookie.yml
```diff
@@ -0,0 +1,15 @@
...
+    owner: "{{ specification.rabbitmq_user }}"
+    group: "{{ specification.rabbitmq_group }}"
+    mode: u=r,go=
+  register: cookie
+
+- name: Restart service
+  when: cookie.changed
+  service:
+    name: rabbitmq-server
+    state: restarted
```
rpudlowski93, line 15: The same as above - I would use handler and notify here.

core/src/epicli/data/common/defaults/configuration/rabbitmq.yml
```diff
@@ -2,16 +2,16 @@ kind: configuration/rabbitmq
 title: "RabbitMQ"
 name: default
 specification:
-  version: 3.8.3
```
rpudlowski93, line 5: ❤️

docs/home/howto/SECURITY.md
```diff
@@ -182,19 +182,30 @@ Automatic encryption of storage on Azure is not yet supported by Epiphany. Guide
...
+**NOTE**
+
+To complete installation it's required to execute `epicli apply` the second time
+with `stop_service` set to `false`
+
+---
+
+There is `custom_configurations` setting in Epiphany that extends RabbitMQ configuration
+with the custom one. Also, it can be used to perform TLS configuration of RabbitMQ.
+To customize RabbitMQ configuration you need to pass a list of parameters in format:
```
rpudlowski93:
```suggestion
To customize RabbitMQ configuration you need to pass a list of parameters in the following format:
```

core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/defaults/main.yml
atsikham: This is a version of the package and  they have a different format depending on Linux distribution, so I would leave it as is. It's done the same for other roles - https://github.com/epiphany-platform/epiphany/blob/develop/core/src/epicli/data/common/ansible/playbooks/roles/docker/defaults/main.yml#L14

core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/configure-rabbitmq.yml
atsikham: This action have to be executed in place, it's done so here not to create complicated configuration. See https://docs.ansible.com/ansible/latest/user_guide/playbooks_handlers.html#controlling-when-handlers-run.
We can discuss using handlers and `meta: flush_handlers`, but as this is a backporting task, I see no reason to do it in this PR. Probably that's also related to the task with fixing all warnings provided by Ansible linter: defining a code style, which warnings we accept, which are ignored and so on.

---
I would leave it as is as flushing handlers requires control on that part when handlers are changed.

core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/update-erlang-cookie.yml
atsikham: The same as above.

core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/set-erlang-cookie.yml
atsikham: In my opinion this shouldn't be hidden somehow from local log files, @erzetpe ant @to-bar what do you think? 

```
RabbitMQ nodes and the CLI tools use an erlang cookie to determine whether or not they have permission to communicate. The cookie is a string of alphanumeric characters and can be as short or as long as you would like.
```

core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/configure-rabbitmq.yml
atsikham: I would add it here and probably even for erlang cookie, but in this case we also need to do so at least for `develop` branch. @to-bar what do you think?

core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/configure-rabbitmq.yml
rpudlowski93: Yes, lets leave it as it is and I will return to the topic in another task.

core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/defaults/main.yml
rpudlowski93: ok, it is in defaults so let's leave it as it is.

core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/set-erlang-cookie.yml
to-bar: > RabbitMQ nodes and CLI tools (e.g. rabbitmqctl) use a cookie to determine whether they are allowed to communicate with each other. For two nodes to be able to communicate they must have the same shared secret called the Erlang cookie. The cookie is just a string of alphanumeric characters up to 255 characters in size. It is usually stored in a local file. The file must be only accessible to the owner (e.g. have UNIX permissions of 600 or similar). Every cluster node must have the same cookie.

Since Erlang cookie is called "shared secret" and file with it must be only accessible to the owner, I would treat it as sensitive data and use `no_log` parameter here.

core/src/epicli/data/common/ansible/playbooks/roles/rabbitmq/tasks/configure-rabbitmq.yml
to-bar: I've tested `ansible-playbook` with full `verbose` and there was no destination file contents in the output. So I would not add `no_log` here.

to-bar (approved):

erzetpe (approved):

PR #2088, opened 2021-03-01
Replace Skopeo with Crane

seriva:
Replace Skopeo with Crane. Note that Crane does not support older V1 repos, only images in our current lineup for which this was an issues are the following as quay.io was notoriously slow in adopting V2:

- quay.io/coreos/flannel:v0.11.0-amd64 
- quay.io/coreos/flannel:v0.11.0

These images where pulled, re-taged and pushed here:

- epiphanyplatform/flannel:v0.11.0-amd64
- epiphanyplatform/flannel:v0.11.0

```epiphany/core/src/epicli/clusters/build/azureu/ansible/roles/image_registry/tasks/load-image.yml``` will re-tag these automatically when loading epiphanyplatform/flannel:v0.11.0-xxxx -> quay.io/coreos/flannel:v0.11.0-xxxx

### CODE REVIEW

File: docs/home/COMPONENTS.md [hunk 1]
```diff
@@ -198,7 +197,6 @@ Note that versions are default versions and can be changed in certain cases thro
 | s3transfer | 0.3.4 | https://github.com/boto/s3transfer | [Apache License 2.0](https://api.github.com/repos/boto/s3transfer/license) |
 | scp | 0.13.3 | https://github.com/jbardin/scp.py | [Other](https://api.github.com/repos/jbardin/scp.py/license) |
 | six | 1.15.0 | https://github.com/benjaminp/six | [MIT License](https://api.github.com/repos/benjaminp/six/license) |
-| skopeo-bin | 1.0.3 | https://github.com/epiphany-platform/skopeo-bin | [Apache License 2.0](https://api.github.com/repos/epiphany-platform/skopeo-bin/license) |
 | sshtunnel | 0.1.5 | https://github.com/pahaz/sshtunnel | [MIT License](https://api.github.com/repos/pahaz/sshtunnel/license) |
 | tabulate | 0.8.7 | https://github.com/astanin/python-tabulate | [MIT License](https://api.github.com/repos/astanin/python-tabulate/license) |
 | terraform-bin | 1.0.1 | https://github.com/epiphany-platform/terraform-bin | [Apache License 2.0](https://api.github.com/repos/epiphany-platform/terraform-bin/license) |
```
atsikham, line 202: Why Crane is not listed here and not in licenses?

core/src/epicli/cli/epicli.py
```diff
@@ -460,7 +460,6 @@ def dump_external_debug_info(title, args):
         dump_external_debug_info('ANSIBLE CONFIG', ['ansible-config', 'dump'])
         dump_external_debug_info('ANSIBLE-VAULT VERSION', ['ansible-vault', '--version'])
         dump_external_debug_info('TERRAFORM VERSION', ['terraform', '--version'])
-        dump_external_debug_info('SKOPEO VERSION', ['skopeo', '--version'])
```
atsikham, line 463: Should Crane version be added here instead?

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/crane/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+[`k8schain`](pkg/authn/k8schain/README.md) implements the authentication
+semantics use by kubelets in a way that is easily consumable by this library.
+
+`k8schain` is not a standalone tool, but it's linked here for visibility.
+
+### Emeritus: [`ko`](https://github.com/google/ko)
+
+This tool was originally developed in this repo but has since been moved to its
+own repo.
```
atsikham, line 147: Do we need to store this file and the license in repository?

atsikham (changes requested):
General question: what do you think abou not storing binary file in the repository, but download necessary version instead?

docs/home/COMPONENTS.md
seriva: I did not see the need to wrap Crane as a custom Python package is we did with Skopeo:

https://github.com/epiphany-platform/skopeo-bin

The binary is not listed under here: 

https://github.com/epiphany-platform/epiphany/blob/ac8524bc5b6435721371cf5bb19857fe4df008dd/docs/home/COMPONENTS.md#epicli-binary-dependencies

core/src/epicli/cli/epicli.py
seriva: Nope, as I decided not to wrap the the Crane binary inside a custom python package like we did for Skopeo:

https://github.com/epiphany-platform/skopeo-bin

Since Crane never run inside the container itsself but only copied to the repo machine or being returned by ```epicli prepare ...``` in the requirements package.

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/crane/README.md
seriva: Just good practice to state  the readme and resource where something came from.

seriva:
> General question: what do you think abou not storing binary file in the repository, but download necessary version instead?

Before we moved Epicli in a container we had the idea that maybe we could distribute Epicli as a PyPi python package. Hence we have these 2 PyPi packages to install Terraform and Skopeo:

- https://github.com/epiphany-platform/skopeo-bin
- https://github.com/epiphany-platform/terraform-bin

However since this scenario will not happen for epicli, instead of making a similar complicated to maintain PyPi package for Crane I opted to just include the binary. This will also be easier when we need to include the ARM version which is one of the reasons for this switch as well. 

Another option I thought about was to include it on the docker image level and include it, but I prefer the approach of just including it in the repository role as that is where its used for.

erzetpe (approved):

atsikham (approved):

PR #2091, opened 2021-03-02
Changed Ignite download location.

seriva:


przemyslavic:
@seriva The problem with this link may be that it will likely stop working after Ignite 3.0.0 is released.

seriva:
> @seriva The problem with this link may be that it will likely stop working after Ignite 3.0.0 is released.

@przemyslavic From what I see with all official mirrors they only host the last 2 versions of most apache projects:

https://downloads.apache.org/ignite/
https://ftp.ps.pl/pub/apache/ignite/
https://downloads.apache.org//ignite/

Only the archive hosts all of them:

https://ftp.ps.pl/pub/apache/ignite/

So if this is for LTS we might want to stick with what we already have? Yes its slow but its guaranteed to work.

przemyslavic:
@seriva Indeed, it's better to have a stable link. The same problem applies to Kafka and ZooKeeper packages, both of which are downloaded from the apache archive, but luckily these files are much smaller. Hopefully the download speed will increase, and if not, I will make some ad-hoc changes to the pipelines, but this is only a temporary solution, because the links will expire in some time.
@to-bar and I were also thinking about a solution that would check if the "fast" link is available and if not, then download the file from the archive.

seriva:
@przemyslavic For now I am closing this PR and moving the story back to grooming as we need to think of a simple way to add multiple URLs maybe for speed and redundancy. Maybe something as simple as a , separated list would suffice.

przemyslavic:
@serive agree. Everything seems to be back to normal now :)
image

PR #2106, opened 2021-03-08
Prometheus refactoring

rafzei:
Finally, some cleanup..


core/src/epicli/data/common/ansible/playbooks/roles/prometheus/tasks/preflight.yml
```diff
@@ -1,51 +0,0 @@
----
```
mkyc, line 1: Sorry for stupid question, but ... can you explain why all of this gets removed?

plirglo (approved):

seriva (approved):

core/src/epicli/data/common/ansible/playbooks/roles/prometheus/tasks/preflight.yml
rafzei: Because it was not used :cry: . Looks like somebody copied it from another Ansible role with its values and forgot to use/remove.

Issue #2023, opened 2021-02-05
[FEATURE REQUEST] Design doc about VMs naming convention


mkyc:
**Is your feature request related to a problem? Please describe.**
We have different naming conventions in different places. 

**Describe the solution you'd like**
Create naming convention document to stick to it in all the modules creating machines and also in tests. 

**Describe alternatives you've considered**
Nope.

**Additional context**
New modules use different naming convention than epicli.
Example:
was: dec-dev12-logging-nic-0
is: dec-dev12-logging-0-nic
See: https://docs.microsoft.com/en-us/azure/cloud-adoption-framework/ready/azure-best-practices/resource-naming

mkyc:
After design doc is created it should probably be added to release checklists.

PR #2122, opened 2021-03-11
Added naming convention document

atsikham:
Task #2023

### DIFF
--- a/docs/design-docs/infrastructure/naming-convention.md
+++ b/docs/design-docs/infrastructure/naming-convention.md
@@ -27,6 +27,10 @@ with a dash `-` as a delimiter.
 | Application or service name | yes | Name of the application, workload, or service that the resource is a part of | kafka, ignite, opendistro |
 | Environment | no | The stage of the development lifecycle for the workload that the resource supports | prod, dev, qa |
 
+## VM group
+
+The name of VM group that resource is created for. Required only for the basic infrastructure modules.
+
 ## Resource type
 
 Resource type is a short name of resource that is going to be created. Examples:
--- a/docs/design-docs/infrastructure/naming-convention.md
+++ b/docs/design-docs/infrastructure/naming-convention.md
@@ -11,10 +11,6 @@ In general resource name should match following schema:
 
 `<prefix>-<resource_type>-<index>`
 
-or
-
-`<prefix>-<vm_group>-<resource_type>-<index>` when resource name is related to VM group.
-
 ## Prefix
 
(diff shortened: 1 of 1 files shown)


mkyc (commented):
where would vm_groups part go?

atsikham:
> where would vm_groups part go?

That's added after the prefix for both BI modules, added an information about that.

docs/design-docs/infrastructure/naming-convention.md
```diff
@@ -0,0 +1,40 @@
...
+## Prefix
+
+Epiphany modules are developed in the way that allows user specify a prefix for created resources. This approach gives
+such benefits as ordered sorting and identifying who is the owner of the resource. Prefix can include following parts
+with a dash `-` as a delimiter.
+
+| Type | Required | Description | Examples |
+| ---- | -------- | ----------- | -------- |
+| Owner | yes | The name of the person or team which resource belongs to | epiphany |
+| Application or service name | yes | Name of the application, workload, or service that the resource is a part of | kafka, ignite, opendistro |
```
mkyc: One more question regarding required=yes here. I'm not sure, but that doesn't look like required to me. Is resource group part of some application for example? I believe owner is one required field for sure. 

At second look at this document it also looks that vm_group is just one of elements of prefix.

atsikham: Please check now

mkyc (approved):
I think it's good policy now. It's clear, short, general and extendable.

seriva (approved):
For Azure this is fine, but how about AWS?  IMO more problematic as we dont have anything like a proper resource type of thing to group everything under. Note that this can be in a seperate issue for me, hense my approval.

edit: also note that not all resources can follow this conventions like storage account name: https://github.com/epiphany-platform/epiphany/blob/90b55f8d9dc44586217fc5035b855e9b89bfa02e/core/src/epicli/cli/helpers/naming_helpers.py#L34

atsikham:
> For Azure this is fine, but how about AWS? IMO more problematic as we dont have anything like a proper resource type of thing to group everything under. Note that this can be in a seperate issue for me, hense my approval.
> 
> edit: also note that not all resources can follow this conventions like storage account name:

As I remember this issue originally came from modules, when we started naming in different ways, without care about delimiters, index and resource type positions. I checked AWS BI module and don't see resources there that can't follow these rules.

In my opinion we can try following these recommendations and update the document if something wasn't taken into account.

PR #2126, opened 2021-03-11
Retry function for download-requirements.sh

ar3ndt:


core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/ubuntu-18.04/download-requirements.sh
```diff
@@ -125,7 +126,11 @@ echol "Packages to be downloaded:"
 cat -n "${deplist}"
 
 # download dependencies (apt-get sandboxing warning when running as root are harmless)
-cd $dst_dir_packages && xargs --no-run-if-empty --arg-file=${deplist} --delimiter='\n' ${download_cmd} | tee -a ${logfile}
+export -f run_cmd_with_retries
+export -f run_cmd
+export -f print_array_as_shell_escaped_string
+export -f get_shell_escaped_array
```
to-bar: Do we need these exports?
These functions are sourced in line 35:
```
# source common functions
. "${script_path}/common.sh"
```

ar3ndt: yup, this is because of using xargs so it works in separate context

to-bar (changes requested):
Please update changelog

ar3ndt:
> Please update changelog

done

mkyc (dismissed):

to-bar:
Since in changelog we have plural form `retry feature for failed commands` I would recommend extending the functionality to `wget` and optionally `crane` or editing the changelog entry to be more precise - that this is only for downloading packages.

to-bar (changes requested):
Since in changelog we have plural form `retry feature for failed commands` I would recommend extending the functionality to `wget` and optionally `crane` or editing the changelog entry to be more precise - that this is only for downloading packages.

plirglo (dismissed):

seriva:
> Since in changelog we have plural form `retry feature for failed commands` I would recommend extending the functionality to `wget` and optionally `crane` or editing the changelog entry to be more precise - that this is only for downloading packages.

Lets leave as is for now and not introduce more changes before release. Lets only change the changelog. We can fully add the retry when we re-write the download scripts and share common code between them.

to-bar (approved):

mkyc (approved):

PR #2138, opened 2021-03-16
offline mode in modules

mkyc:


mkyc (commented):
See it [here](https://github.com/epiphany-platform/epiphany/blob/feature/1481-offline-in-modules/docs/design-docs/modularization/offline.md).

atsikham (approved):
Reviewed. If I understand everything correctly, as repository is a machine with Apache web server installed, import/export steps can be done just by downloading dependencies from "downloader" to the correct place on "repository" machine without any archival. LGTM

seriva (approved):

PR #2150, opened 2021-03-23
Check if user has root privileges through sudo

to-bar:
- Refactor OS validation
- Disable become in preflight playbook
- Check if user has root privileges through sudo
- Update docs


core/src/epicli/data/common/defaults/configuration/shared-config.yml
```diff
@@ -7,33 +7,5 @@ specification:
   download_directory: /tmp # directory where files and images will be stored just before installing/loading
   vault_location: '' # if empty "BUILD DIRECTORY/vault" will be used
   vault_tmp_file_location: SET_BY_AUTOMATION
```
seriva, line 9: We dont want to have supported OSs configurable anymore?

to-bar: Yes, I've changed this intentionally.

mkyc (approved):
LGTM

seriva (approved):

PR #2156, opened 2021-03-26
block arm unsupported components

ar3ndt:


core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/main.yml
```diff
@@ -1,6 +1,32 @@
 ---
 - include_tasks: check-user.yml
 
+- name: Set is architecture supported to true
+  set_fact:
+     is_arch_supported: true
+  changed_when: true
+
+- name: Set is architecture supported to false if needed roles are not present in supported list of components
```
atsikham:
```suggestion
- name: Set is_arch_supported fact to false if needed roles are not present in supported list of components
```

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/main.yml
```diff
@@ -1,6 +1,32 @@
 ---
 - include_tasks: check-user.yml
 
+- name: Set is architecture supported to true
```
atsikham:
```suggestion
- name: Set is_arch_supported fact to true
```

core/src/epicli/data/common/defaults/configuration/shared-config.yml
```diff
@@ -7,5 +7,8 @@ specification:
   download_directory: /tmp # directory where files and images will be stored just before installing/loading
   vault_location: '' # if empty "BUILD DIRECTORY/vault" will be used
   vault_tmp_file_location: SET_BY_AUTOMATION
+  supported_architecture:
```
atsikham: What's about x86_64?

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/main.yml
```diff
@@ -1,6 +1,32 @@
...
+
+- name: Set is architecture supported to false if needed roles are not present in supported list of components
+  set_fact:
+     is_arch_supported: false
+  with_nested:
+    - "{{ supported_architecture }}"
+    - "{{ group_names }}"
+  when:
+    - item[1] not in item[0].components
+    - item[0].name == ansible_architecture
```
atsikham: It's possible to use intersection or symmetric_difference filter here without a loop
https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#selecting-from-sets-or-lists-set-theory

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/main.yml
```diff
@@ -1,6 +1,32 @@
...
+
+- name: Check architecture
+  assert:
+    that: is_arch_supported
+    fail_msg: "Your architecture {{ ansible_architecture  }} is not supported by all components in inventory."
+    success_msg: "Your architecture {{ ansible_architecture  }} is supported by all components in inventory."
```
atsikham: It's possible to use difference filter here to print unsupported components

atsikham (changes requested):

core/src/epicli/data/common/defaults/configuration/shared-config.yml
to-bar: I don't think this is a right place for this property for two reasons:
1) supported components are not configurable, they are defined by us (dev team) not by users
2) this property is used only by preflight role so it's better to be local
My proposal is to move it to: `data/common/ansible/playbooks/roles/preflight/defaults/main.yml`

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/main.yml
```diff
@@ -1,6 +1,32 @@
...
+
+- name: Set is OS supported to false
```
to-bar: It looks like unused fact.

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/main.yml
```diff
@@ -1,6 +1,32 @@
 ---
 - include_tasks: check-user.yml
 
+- name: Set is architecture supported to true
+  set_fact:
+     is_arch_supported: true
+  changed_when: true
```
to-bar: This parameter has no sense for `set_fact` task since it doesn't change state.

to-bar (changes requested):

core/src/epicli/data/common/defaults/configuration/shared-config.yml
ar3ndt: not needed for now.

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/main.yml
ar3ndt: used just difference, loop is needed but at least not nested

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/main.yml
ar3ndt: it actually changes when found components not listed

to-bar: I mean `set_fact` task operates in memory so it's doesn't change the state.

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/main.yml
```diff
@@ -1,6 +1,25 @@
 ---
 - include_tasks: check-user.yml
 
+- name: Set is_arch_supported fact to true
+  set_fact:
+     is_arch_supported: true
+  changed_when: true
+
+- name: Set is is_arch_supported fact to false if needed roles are not present in supported list of components
```
to-bar:
```suggestion
- name: Set is_arch_supported fact to false if needed roles are not present in list of supported components
```

core/src/epicli/data/common/ansible/playbooks/roles/preflight/defaults/main.yml
```diff
@@ -12,3 +12,5 @@ supported_oses:
     version:
       min: 18.04
       max: 18.04
+supported_architecture:
```
to-bar: Maybe something like this (just example):

```
unsupported_roles:
  - distro: Ubuntu
    arch: x86_64
    roles: []  # all supported
  - distro: Ubuntu
    arch: aarch64
    roles: "{{ group_names }}"  # all unsupported
  - distro: CentOS
    arch: aarch64
    roles:
      - elasticsearch_curator
```

So by default everything is supported unless present here. Note that unfortunately features have different naming than roles, for example:
feature: `elasticsearch-curator`,  role: `elasticsearch_curator`

core/src/epicli/data/common/defaults/configuration/shared-config.yml
ar3ndt: it is done as assumed in the task:
aarch64: [] -> this empty list means ARM is supported by none of components  so it will fail everytime we apply epi on the machine with ARM arch. Once we support minimal list of components - it should be added here as starting point.

missing x86_64 -> not figuring arch means this arch is supported by all components so no point to list them.

core/src/epicli/data/common/ansible/playbooks/roles/preflight/defaults/main.yml
seriva: Yeah something like this makes sense as we need to check it on an architecture and OS level.

ar3ndt: ok, ill refactor it then

ar3ndt:
changes adjusted

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/main.yml
```diff
@@ -1,6 +1,25 @@
...
+  set_fact:
+     is_arch_supported: false
+  when: unsupported_components_checklist | length > 0
+
+- name: Check architecture
+  assert:
+    that: is_arch_supported
+    fail_msg: "Your architecture {{ ansible_architecture }} is not supported at {{ ansible_distribution }} by following components in inventory -> {{ unsupported_components_checklist }}."
+    success_msg: "Your architecture {{ ansible_architecture }} is supported at {{ ansible_distribution }} by all components in inventory."
+
```
atsikham:
```suggestion
- name: Set unsupported_components_checklist fact
  set_fact:
     unsupported_components_checklist: "{{ unsupported_roles | selectattr('distro','equalto', ansible_distribution) | selectattr('arch','equalto', ansible_architecture) | map(attribute='roles') | list | flatten | intersect(group_names) }}"

- name: Check architecture
  assert:
    that: unsupported_components_checklist | length > 0
    fail_msg: "Your architecture {{ ansible_architecture }} is not supported at {{ ansible_distribution }} by following components in inventory -> {{ unsupported_components_checklist }}."
    success_msg: "Your architecture {{ ansible_architecture }} is supported at {{ ansible_distribution }} by all components in inventory."

```

atsikham (changes requested):

atsikham (approved):

CHANGELOG-0.10.md
```diff
@@ -9,6 +9,7 @@
 - [#2081](https://github.com/epiphany-platform/epiphany/issues/2081) - Replace Skopeo with Crane
 - [#1323](https://github.com/epiphany-platform/epiphany/issues/1323) - Documentation how to handle data in Opendistro for Elasticsearch
 - [#1789](https://github.com/epiphany-platform/epiphany/issues/1789) - [Ubuntu] Add retry feature for downloading packages in download-requirements.sh
+- [#2118](https://github.com/epiphany-platform/epiphany/issues/2118) - Block the install of unsupported components/roles on arm64
```
seriva: This is not for 0.10.x, pls add this to the CHANGELOG-0.11.md

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/main.yml
```diff
@@ -1,6 +1,16 @@
 ---
 - include_tasks: check-user.yml
 
+- name: Set unsupported_components_checklist fact
+  set_fact:
+     unsupported_components_checklist: "{{ unsupported_roles | selectattr('distro','equalto', ansible_distribution) | selectattr('arch','equalto', ansible_architecture) | map(attribute='roles') | list | flatten | intersect(group_names) }}"
+
+- name: Check architecture
```
seriva: - name: Check if components are supported by the architecture

core/src/epicli/data/common/ansible/playbooks/roles/preflight/defaults/main.yml
```diff
@@ -1,14 +1,118 @@
...
+      - haproxy_exporter
+      - vault
+      - applications
+      - ignite
+  - distro: CentOS
+    arch: x86_64
+    roles: []  # all supported
+  - distro: CentOS
+    arch: aarch64
+    roles:
```
seriva, line 57: Please enable the components we already support inside the ARM branch till now and make sure they install correctly:

      - repository
      - kafka
      - image_registry
      - kubernetes_master
      - kubernetes_node
      - helm
      - zookeeper
      - haproxy
      - vault
      - firewall

core/src/epicli/data/common/ansible/playbooks/roles/preflight/defaults/main.yml
```diff
@@ -1,14 +1,88 @@
...
+      min: 18.04
+      max: 18.04
+
+unsupported_roles:
+  - distro: Ubuntu
+    arch: x86_64
+    roles: []  # all supported
+  - distro: Ubuntu
+    arch: aarch64
+    roles:
```
seriva, line 22: On Ubuntu or Redhat we should not support any roles at this point

core/src/epicli/data/common/ansible/playbooks/roles/preflight/defaults/main.yml
seriva: Also for centos pls just add all roles are going to support as postgress and applications are already there. So missing would be everything involving logging, monitoring and rabbitmq.

core/src/epicli/data/common/ansible/playbooks/roles/preflight/defaults/main.yml
ar3ndt: ok

seriva (approved):

PR #2162, opened 2021-03-26
Remove epirepo before upgrade to free up space

to-bar:


seriva (approved):

przemyslavic:
The fix works in the sense that it removes unnecessary files, packages and images, but for some configurations it is still insufficient. With these changes I was able to successfully update 3/10 clusters from version 0.6 to dev and 8/10 clusters from version 0.7. In other cases, there was no disk space anyway. The problem applies to old versions of clusters (before the introduction of the repository vm in version 0.8), where Azure disks were 32 GB and AWS 30 GB.

przemyslavic (approved):

mkyc (approved):
LGTM

PR #2173, opened 2021-03-31
Fix for connection failure due to unknown certificate when upgrading ODFE

to-bar:


to-bar:
Closing since we merged alternative fix.

PR #2196, opened 2021-04-09
v0.6 Remove new line from version.txt.py

przemyslavic:
Removed new line due to image building issues (reverted to what it was before).
```
WARNING: Requirement 'epicli-0.6.2\r-py3-none-any.whl' looks like a filename, but the file does not exist
Processing ./epicli-0.6.2
-py3-none-any.whl
ERROR: Could not install packages due to an OSError: [Errno 2] No such file or directory: '/epicli/epicli-0.6.2-py3-none-any.whl'
```


rafzei (approved):
:rocket:

atsikham (approved):

Issue #2113, opened 2021-03-10
[FEATURE REQUEST] Add arm64 support to 'applications' Ansible role to install 'Keycloak'


seriva:
**Is your feature request related to a problem? Please describe.**
Currently we only support the x86_64 architectures but we need to add support for arm64.

**Describe the solution you'd like**
- We want the ability to install ```keycloak``` on K8s running on arm64 machines running CentOS.
- Use the following multi-arch image: https://hub.docker.com/repository/docker/epiphanyplatform/keycloak
- Only the latest versions we support: 9
- Any requirements missing or not added in #2033 will be added for the effected roles.
- Any other deployments available inside the application role should be blocked from installing on arm64.

Effected Ansible roles:
- applications

**Describe alternatives you've considered**
None.

**Additional context**
No.

**Extra info**
Design document and resources:
https://github.com/epiphany-platform/epiphany/tree/develop/docs/design-docs/arm

Terraform scripts for ARM hardware on AWS:
https://github.com/mkyc/aws-arm-simple-setup

YAML for testing:
```YAML
TODO
```

---

**DoD checklist**

* [x] Changelog updated
* [x] COMPONENTS.md updated / doesn't need to be updated
* [x] Feature has automated tests
* [x] Automated tests passed (QA pipelines)
  * [x] apply
  * [ ] upgrade
* [ ] Idempotency tested
* [x] Documentation added / updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Solution meets requirements and is done according to design doc
* [x] Usage compliant with license

rpudlowski93:
✅ Tested, works fine, I was able to execute tests from our code in https://github.com/epiphany-platform/epiphany/tree/ARM/core/src/epicli/data/common/tests/spec

atsikham:
In addition need to verify that it's possible to install only supported apps. It's validated in preflight checks.

przemyslavic:
Assertion was added in preflight check to fail if there are unsupported applications enabled.
:heavy_check_mark: Tested on aarch64 instances.
```
2021-04-26T14:20:59.1105308Z[38;21m14:20:59 INFO cli.engine.ansible.AnsibleCommand - TASK [preflight : Check if enabled applications are supported for the architecture] ***
[...]

Issue #2112, opened 2021-03-10
[FEATURE REQUEST] Add arm64 support to 'postgresql' component


seriva:
**Is your feature request related to a problem? Please describe.**
Currently we only support the x86_64 architectures but we need to add support for arm64.

**Describe the solution you'd like**
- We want the ability to install the ```postgresql``` on arm64 machines running CentOS.
- We only want to support the single master node configuration. Any additional configuration (replication, Pgpool etc) should not be available for arm64 for now.
- Only the latest versions we support: 10
- Any requirements missing or not added in #2033 will be added for the effected roles.
- postgres-exporter, node-exporter, filebeatcan be skipped at this point.

Affected Ansible roles:
- postgresql
- applications
- firewall

Skip the following Ansible roles for now by using a custom ```feature-mapping``` config:
- postgres-exporter
- node-exporter
- filebeat

**Describe alternatives you've considered**
None.

**Additional context**
No.

**Extra info**
Design document and resources:
https://github.com/epiphany-platform/epiphany/tree/develop/docs/design-docs/arm

Terraform scripts for ARM hardware on AWS:
https://github.com/mkyc/aws-arm-simple-setup

YAML for testing:
```YAML
TODO
```

---

**DoD checklist**

* [x] Changelog updated
* [x] COMPONENTS.md updated / doesn't need to be updated
* [x] Feature has automated tests
* [x] Automated tests passed (QA pipelines)
  * [x] apply
  * [ ] upgrade
* [ ] Idempotency tested
* [x] Documentation added / updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Solution meets requirements and is done according to design doc
* [x] Usage compliant with license

DaOsie:
There are arm64 packages available on official postgresql yum repo: https://yum.postgresql.org/rpmchart/

rpudlowski93:
✅ Tested, works fine, db is available and I was able to execute tests from our code in https://github.com/epiphany-platform/epiphany/tree/ARM/core/src/epicli/data/common/tests/spec

atsikham:
In addition need to verify that it's possible to install only supported PostgreSQL extensions (pgaudit). It's validated in preflight checks.

przemyslavic:
Assertion was added in preflight check to fail if there is PgBouncer extension or PostgreSQL replication with RepManager enabled.
:heavy_check_mark: Tested on aarch64 instances.
```
[...]

PR #2202, opened 2021-04-09
ARM support for PostgreSQL and Keycloak

atsikham:
Tasks #2112 #2113


docs/home/howto/DATABASES.md
```diff
@@ -209,6 +217,15 @@ PostgreSQL: resources management | https://www.postgresql.org/docs/10/runtime-co
 
 ### Installing PgBouncer and Pgpool
 
+---
+**NOTE**
+
+PgBouncer and Pgpool Docker images are not supported for ARM.
+If these applications are enabled in configuration, pods will not start on ARM nodes.
```
seriva: , they will not be deployed on ARM.

docs/home/howto/DATABASES.md
```diff
@@ -7,6 +7,14 @@ Collections, but are supported by PostgreSQL installed from official PostgreSQL
 
 ### PostgreSQL servers
 
+---
+**NOTE**
+
+PgBouncer and replication (Repmgr) extensions are not supported on ARM and these components will not be installed
+even if enabled in configuration. Check ARM design [document](../../design-docs/arm/arm.md)
```
seriva: I would suggest to not link directly to this document. We need to specify this in more "offical" document later.

docs/home/howto/DATABASES.md
```diff
@@ -209,6 +217,15 @@ PostgreSQL: resources management | https://www.postgresql.org/docs/10/runtime-co
 
 ### Installing PgBouncer and Pgpool
 
+---
+**NOTE**
+
+PgBouncer and Pgpool Docker images are not supported for ARM.
+If these applications are enabled in configuration, pods will not start on ARM nodes.
+Check ARM design [document](../../design-docs/arm/arm.md)
```
seriva: Same comment as above. Lets not refer to this document in this way.

core/src/epicli/data/common/ansible/playbooks/roles/applications/templates/pgbouncer/03-deployment.yml.j2
```diff
@@ -68,6 +68,19 @@ spec:
               subPath: pgpool_postgres_password
             - name: pgpool-shared-init-dir
               mountPath: /epiphany/shared_dir
+      affinity:
```
seriva: While this will work for the deployment I think it will still create some resources like namespaces etc. For applications in general we need some mechanism that will avoid installation of anything else then the auth-service on ARM:

image

seriva (approved):

core/src/epicli/data/common/ansible/playbooks/roles/recovery/tasks/postgresql.yml
```diff
@@ -132,11 +129,8 @@
         service:
           name: "{{ repmgr_service_name[ansible_os_family] }}"
           state: started
-      when:
-        - component_vars.specification.extensions.replication.enabled is defined
-        - component_vars.specification.extensions.replication.enabled
-        - component_vars.specification.extensions.replication.use_repmgr is defined
-        - component_vars.specification.extensions.replication.use_repmgr
+        when:
```
mkyc: I just want to ensure that this `when` was moved intentionally.

atsikham: Fixed

mkyc (approved):

PR #2216, opened 2021-04-15
Add ARM support for rabbitmq

rpudlowski93:
PR for the task #2114 

- Support ARM for rabbitmq component


atsikham (approved):
Single Erlang source would be better, but as I understand, there are no ARM packages published on GitHub. LGTM.

seriva (approved):

Issue #2118, opened 2021-03-10
[FEATURE REQUEST] Block the install of unsupported components/roles on arm64


seriva:
**Is your feature request related to a problem? Please describe.**
Currently we only support the x86_64 architectures but we need to add support for arm64.

**Describe the solution you'd like**
- We want the ability to block the install of unsupported roles on arm64 machines running CentOS. More information [here](https://github.com/epiphany-platform/epiphany/blob/develop/docs/design-docs/arm/arm.md#artitecture-support-for-each-componentrole) and [here](https://github.com/epiphany-platform/epiphany/blob/develop/docs/design-docs/arm/arm.md#pre-flight-check).

Effected Ansible roles:
- all

**Describe alternatives you've considered**
None.

**Additional context**
No.

**Extra info**
Design document and resources:
https://github.com/epiphany-platform/epiphany/tree/develop/docs/design-docs/arm

Terraform scripts for ARM hardware on AWS:
https://github.com/mkyc/aws-arm-simple-setup

---

**DoD checklist**

* [x] Changelog updated
* [x] COMPONENTS.md updated / doesn't need to be updated
* [ ] Feature has automated tests
* [ ] Automated tests passed (QA pipelines)
  * [ ] apply
  * [ ] upgrade
* [ ] Idempotency tested
* [ ] Documentation added / updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Solution meets requirements and is done according to design doc
* [x] Usage compliant with license

przemyslavic:
@ar3ndt as far as I know, we do not support Ignite installation on arm machines yet.
I think that Ignite role should be removed from [here ](https://github.com/epiphany-platform/epiphany/pull/2156/files#diff-f103929a992efef8f98a45eeccb6483e5f6be8ea67731c9bb30d6d03c1b4db22R75) and from other distros.

The installation fails on task `download : Download file apache-ignite-2.9.1-bin.zip` as Ignite package is not being downloaded at all. It is commented out in the requirements.txt file and therefore not available in epirepo.
```
2021-04-16T11:32:11.4557541Z[38;21m11:32:11 INFO cli.engine.ansible.AnsibleCommand - TASK [download : Download file apache-ignite-2.9.1-bin.zip] ********************
[...]

PR #2224, opened 2021-04-19
Fix and update list of unsupported components for ARM

ar3ndt:


seriva (approved):

mkyc (approved):

przemyslavic:
Still an issue:
https://github.com/epiphany-platform/epiphany/issues/2118#issuecomment-823041751

atsikham (changes requested):
Another thing to think about - preflight role is executed for all hosts. In my opinion we shouldn't do such/similar check for all of them.

ar3ndt:
> Another thing to think about - preflight role is executed for all hosts. In my opinion we shouldn't do such/similar check for all of them.

how should we limit it then?

atsikham:
> > Another thing to think about - preflight role is executed for all hosts. In my opinion we shouldn't do such/similar check for all of them.
> 
> how should we limit it then?

Up to you, but I would execute this check once, delegate it to localhost and use a loop to compare enabled components and architectures (retrieved from facts e.g. `hostvars[groups['kubernetes_master'][0]]['ansible_architecture']`)

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/main.yml
```diff
@@ -2,14 +2,18 @@
 - include_tasks: check-user.yml
 
 - name: Set unsupported_components_checklist fact
+  delegate_to: localhost
+  become: false
   set_fact:
-     unsupported_components_checklist: "{{ unsupported_roles | selectattr('distro','equalto', ansible_distribution) | selectattr('arch','equalto', ansible_architecture) | map(attribute='roles') | list | flatten | intersect(group_names) }}"
+     unsupported_components_checklist: "{{ unsupported_roles | selectattr('distro','equalto', hostvars[groups['repository'][0]]['ansible_distribution']) | selectattr('arch','equalto', hostvars[groups['repository'][0]]['ansible_architecture']) | map(attribute='roles') | list | flatten | intersect(group_names) }}"
```
to-bar: IMO we should check all groups, not only `repository` since we have "custom repository" feature so it's possible to have no `repository` group.

More info here:
https://github.com/epiphany-platform/epiphany/blob/develop/docs/home/howto/CLUSTER.md#how-to-create-an-epiphany-cluster-using-custom-system-repository-and-docker-image-registry

to-bar (changes requested):

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/main.yml
```diff
@@ -2,14 +2,18 @@
 - include_tasks: check-user.yml
 
 - name: Set unsupported_components_checklist fact
+  delegate_to: localhost
```
to-bar: I would write this using block vars (not tested):

```
- name: Check if components are supported for architecture
  vars:
    architecture: "{{ hostvars[groups.all[0]]['ansible_architecture'] }}"
    os_distro:    "{{ hostvars[groups.all[0]]['ansible_distribution'] }}"
  block:
    - name: Set unsupported_components_checklist fact
      delegate_to: localhost
      become: false
      set_fact:
        unsupported_components_checklist: "{{ unsupported_roles | selectattr('distro','equalto', os_distro)
                                                                | selectattr('arch','equalto', architecture)
                                                                | map(attribute='roles') | list | flatten
                                                                | intersect(group_names) }}"

    - name: Check if components are supported for current architecture
      delegate_to: localhost
      become: false
      assert:
        that: unsupported_components_checklist | length == 0
        fail_msg: >-
          The following components in inventory are not supported for {{ architecture }} architecture on {{ os_distro }} OS:
          {{ unsupported_components_checklist | join(', ') }}.
          If the components are not neccessary, you may remove them from the feature-mapping
          (see https://github.com/epiphany-platform/epiphany/blob/develop/docs/home/howto/CLUSTER.md#how-to-create-custom-cluster-components).
        success_msg: All components supported for {{ architecture }} architecture
```

to-bar (approved):

atsikham (approved):

atsikham (changes requested):
#2250 was merged, vault is supported now

atsikham (approved):

PR #2246, opened 2021-04-22
ARM support for Ignite and RabbitMQ applications

atsikham:


seriva (approved):

mkyc:
Shouldn't we also mention in change log and in components that we upgrade Istio version?

atsikham:
> Shouldn't we also mention in change log and in components that we upgrade Istio version?

There is only Ignite changes, version was just changed in docs to be the same as we use.

mkyc:
> > Shouldn't we also mention in change log and in components that we upgrade Istio version?
> 
> There is only Ignite changes, version was just changed in docs to be the same as we use.

Ohhh ... sorry for typo, I meant ignite. Yeah I just checked that versions are correct.

seriva (changes requested):
Does this actually make the change for Redhat?

atsikham:
> Does this actually make the change for Redhat?

Good catch

mkyc (approved):

to-bar (approved):

PR #2238, opened 2021-04-21
[0.8][Backport] opendistro_for_elasticsearch: fix upgrades after lock_timeout refactor

to-bar:
Additional PR for task #2209

PR #1793, opened 2020-10-23
Upgrade filebeat to 7.9.2

sk4zuzu:
I tested this in 7 scenarios:
- on-prem ("any" provider) Ubuntu Epiphany starting from `develop`  (running `epicli apply` and `epicli upgrade`)
- on-prem ("any" provider) Centos Epiphany starting from `develop`  (running `epicli apply` and `epicli upgrade`)
- on-prem ("any" provider) RedHat Epiphany starting from `develop`  (running `epicli apply` and `epicli upgrade`)
- cloud with the `k8s_as_cloud_service` flag enabled (to test helm changes) using https://github.com/epiphany-platform/m-generic-epiphany-on-aks module

```shell
m-generic-epiphany-on-aks/examples/basic_flow$ kubectl --kubeconfig shared/build/azepi/kubeconfig get pods
NAME                      READY   STATUS    RESTARTS   AGE
filebeat-filebeat-dtkj6   1/1     Running   0          27m
filebeat-filebeat-s5c9x   1/1     Running   0          27m
node-exporter-w2z86       1/1     Running   0          25m
node-exporter-zt4t6       1/1     Running   0          25m
```

No issues encountered! :+1::innocent:

PR #2256, opened 2021-04-27
Fix Filebeat version reference

plirglo:


przemyslavic:
Fix for the issue introduced in #2219 (that has been merged but not tested yet).

core/src/epicli/data/common/ansible/playbooks/roles/filebeat/tasks/install-filebeat-as-system-service.yml
```diff
@@ -8,9 +8,9 @@
   vars:
     _packages:
       Debian:
-        - filebeat={{ filebeat_version }}
+        - filebeat={{ specification.filebeat_version }}
```
atsikham, line 11: Why it's taken from specification if we have that specified in the role's defaults?

atsikham (changes requested):

atsikham:
We also have #2238 PR, what do you think about combining them both?

przemyslavic:
We have it in 2 places 😱 
https://github.com/epiphany-platform/epiphany/blob/v0.8/core/src/epicli/data/common/defaults/configuration/filebeat.yml

https://github.com/epiphany-platform/epiphany/blob/v0.8/core/src/epicli/data/common/ansible/playbooks/roles/filebeat/defaults/main.yml

Looks like something has not been properly backported.

core/src/epicli/data/common/ansible/playbooks/roles/filebeat/tasks/install-filebeat-as-system-service.yml
plirglo: Yes, it looks like: https://github.com/epiphany-platform/epiphany/blob/v0.8/core/src/epicli/data/common/ansible/playbooks/roles/filebeat/defaults/main.yml

atsikham:
I see that this refactoring was done in #1793. Imo the right way is to remove this property from `defaults/configuration/filebeat.yml` as it's done in latest versions, not to allow user specify it. Then rename it in role's defaults, `specification` shouldn't be there.

rafzei:
That's right. The role was changed in v0.9.0. Changes wasn't backported so I think we shouldn't merge this PR, instead do a backport.

plirglo:
Maybe we should just fix existing bug to have working configuration (since it's already done in this PR) and create new issue with backporting Filebeat upgrade and role refactoring?

rafzei (approved):
@plirglo  please create a new issue with backporting missing changes to 0.8

atsikham (approved):

plirglo:
Created issue #2260

Issue #2241, opened 2021-04-22
[FEATURE REQUEST] Add arm64 support to 'vault'


rpudlowski93:
**Is your feature request related to a problem? Please describe.**
Currently we only support the x86_64 architectures but we need to add support for arm64.

**Describe the solution you'd like**
- We want the ability to install the ```vault``` on arm64 machines running CentOS.

Affected Ansible roles:
- vault

**Describe alternatives you've considered**
None.

**Additional context**
No.

**Extra info**
Design document and resources:
https://github.com/epiphany-platform/epiphany/tree/develop/docs/design-docs/arm

Terraform scripts for ARM hardware on AWS:
https://github.com/mkyc/aws-arm-simple-setup

YAML for testing:
```YAML
TODO
```

---

**DoD checklist**

* [x] Changelog updated
* [x] COMPONENTS.md updated / doesn't need to be updated
* [ ] Feature has automated tests
* [x] Automated tests passed (QA pipelines)
    * [x] apply
    * [ ] upgrade
* [ ] Idempotency tested
* [x] Documentation added / updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Solution meets requirements and is done according to design doc
* [x] Usage compliant with license

przemyslavic:
The installation was successful. More testing in progress...

przemyslavic:
@rpudlowski93 Installation was successful, but Vault is not working. When trying to deploy a sample application with the appropriate annotations to test Vault, it fails to download the Vault image.
```
Failed to pull image "ec2-1-1-1-1.eu-west-1.compute.amazonaws.com:5000/vault:1.7.0"
```
From what I can see Vault Helm Chart v0.11.0 requires an image in 1.7.0 version and in epirepo we still have `vault:1.6.1` which was ok but for Helm Chart v0.9.0.

Moving back to TODO.

przemyslavic:
Testing PR #2268.
:heavy_check_mark: AWS CentOS [arm]
:heavy_check_mark: AWS/Azure x Ubuntu/RHEL/CentOS [x64]

I deployed the sample app on kubernetes and was able to access the secrets.

PR #2268, opened 2021-04-29
[Fix] Vault image updated to version 1.7.0

rpudlowski93:
Fix for the issue in the task #2241

- Hashicorp Vault image updated to version 1.7.0


atsikham (changes requested):
PR should be modified, check commits and changed files.

atsikham (approved):

seriva (approved):

Issue #2215, opened 2021-04-15
[BUG] Grafana does not show metrics from kafka and haproxy exporters


rpudlowski93:
**Describe the bug**
Grafana does not show metrics from kafka and haproxy exporters. The problem occurs for the standard dashboards such as [Kafka Exporter Overview Dashboard](https://grafana.com/grafana/dashboards/7589). Looks like a problem with Grafana since using Prometheus it is possible to scrape metrics from the mentioned above exporters.

**How to reproduce**
1. execute `epicli init ... `
2. edit config file and add logging, loadbalancer and kafka components
3. execute `epicli apply ...`
4. check grafana dashboard

**Expected behavior**
I would like to check metrics from kafka and haproxy exporters in Grafana dashboards.

**Environment**
- Cloud provider: [AWS | Azure, On-prem]
- OS: all[ARM][AMD]

**epicli version**: 0.10.0

**Additional context**

---

**DoD checklist**

* [x] Changelog updated (if affected version was released)
* [x] COMPONENTS.md updated / doesn't need to be updated
* [x] Automated tests passed (QA pipelines)
  * [x] apply
  * [ ] upgrade
* [x] Case covered by automated test (if possible)
* [ ] Idempotency tested
* [x] Documentation updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Backport tasks created / doesn't need to be backported

rafzei:
Check if Prometheus targets & Grafana's sources are set properly if so test with other dashboards.

rpudlowski93:
Done in PR #2304

przemyslavic:
Imported `Kafka Exporter Overview` and `HAProxy` dashboards.
After generating traffic (e.g. after running tests), metrics for both exporters were visible.

PR #2304, opened 2021-05-07
[BUG] Grafana does not show metrics from kafka and haproxy exporters

rpudlowski93:
PR related to task #2215

- In case of Kafka Exporter everything works fine, it is necessary to generate traffic (create topic, send message and pull message from the topic using consumers group" - added small note in documentation 
- In case of HAProxy it is necessary to generate traffic and I also added csv format to stats uri which exporter use

### DIFF
--- a/docs/home/howto/MONITORING.md
+++ b/docs/home/howto/MONITORING.md
@@ -187,6 +187,8 @@ List of monitoring components - so called exporters:
 
 When dashboard creation or import succeeds you will see it on your dashboard list.
 
+`Note: Depends on the dashboards which rely on variouse variables, in order to visualize the data it is necessary to generate traffic to the components which are monitored`
+
 # Kibana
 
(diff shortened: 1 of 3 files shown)


seriva (dismissed):

docs/home/howto/MONITORING.md
```diff
@@ -187,6 +187,8 @@ List of monitoring components - so called exporters:
 
 When dashboard creation or import succeeds you will see it on your dashboard list.
 
+`Note: Depends on the dashboards which rely on variouse variables, in order to visualize the data it is necessary to generate traffic to the components which are monitored`
```
to-bar:
```suggestion
`Note: For some dashboards, there is no data to visualize until there is traffic activity for the monitored component.`
```

seriva (approved):

rafzei (approved):

Issue #2204, opened 2021-04-13
[BACKPORT] [0.7] PostgreSQL replication with repmgr: hot_standby not set in config file


plirglo:
Backport of #1273  into 0.7 (include into 0.7.4 Release version)

---

**DoD checklist**

* [x] Changelog updated
* [x] COMPONENTS.md updated / doesn't need to be updated
* [x] Feature has automated tests
* [x] Automated tests passed (QA pipelines)
  * [x] apply
  * [ ] upgrade
* [x] Idempotency tested
* [x] Documentation added / updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Solution meets requirements and is done according to design doc
* [x] Usage compliant with license

przemyslavic:
It looks like one more backport is missing here: #1280 (PR #1845). 
@atsikham please let me know if it should also be backported?

przemyslavic:
Tested after #2317. 
✔️ `epicli apply:` `AWS`/`Azure` x `Ubuntu`/`RHEL`
Looks good 😀

PR #2311, opened 2021-05-10
Support only one replication type for PostgreSQL (with repmgr)

atsikham:
Task #2237 

* Modified preflight architecture check for PostgreSQL
* Removed PostgreSQL native replication
* Simplified PostgreSQL role structure
* `use_repmgr` setting is removed from specification, documentation updated

PR #2317, opened 2021-05-12
[Backport][0.7] Pgpool not showing Replication State

atsikham:
Yet another backport for task #2204

### DIFF
--- a/core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/configure-postgresql.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/configure-postgresql.yml
@@ -30,11 +30,6 @@
   when:
     - ansible_os_family == 'RedHat'
 
-- name: Enable postgresql service
-  systemd:
-    name: "{{ pg_service_name[ansible_os_family] }}"
-    enabled: yes
-
 - name: Change pg_hba.conf
   replace:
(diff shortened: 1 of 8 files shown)


przemyslavic:
👌

przemyslavic (approved):

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/configure-postgresql.yml
```diff
@@ -85,6 +80,17 @@
     backup: yes
   register: change_postgresql_conf
 
+# Placed here to avoid errors in log due to nonexistent directory
```
plirglo, line 83: What kind of errors? Which step generates this errors?
I know it's not part of this task to make refactoring of whole role, but i think we don't need to archive WAL logs since we don't use them in any part of our code (i didn't find any reference). But it's good step to start with Point-in-Time-Recovery solution for Postgres (which we  currently don't support). AFAIK repmgr uses psql to connect to database and replicate data.
OK, maybe we should leave it like it is, since it's backport and clean this in develop and maybe LTS since it can unnecessary fill up disk space.

atsikham: This task will be removed in latest versions with PR #2311. It can be removed here, but not sure that we need to do that for backport.

plirglo (approved):

PR #2320, opened 2021-05-12
Fix for RHUI client certificate expiration

plirglo:


seriva (approved):

atsikham (changes requested):
There are 12 commits for a few changes, not all of them are necessary.

seriva:
> There are 12 commits for a few changes, not all of them are necessary.

Mostly for testing I think, and that is why we squash when merging a branch. I don't see the need to change this.

plirglo:
> There are 12 commits for a few changes, not all of them are necessary.

I will squash while merging.

atsikham (approved):

Issue #2076, opened 2021-02-19
[UPGRADE] Upgrade of PgBouncer [Kubernetes]


erzetpe:
**Is your feature request related to a problem? Please describe.**
We need to upgrade PgBouncer if possible and needed to the newer version. This task should be done after we will upgrade PostgreSQL to v13.
This task applies to K8s application

**Describe the solution you'd like**
We want to upgrade if possible and needed to the newer version of PgBouncer.

**Describe alternatives you've considered**
None.

**Additional context**
This task should be done after upgrade of PostgreSQL to v13.

---

**DoD checklist**

* [ ] Changelog updated
* [ ] COMPONENTS.md updated / doesn't need to be updated
* [ ] Feature has automated tests
* [ ] Automated tests passed (QA pipelines)
  * [ ] apply
  * [ ] upgrade
* [ ] Idempotency tested
* [ ] Documentation added / updated / doesn't need to be updated
* [ ] All conversations in PR resolved
* [ ] Solution meets requirements and is done according to design doc
* [ ] Usage compliant with license

przemyslavic:
We are currently using PgBouncer v1.10.
As for now the latest available is v1.15.0.
Need to update OS packages and docker image for application deployment on kubernetes.

atsikham:
Steps:
1. Fix installation (apply) of a new verion(there is Python3 dependency - create a new role to install).
2. Upgrade with taking into account notes [here](https://github.com/epiphany-platform/epiphany/pull/1845#pullrequestreview-531363344).
3. ARM support?

plirglo:
This tasks basically includes two different tasks:
- Upgrade and configuration pgbouncer as a k8s instance
- Upgrade and configuration pgbouncer installed directly on database hosts as standalone application. In this case Python  dependency is needed (blocked by #2334). Created: [#2453](https://github.com/epiphany-platform/epiphany/issues/2453)

This might be good to split this task up to two different ones.

plirglo:
PR #2462

przemyslavic:
:heavy_check_mark: New installations have been tested, as well as the upgrade from v1.12. I've also deployed a sample application on k8s with `psql` and connected/executed test queries via `pgbouncer` to `postgresql` without any problems.

Issue #1861, opened 2020-11-17
 [UPGRADE] Upgrade PostgreSQL to v13


rafzei:
**Is your feature request related to a problem? Please describe.**
- PostgreSQL v10 has 20 vulnerabilities : [CVE-2019-10211, CVE-2018-16850, CVE-2018-1115, CVE-2019-10208, CVE-2019-10164, CVE-2018-1058, CVE-2018-10925, CVE-2018-10915, CVE-2020-10733, CVE-2020-14350,CVE-2019-9193, 
CVE-2020-14349, CVE-2019-10210, CVE-2018-1053, CVE-2018-1052, CVE-2017-15099, CVE-2020-1720, CVE-2019-10130](https://nvd.nist.gov/vuln/search/results?form_type=Advanced&results_type=overview&seach_type=all&query=cpe:2.3:a:postgresql:postgresql:10.0:*:*:*:*:*:*:*)
- PostgreSQL v12 have also additional functional, performance and security improvements comparing to PostgreSQL v10

**Describe the solution you'd like**
- Upgrade PostgreSQL to version 13
- Backport to supported Epiphany versions (please consider creating new issues for this purpose)

**Describe alternatives you've considered**
None

**Additional context**
External components related to PostgreSQL should be also upgraded: Rep Mgr, PGAudit, pgpool, PgBouncer. There are separate tasks created for those upgrades. We also need to check possibility of this upgrade in context of our components used for replication and if those components are still needed(can be replaced by PostgreSQL embedded or other components functionality).

---

**DoD checklist**

* [ ] Changelog updated
* [ ] COMPONENTS.md updated / doesn't need to be updated
* [ ] Feature has automated tests
* [ ] Automated tests passed (QA pipelines)
  * [ ] apply
  * [ ] upgrade
* [ ] Idempotency tested
* [ ] Documentation added / updated / doesn't need to be updated
* [ ] All conversations in PR resolved
* [ ] Solution meets requirements and is done according to design doc
* [ ] Usage compliant with license

atsikham:
* After Postgres upgrading related components have to be checked: `repmgr, pgpool, pgbouncer, pgaudit`. So this task may be large, especially with testing part.
* Let's discuss backporting tasks after implementation, during DoD.

mkyc:
we should split in 3 tasks: 
 - installation
 - spike about upgrade
 - upgrade

mkyc:
Also ensure that we have all for ARM architecture before we start here.

[...]

PR #2330, opened 2021-05-18
PostgreSQL v13 installation

atsikham:
Installation (apply) part of task #1861 for PostgreSQL, PgAudit, PgPool, Repmgr (except PgBouncer).
PgBouncer will be probably handled in the scope of #2076

* Some thoughts from [this](https://github.com/epiphany-platform/epiphany/pull/1845#pullrequestreview-531363344) comment were taken into account.
* Added Molecule Ansible tests.


atsikham:
Default postgresql.conf difference:

**Debian**

```
data_directory = '/var/lib/postgresql/13/main'          # use data in another directory
                                        # (change requires restart)
hba_file = '/etc/postgresql/13/main/pg_hba.conf'        # host-based authentication file
                                        # (change requires restart)
ident_file = '/etc/postgresql/13/main/pg_ident.conf'    # ident configuration file
                                        # (change requires restart)
# If external_pid_file is not explicitly set, no extra PID file is written.
external_pid_file = '/var/run/postgresql/13-main.pid'                   # write an extra PID file
                                        # (change requires restart)
port = 5432                             # (change requires restart)
unix_socket_directories = '/var/run/postgresql' # comma-separated list of directories
#password_encryption = md5              # md5 or scram-sha-256
ssl = on
ssl_cert_file = '/etc/ssl/certs/ssl-cert-snakeoil.pem'
ssl_key_file = '/etc/ssl/private/ssl-cert-snakeoil.key'
#log_destination = 'stderr'             # Valid values are combinations of
                                        # stderr, csvlog, syslog, and eventlog,
                                        # depending on platform.  csvlog
                                        # requires logging_collector to be on.
# This is used when logging to stderr:
#logging_collector = off                # Enable capturing of stderr and csvlog
                                        # into log files. Required to be on for
                                        # csvlogs.
                                        # (change requires restart)
# These are only used if logging_collector is on:
#log_directory = 'log'                  # directory where log files are written,
                                        # can be absolute or relative to PGDATA
#log_filename = 'postgresql-%Y-%m-%d_%H%M%S.log'        # log file name pattern,
                                        # can include strftime() escapes
#log_truncate_on_rotation = off         # If on, an existing log file with the
                                        # same name as the new log file will be
                                        # truncated rather than appended to.
                                        # But such truncation only occurs on
                                        # time-driven rotation, not on restarts
                                        # or size-driven rotation.  Default is
                                        # off, meaning append to existing files
                                        # in all cases.
#log_rotation_age = 1d                  # Automatic rotation of logfiles will
                                        # happen after that time.  0 disables.
#log_rotation_size = 10MB               # Automatic rotation of logfiles will
                                        # happen after that much log output.
                                        # 0 disables.
log_line_prefix = '%m [%p] %q%u@%d '            # special values:
log_timezone = 'Etc/UTC'
cluster_name = '13/main'                        # added to process titles if nonempty
stats_temp_directory = '/var/run/postgresql/13-main.pg_stat_tmp'
timezone = 'Etc/UTC'
lc_messages = 'C.UTF-8'                 # locale for system error message
lc_monetary = 'C.UTF-8'                 # locale for monetary formatting
lc_numeric = 'C.UTF-8'                  # locale for number formatting
lc_time = 'C.UTF-8'                             # locale for time formatting
#extension_destdir = ''                 # prepend path when loading extensions
                                        # and shared objects (added by Debian)
include_dir = 'conf.d'                  # include files ending in '.conf' from
```

**RedHat**

```
#data_directory = 'ConfigDir'           # use data in another directory
                                        # (change requires restart)
#hba_file = 'ConfigDir/pg_hba.conf'     # host-based authentication file
                                        # (change requires restart)
#ident_file = 'ConfigDir/pg_ident.conf' # ident configuration file
                                        # (change requires restart)

# If external_pid_file is not explicitly set, no extra PID file is written.
#external_pid_file = ''                 # write an extra PID file
                                        # (change requires restart)
#port = 5432                            # (change requires restart)
#unix_socket_directories = '/var/run/postgresql, /tmp'  # comma-separated list of directories
password_encryption = scram-sha-256             # md5 or scram-sha-256
#ssl = off
#ssl_cert_file = 'server.crt'
#ssl_key_file = 'server.key'
log_destination = 'stderr'              # Valid values are combinations of
                                        # stderr, csvlog, syslog, and eventlog,
                                        # depending on platform.  csvlog
                                        # requires logging_collector to be on.
# This is used when logging to stderr:
logging_collector = on                  # Enable capturing of stderr and csvlog
                                        # into log files. Required to be on for
                                        # csvlogs.
                                        # (change requires restart)
# These are only used if logging_collector is on:
log_directory = 'log'                   # directory where log files are written,
                                        # can be absolute or relative to PGDATA
log_filename = 'postgresql-%a.log'      # log file name pattern,
                                        # can include strftime() escapes
log_truncate_on_rotation = on           # If on, an existing log file with the
                                        # same name as the new log file will be
                                        # truncated rather than appended to.
                                        # But such truncation only occurs on
                                        # time-driven rotation, not on restarts
                                        # or size-driven rotation.  Default is
                                        # off, meaning append to existing files
                                        # in all cases.
log_rotation_age = 1d                   # Automatic rotation of logfiles will
                                        # happen after that time.  0 disables.
log_rotation_size = 0                   # Automatic rotation of logfiles will
                                        # happen after that much log output.
                                        # 0 disables.
log_line_prefix = '%m [%p] '            # special values:
log_timezone = 'UTC'
#cluster_name = ''                      # added to process titles if nonempty
#stats_temp_directory = 'pg_stat_tmp'
timezone = 'UTC'
lc_messages = 'en_US.UTF-8'                     # locale for system error message
lc_monetary = 'en_US.UTF-8'                     # locale for monetary formatting
lc_numeric = 'en_US.UTF-8'                      # locale for number formatting
lc_time = 'en_US.UTF-8'                         # locale for time formatting

#include_dir = '...'                    # include files ending in '.conf' from
```

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/apply.yml
```diff
@@ -61,6 +63,13 @@
           {{ (_defined | selectattr('config.enabled') | list) + _undefined }}
         _unsupported_enabled_extensions: >-
           {{ _enabled_extensions | selectattr('name', 'in', unsupported_postgres_extensions[_pg_arch]) | list }}
+
+    - name: Check the number of PostgreSQL hosts for Repmgr
+      when: postgresql.specification.extensions.replication.enabled | default(false)
+      assert:
+        that: groups.postgresql | length <= 2
```
plirglo: Why we need to limit this to 2? I think after postgres role refactoring you've done here it's quite easy to handle bigger clusters, isn't it?

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/centos-7/requirements.x86_64.txt
```diff
@@ -162,8 +162,8 @@ hashicorp/vault-k8s:0.10.0
 vault:1.7.0
 # applications
 epiphanyplatform/ignite:2.9.1
-bitnami/pgpool:4.1.1-debian-10-r29
-brainsam/pgbouncer:1.12
+bitnami/pgpool:4.2.2
```
plirglo, line 167: this change is not part of this task, correct? You just put it here in advance before further work. Am i right?

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/centos-7/requirements.x86_64.txt
```diff
@@ -162,8 +162,8 @@ hashicorp/vault-k8s:0.10.0
...
+bitnami/pgbouncer:1.15.0
```
plirglo, line 168: Same here

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/molecule/README.adoc
```diff
@@ -0,0 +1,35 @@
+= Installation guide
```
seriva: Why not MD?:) Also would be cool if we can run these tests from the container and not only externally.

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/apply.yml
atsikham: This was my assumption to do the same way as we did before.

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/centos-7/requirements.x86_64.txt
atsikham: correct, can be removed if necessary

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/apply.yml
atsikham: Created a new task for that #2346

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/molecule/README.adoc
atsikham: This is a personal preference, I will change that to markdown.

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/molecule/debian-repmgr/prepare.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Converge
```
to-bar:
```suggestion
- name: Prepare
```
Shouldn't be `Prepare`?

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/molecule/debian-repmgr/verify.yml
```diff
@@ -0,0 +1,53 @@
...
+            port: 5432
+            state: started
+            delay: 0
+            timeout: 3
+          register: port_status
+
+        - name: Verify that port status is not changed
+          assert:
+            that:
+              - not port_status.changed
```
to-bar: In which case the `wait_for` task may set `changed` to true? When port is not open, the task fails and `changed` is still `false`.

I would use `listen_ports_facts` module instead of `wait_for`.

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/molecule/debian-repmgr/verify.yml
```diff
@@ -0,0 +1,53 @@
...
+      block:
+        - name: Ensure that postgresql service is started and enabled
+          systemd:
+            name: postgresql
+            enabled: true
+          register: svc_status
+
+        - name: Ensure that postgresql@13-main service is started and enabled
+          systemd:
+            name: postgresql
```
to-bar:
```suggestion
            name: postgresql@13-main
```

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/molecule/debian-repmgr/verify.yml
```diff
@@ -0,0 +1,53 @@
...
+            enabled: true
+          register: svc_main_status
+
+        - name: Verify that service statuses are not changed
+          assert:
+            that:
+              - not svc_status.changed
+              - not svc_main_status.changed
+
+    - name: Check that repmgrd service is started and enabled
```
to-bar: How about having common code for single and repmgr setup and just run repmgr (additional) tasks conditionally?

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/molecule/redhat-repmgr/prepare.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Converge
```
to-bar:
```suggestion
- name: Prepare
```

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/molecule/debian-repmgr/verify.yml
```diff
@@ -0,0 +1,53 @@
...
+          assert:
+            that:
+              - not port_status.changed
+
+    - name: Check that PostgreSQL services are started and enabled
+      block:
+        - name: Ensure that postgresql service is started and enabled
+          systemd:
+            name: postgresql
+            enabled: true
```
to-bar:
```suggestion
            enabled: true
            state: started
```
Shouldn't we have also `state`?

However, I would use `service_facts` module instead of `systemd`.

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/molecule/redhat-single-node/prepare.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Converge
```
to-bar:
```suggestion
- name: Prepare
```

docs/home/COMPONENTS.md
```diff
@@ -31,12 +31,12 @@ Note that versions are default versions and can be changed in certain cases thro
...
 | Postgres Exporter          | 0.9.0    | https://github.com/prometheus-community/postgres_exporter | [Apache License 2.0](https://www.apache.org/licenses/LICENSE-2.0) |
-| PostgreSQL                 | 10       | https://www.postgresql.org/                           | https://opensource.org/licenses/postgresql                        |
+| PostgreSQL                 | 13       | https://www.postgresql.org/                           | [PostgreSQL license](http://www.postgresql.org/about/licence/)    |
 | HAProxy                    | 2.2.2    | https://www.haproxy.org/                              | [GNU General Public License 2.0](https://www.gnu.org/licenses/old-licenses/gpl-2.0.html) |
-| PGAudit                    | 1.2.2    | https://github.com/pgaudit/pgaudit                    | [PostgreSQL license](http://www.postgresql.org/about/licence/)    |
+| PgAudit                    | 1.5.0    | https://github.com/pgaudit/pgaudit                    | [PostgreSQL license](http://www.postgresql.org/about/licence/)    |
 | PgBouncer                  | 1.10.0   | https://github.com/pgbouncer/pgbouncer                | [ISC License](https://opensource.org/licenses/isc)                |
```
plirglo, line 37: Maybe we should also add here pgbouncer k8s version here? But i'm not sure, it's not a part of this task. Can be added with pgbouncer update.

plirglo (dismissed):

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/extensions/replication/extension.yml
```diff
@@ -21,19 +21,22 @@
...
 
-- name: Extensions | Repmgr | Create super user in postgresql
-  become_user: postgres
-  postgresql_user:
-    user: "{{ specification.extensions.replication.privileged_user_name }}"
-    password: "{{ specification.extensions.replication.privileged_user_password }}"
-    role_attr_flags: LOGIN,SUPERUSER
-
-- name: Extensions | Repmgr | Create replication user in postgresql
+- name: Extensions | Repmgr | Create replication and super users in PostgreSQL
```
to-bar:
```suggestion
- name: Extensions | Repmgr | Create super user and replication user in PostgreSQL
```

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/extensions/replication/extension.yml
```diff
@@ -102,58 +93,44 @@
...
-# --- Configure primary node ---
-
-- name: Extensions | Repmgr | Tasks to run only on primary node
-  when: groups.postgresql[0] == inventory_hostname
+- name: Extensions | Repmgr | Check cluster status
+  become_user: postgres
+  command: "{{ repmgr.bindir[ansible_os_family] }}/repmgr cluster show --csv"
+  register: pg_cluster_status
+  changed_when: false
+  failed_when: false
```
to-bar:
```suggestion
  # rc = 25 means that there is an error with at least one registered node
  # sometimes 'repmgr cluster show' returns 25 code with empty stderr and no explanation
  failed_when: pg_cluster_status.rc not in [0, 25]
```

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/extensions/replication/extension.yml
```diff
@@ -102,58 +93,44 @@
...
+  no_log: true
```
to-bar: Does it return any sensitive data?

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/extensions/replication/extension.yml
```diff
@@ -102,58 +93,44 @@
...
+
+# 1. There is a preflight check that no more than 2 cluster nodes are supported
+# 2. rc = 25 means that there is an error with at least one registered node
+#    sometimes 'repmgr cluster show' returns 25 code with empty stderr and no explanation
+- name: Extensions | Repmgr | Set 'pg_is_clustered' fact
+  set_fact:
+    pg_is_clustered: "{{ pg_cluster_status.rc in [0, 25] and pg_cluster_status.stdout_lines | length  > 1 }}"
+
+- name: Extensions | Repmgr | Configure PostgreSQL cluster
+  when: not pg_is_clustered|bool
```
to-bar:
```suggestion
  when: not pg_is_clustered
```

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/extensions/replication/extension.yml
```diff
@@ -162,56 +139,63 @@
...
-            state: restarted
+    - name: Extensions | Repmgr | Restart and enable Repmgr service
+      systemd:
+        name: "{{ repmgr.service_name[ansible_os_family] }}"
+        state: restarted
+        enabled: true
 
-    - name: Extensions | Repmgr | Check if node is not attached to repmgr
+- name: Extensions | Repmgr | Check already configured cluster
+  when: pg_is_clustered|bool
```
to-bar:
```suggestion
  when: pg_is_clustered
```

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/main.yml
```diff
@@ -1,141 +1,150 @@
...
     _packages:
       Debian:
-        - postgresql-10
-        - postgresql-contrib-10
-        - python-psycopg2 # required for postgresql ansible management
+        - postgresql-client-common
+        - postgresql-common
+        - postgresql-client-13
+        - postgresql-13
+        - postgresql-contrib-13
```
to-bar:
```suggestion
```

> part of the postgresql-xx package in version 10 and later

Source: https://www.postgresql.org/download/linux/ubuntu/

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/main.yml
```diff
@@ -1,141 +1,150 @@
...
+- name: Ensure that PostgreSQL service is started
+  block:
+    - name: Ensure that PostgreSQL service is started
+      systemd:
+        name: "{{ pg.service_name[ansible_os_family] }}"
+        state: started
+        enabled: true
+
+    - &check-debian-service
+      name: Debian | Check that the main service is started
```
to-bar: Do we need this? In PG 10 the parent service starts it's children.

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/templates/pg_hba.conf.j2
```diff
@@ -0,0 +1,98 @@
...
+host    all             all             0.0.0.0/0               scram-sha-256
+# IPv6 local connections:
+host    all             all             ::1/128                 scram-sha-256
+# Allow replication connections from localhost, by a user with the
+# replication privilege.
+local   replication     all                                     peer
+host    replication     all             127.0.0.1/32            scram-sha-256
+host    replication     all             ::1/128                 scram-sha-256
+
+{% if specification.extensions.replication.enabled is defined and specification.extensions.replication.enabled %}
```
to-bar:
```suggestion
{% if specification.extensions.replication.enabled | default(False) %}
```

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/apply.yml
to-bar:
```suggestion
        that: groups.postgresql | length == 2
```
How about requiring 2 hosts when replication is enabled?

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/templates/postgresql.conf.j2
```diff
@@ -0,0 +1,781 @@
...
+
+# These options allow settings to be loaded from files other than the
+# default postgresql.conf.  Note that these are directives, not variable
+# assignments, so they can usefully be given more than once.
+
+#include_dir = '...'			# include files ending in '.conf' from
+					# a directory, e.g., 'conf.d'
+#include_if_exists = '...'		# include file only if it exists
+#include = '...'			# include file
+include = 'postgresql-epiphany.conf'	# Epiphany managed configuration that overrides settings above
```
to-bar, line 774: How about making this file shorter?
I would keep only the first 33 lines + `CONFIG FILE INCLUDES` section, since we have `backup: true` in the `template` task so one may check the original content.

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/centos-7/requirements.x86_64.txt
```diff
@@ -78,8 +78,8 @@ perl-libs # for vim
 perl-Pod-Perldoc # for vim
 perl-Pod-Simple # for vim
 perl-Pod-Usage # for vim
-pgaudit12_10
-pgbouncer-1.10.*
+pgaudit15_13
+#pgbouncer-1.10.*
```
to-bar, line 84: Please add a comment why this is commented out.

core/src/epicli/data/common/defaults/configuration/postgresql.yml
```diff
@@ -40,14 +40,6 @@ specification:
               - name: wal_level
                 value: replica
                 when: replication
-          - name: Archiving
```
to-bar, line 43: How about keeping these settings but in accordance with Repmgr docs?

> archive_mode
>
>
> We suggest setting archive_mode to on (and archive_command to /bin/true; see below) even if you are currently not planning to use WAL file archiving.
> 
> This will make it simpler to set up WAL file archiving if it is ever required, as changes to archive_mode require a full PostgreSQL server restart, while archive_command changes can be applied via a normal configuration reload.
> 
> However, repmgr itself does not require WAL file archiving.

Source: https://repmgr.org/docs/repmgr.html#CONFIGURATION-POSTGRESQL

to-bar (changes requested):

docs/home/COMPONENTS.md
atsikham: Should be added, but I also would do that in the scope of the task you mentioned. I didn't add that as my assumption that PgBouncer version and license should be the same for any installation type.

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/apply.yml
atsikham: Seems like hard restriction to user, as repmgr could be installed without a second node. Now I don't see any case to do so, but who knows what user wants?

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/templates/postgresql.conf.j2
atsikham: Usually you want to modify something in a standard config file and use these comments to read a description and check defaults in particular. Leaving that only in backup file will force user switch to documentation as there is a 99% chance that they don't know how it works internally.

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/centos-7/requirements.x86_64.txt
atsikham: discussed with @plirglo that it can be added here for now

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/centos-7/requirements.x86_64.txt
atsikham: discussed with @plirglo that it can be added here for now

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/molecule/debian-repmgr/verify.yml
atsikham: Not sure how you would like to do it conditionally in a separate `verify` playbook, but hope with the latest changes it's ok.

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/extensions/replication/extension.yml
atsikham: As I remember it's done here as this command doesn't work in one of OS distributions just after package installation with no configuration. So If this command fails, `pg_is_clustered` is set to false.

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/extensions/replication/extension.yml
atsikham: It may fail, but it doesn't mean that something is wrong, only that there is no cluster configured. Because of that its output is hidden.

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/templates/postgresql.conf.j2
atsikham: But.. then it will be necessary to maintain, so let's make a decision.

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/main.yml
atsikham: In 13th version it's also so, but this service doesn't care about its children state/status.

```text
cat /etc/systemd/system/multi-user.target.wants/postgresql.service
# systemd service for managing all PostgreSQL clusters on the system. This
# service is actually a systemd target, but we are using a service since
# targets cannot be reloaded.
[Unit]
Description=PostgreSQL RDBMS
[Service]
Type=oneshot
ExecStart=/bin/true
ExecReload=/bin/true
RemainAfterExit=on
[Install]
WantedBy=multi-user.target
```

to-bar (approved):

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/extensions/replication/extension.yml
atsikham: Changed by @to-bar

core/src/epicli/data/common/ansible/playbooks/roles/preflight/tasks/apply.yml
atsikham: changed by @to-bar

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/extensions/replication/extension.yml
atsikham: changed by @to-bar

plirglo (approved):

PR #2347, opened 2021-05-27
Elasticsearch upgrade improvements

to-bar:
- Wait for yellow status with additional conditions
- Handle copying certificate files on re-run when first host is skipped
- Store initial cluster status and assert it after upgrade


seriva (approved):
Looks good, just one questions about the retries of wait-for-node-to-join.yml and wait-for-shard-allocation.yml. This was tested with some data inside? As that might influence a mechanism like this.

to-bar:
@seriva Good point. I tested upgrade only with really small amount of data and it was quick but I've chosen bigger numbers for `retries` just in case. The steps are optimized for performance but it would be cool to test upgrade of a cluster with bigger amount of data (with many indices).

atsikham (approved):
LGTM

PR #2356, opened 2021-05-31
[Bug][Kafka] Incorrect number of brokers/queues available after scaling up/down

plirglo:
meta.properties is re-created by Kafka every time when service is starting and file does not exist. It contains broker.id and cluster.id which are shared between brokers using zookeeper.


seriva (approved):

seriva:
I think this would be something we want backport to 1.0.x, 0.9.x and 0.8.x Lets not bother to 0.7.x as it will be deprecated soon.

przemyslavic (approved):

Issue #2418, opened 2021-06-30
[SPIKE] Research support for additional Istio components and configurations.


seriva:
**Is your feature request related to a problem? Please describe.**
Currently we install Istio only with the basic profiles as described here:

https://istio.io/latest/docs/setup/additional-setup/config-profiles/

**Describe the solution you'd like**
While the above profiles are good for basic usage, real word usage often involves specific configuration of the Istio control plane components. This spike should document what is required by teams and what what things we should support:

**Describe alternatives you've considered**
-

**Additional context**
More information here:
https://istio.io/latest/docs/setup/additional-setup/customize-installation/

---

**DoD checklist**

* [ ] Changelog updated
* [ ] COMPONENTS.md updated / doesn't need to be updated
* [ ] Feature has automated tests
* [ ] Automated tests passed (QA pipelines)
  * [ ] apply
  * [ ] upgrade
* [ ] Idempotency tested
* [ ] Documentation added / updated / doesn't need to be updated
* [ ] All conversations in PR resolved
* [ ] Solution meets requirements and is done according to design doc
* [ ] Usage compliant with license
* [ ] Backport tasks created / doesn't need to be backported

erzetpe:
Closing because of deprecation of Istio component.

PR #2419, opened 2021-06-30
Added documentation for steps to upgrade cluster with Istio installed.

seriva:
We currently only support installing Istio through its pre-defined profiles:

https://istio.io/latest/docs/setup/additional-setup/config-profiles/

Fixing this issue properly would require adding support for customized configurations:

https://istio.io/latest/docs/setup/additional-setup/customize-installation/

Before we can do that I feel we need to spike it first, and more importantly there should be a need for it as currently no one uses Istio. 

So I documented a work around and issued a new spike if we ever need to pick this up again: #2418


docs/home/howto/UPGRADE.md
```diff
@@ -255,3 +255,31 @@ With the latest Epiphany version it's possible to upgrade RabbitMQ to v3.8.9.
 It requires Erlang system packages upgrade that is done automatically to v23.1.4.
 Upgrade is performed in offline mode after stopping all RabbitMQ nodes.
 [Rolling upgrade](https://www.rabbitmq.com/upgrade.html#rolling-upgrades) is not supported by Epiphany and it is advised not to use this approach when Erlang needs to be upgraded.
+
+## Kubernetes upgrade
+
+---
+**NOTE**
+
+If the cluster to be upgraded has the Istio control plane application deployed you will run into issues when trying to upgrade the cluster.
```
erzetpe: This statement is written in a little bit confusing way. Probably the verb is missing.

docs/home/howto/UPGRADE.md
```diff
@@ -255,3 +255,31 @@ With the latest Epiphany version it's possible to upgrade RabbitMQ to v3.8.9.
...
+The default [profiles](https://istio.io/latest/docs/setup/additional-setup/config-profiles/) we currently support for installing Istio only deploy a single replica for the control services with a `PodDisruptionBudgets` value of 0. This will results in the following error while its draining a these pods during an upgrade:
+
+```shell
+Cannot evict pod as it would violate the pods disruption budget.
+```
+
+As we currently dont support any kind of advanced configuration of the Istio control plane components outside of the default profiles we need to scale up all components manually before the upgrade. This can be done with the following command:
```
erzetpe:
```suggestion
As we currently don't support any kind of advanced configuration of the Istio control plane components outside of the default profiles we need to scale up all components manually before the upgrade. This can be done with the following command:
```

docs/home/howto/UPGRADE.md
```diff
@@ -255,3 +255,31 @@ With the latest Epiphany version it's possible to upgrade RabbitMQ to v3.8.9.
...
+
+```shell
+kubectl scale deploy -n istio-system --replicas=2 --all 
+```
+
+After the upgrade, the deployements can be scaled down to there original capacity:
```
erzetpe:
```suggestion
After the upgrade, the deployements can be scaled down to the original capacity:
```

docs/home/howto/UPGRADE.md
erzetpe:
```suggestion
If the cluster has to be upgraded has the Istio control plane application deployed you will run into issues when trying to upgrade the cluster.
```

erzetpe (changes requested):

erzetpe (approved):

rafzei (approved):

PR #2428, opened 2021-07-12
Feature/upgrade pgpool (only new installations)

plirglo:
- Configuration with pgpool version 4.2.3 (upgraded)
- Added SHA-256 password support
- Added configurable MD5 or SHA-256 password usage for PostgreSQL and Pgpool
- libpgpool.sh scipt is mostly developed by Bitnami, i've only added steps to create SHA-256 password encryption for epi_pgpool_postgres_admin - original file is located [here](https://github.com/bitnami/bitnami-docker-pgpool/blob/4.2.3-debian-10-r46/4/debian-10/rootfs/opt/bitnami/scripts/libpgpool.sh)
- some other modifications in init scripts required by newer version of pgpool
- minor required changes in pgpool.conf file
- modified core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/main.yml file to check for primary node before starting pgpool configuration (Fix for #1702)

### DIFF
--- a/core/src/epicli/data/common/ansible/playbooks/roles/applications/templates/pgpool/01-secrets.yml.j2
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/applications/templates/pgpool/01-secrets.yml.j2
@@ -11,4 +11,3 @@ data:
   pgpool_admin_password: "{{ data.pgpool.secrets.pgpool_admin_password | b64encode }}"
   pgpool_postgres_password: "{{ data.pgpool.secrets.pgpool_postgres_password | b64encode }}"
   pgpool_sr_check_password: "{{ data.pgpool.secrets.pgpool_sr_check_password | b64encode }}"
-  pgpool_sha_encryption_key: "{{ data.pgpool.secrets.pgpool_sha_encryption_key | b64encode }}"
--- a/core/src/epicli/data/common/ansible/playbooks/roles/applications/templates/pgpool/04-deployment.yml.j2
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/applications/templates/pgpool/04-deployment.yml.j2
@@ -76,13 +76,10 @@ spec:
             - name: pgpool-passwords
               mountPath: /opt/bitnami/pgpool/secrets/pgpool_admin_password
               subPath: pgpool_admin_password
-            - name: pgpool-passwords
-              mountPath: /opt/bitnami/pgpool/secrets/pgpool_sha_encryption_key
-              subPath: pgpool_sha_encryption_key
-            - name: pgpool-config-dir
+            - name: pgpool-config-files
               mountPath: /opt/bitnami/pgpool/conf/pgpool.conf
               subPath: pgpool.conf
-            - name: pgpool-config-dir
+            - name: pgpool-config-files
               mountPath: /opt/bitnami/pgpool/conf/pool_hba.conf
               subPath: pool_hba.conf
             - name: pgpool-shared-init-dir
@@ -94,9 +91,6 @@ spec:
             - name: pgpool-init-scripts
               mountPath: /docker-entrypoint-initdb.d/init_pcppass.sh
               subPath: init_pcppass.sh
-            - name: pgpool-init-scripts
-              mountPath: /opt/bitnami/scripts/libpgpool.sh
-              subPath: libpgpool.sh
       initContainers:
 {#      pgpool-init container creates/synchronizes pool_passwd file, users with passwords are fetched directly from PostgreSQL server #}
         - name: pgpool-init
@@ -105,11 +99,8 @@ spec:
 {% else %}
           image: {{ data.image.path }}
 {% endif %}
-          command: ["/bin/sh", "-c"]
-          args: 
-            - /epiphany/scripts/init_pool_passwd_file.sh;
-              cp /epiphany/bitnami/pgpool/conf/pgpool.conf /opt/bitnami/pgpool/conf/ || echo "Error while copying pgpool.conf";
-              cp /epiphany/bitnami/pgpool/conf/pool_hba.conf /opt/bitnami/pgpool/conf/ || echo "Error while copying pool_hba.conf";
+          command:
+            - /epiphany/scripts/init_pool_passwd_file.sh
           envFrom:
             - configMapRef:
                 name: pgpool-container-env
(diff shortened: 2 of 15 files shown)


core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/main.yml
```diff
@@ -157,35 +157,31 @@
...
+                                                                 | selectattr('enabled')
+                                                                 | map(attribute='pgpool')
+                                                                 | list
+                                                                 | first
+                                                                 | default({}) }}
       when: 
         - not in_recovery_state.in_recovery
   when:
-    - pgpool.env is defined
+    - applications_role_config.specification.applications | selectattr("name", "equalto", "pgpool") | selectattr('enabled')
```
atsikham: In this condition it's better to use defaults like in 'pgpool' var. In fact if 'pgpool' var is not empry dictionary, then pgpool is enabled. WDYT about something like

`pgpool.keys()|length > 0`

core/src/epicli/data/common/ansible/playbooks/roles/applications/templates/pgpool/04-deployment.yml.j2
```diff
@@ -113,6 +122,14 @@ spec:
               subPath: pgpool_postgres_password
             - name: pgpool-shared-init-dir
               mountPath: /epiphany/shared_dir
+            - name: pgpool-config-files
+              mountPath: /epiphany/bitnami/pgpool/conf/pgpool.conf
+              subPath: pgpool.conf
+            - name: pgpool-config-files
+              mountPath: /epiphany/bitnami/pgpool/conf/pool_hba.conf
+              subPath: pool_hba.conf
```
atsikham, line 130: My suggestion is to split pgpool-config-files configmap to different ones and mount config files into `/epiphany/bitnami/pgpool/conf/` at one time without `subPath` that has an affect of missing possibility to update config without pod restart.

atsikham (changes requested):

core/src/epicli/data/common/ansible/playbooks/roles/applications/templates/pgpool/04-deployment.yml.j2
plirglo: I think this needs to stay as it is since we also mounting here pool_passwd file in the same location but from different volume. But we can talk, i'm open for suggestions.

core/src/epicli/data/common/defaults/configuration/postgresql.yml
```diff
@@ -2,6 +2,8 @@ kind: configuration/postgresql
 title: PostgreSQL
 name: default
 specification:
+  configuration:
+    password_encryption: md5  # md5 or scram-sha-256
```
erzetpe, line 6: Why by default we don't have this set to scram?

erzetpe (changes requested):

core/src/epicli/data/common/defaults/configuration/postgresql.yml
plirglo: I wanted to leave it as it is since for upgrade existing clusters (with existing md5 passwords) that might be difficult to force users to re-create passwords. I think this should be applied very intentionally by users. But maybe when all upgrade procedure will be ready we can finally decide what is better.

core/src/epicli/data/common/ansible/playbooks/roles/applications/templates/pgpool/01-secrets.yml.j2
```diff
@@ -11,3 +11,4 @@ data:
   pgpool_admin_password: "{{ data.pgpool.secrets.pgpool_admin_password | b64encode }}"
   pgpool_postgres_password: "{{ data.pgpool.secrets.pgpool_postgres_password | b64encode }}"
   pgpool_sr_check_password: "{{ data.pgpool.secrets.pgpool_sr_check_password | b64encode }}"
+  pgpool_sha_encryption_key: "{{ data.pgpool.secrets.pgpool_sha_encryption_key | b64encode }}"
```
rafzei, line 14:
```suggestion
  {% if data.pgpool.secrets.pgpool_sha_encryption_key is defined %}
    pgpool_sha_encryption_key: "{{ data.pgpool.secrets.pgpool_sha_encryption_key | b64encode }}"
  {% endif %}
```
If data.pgpool.secrets.pgpool_sha_encryption_key will be optional, shouldn't we add condition here?
I think we can also use this one if we are sourcing it: postgresql_role_config.specification.configuration.password_encryption
would be nice to keep it consistent

core/src/epicli/data/common/ansible/playbooks/roles/applications/templates/pgpool/04-deployment.yml.j2
```diff
@@ -76,10 +76,13 @@ spec:
             - name: pgpool-passwords
               mountPath: /opt/bitnami/pgpool/secrets/pgpool_admin_password
               subPath: pgpool_admin_password
-            - name: pgpool-config-files
+            - name: pgpool-passwords
+              mountPath: /opt/bitnami/pgpool/secrets/pgpool_sha_encryption_key
+              subPath: pgpool_sha_encryption_key
```
rafzei, line 81: Maybe add a condition here as well? Why mount names are the same? is it intended?

core/src/epicli/data/common/defaults/configuration/applications.yml
```diff
@@ -119,13 +119,16 @@ specification:
...
+        PGPOOL_CHILD_LIFE_TIME: 0
         PGPOOL_POSTGRES_PASSWORD_FILE: /opt/bitnami/pgpool/secrets/pgpool_postgres_password
         PGPOOL_SR_CHECK_PASSWORD_FILE: /opt/bitnami/pgpool/secrets/pgpool_sr_check_password
         PGPOOL_ADMIN_PASSWORD_FILE: /opt/bitnami/pgpool/secrets/pgpool_admin_password
+        PGPOOLKEYFILE: /opt/bitnami/pgpool/secrets/pgpool_sha_encryption_key
       secrets:
         pgpool_postgres_password: PASSWORD_TO_CHANGE
         pgpool_sr_check_password: PASSWORD_TO_CHANGE
         pgpool_admin_password: PASSWORD_TO_CHANGE
+        pgpool_sha_encryption_key: PASSWORD_TO_CHANGE  # used only if SCRAM SHA-256 passwords encryption enabled
```
rafzei, line 131: If we are going to leave md5 as default I would comment it out and put into documentation information about the possibility of using SCRAM

plirglo:
Closing this PR since new one #2428 includes also those changes

Issue #2074, opened 2021-02-19
[UPGRADE] Upgrade repmgr to 5.2.1


erzetpe:
**Is your feature request related to a problem? Please describe.**
We need to upgrade Replication Manager if possible and needed to the newer version. 

**Describe the solution you'd like**
We want to upgrade if possible and needed to the newer version of Replication Manager.

**Describe alternatives you've considered**
None.

**Additional context**
This task can be done before upgrading PostgreSQL.

---

**DoD checklist**

* [x] Changelog updated
* [x] COMPONENTS.md updated / doesn't need to be updated
* [x] Feature has automated tests
* [x] Automated tests passed (QA pipelines)
  * [x] apply
  * [x] upgrade
* [x] Idempotency tested
* [x] Documentation added / updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Solution meets requirements and is done according to design doc
* [x] Usage compliant with license

mkyc:
isn't it duplicate off #1081 ?

rpudlowski93:
Yes it is, I closed the task #1081

atsikham:
https://repmgr.org/docs/current/upgrading-repmgr-extension.html

plirglo:
Starting with PostgreSQL 13 there is a name and value change of one of the parameters used for the streaming replication of a replica server. Until Postgres 12 it was called “wal_keep_segments” but with PostgreSQL 13 it will become “wal_keep_size”.

to-bar:
It was decided to upgrade repmgr for PostgreSQL 10 first, then integrate it with PostgreSQL 13. This means that this task may be done independently from PostgresSQL upgrade.

to-bar:
Moved to Needs review because of PR #2451

przemyslavic:
Repmgr fails to upgrade on Azure CentOS environments. Reproduced while trying to upgrade from epicli v1.0 and v1.1.
```
2021-07-26T09:56:21.0414645Z[38;21m09:56:21 INFO cli.engine.ansible.AnsibleCommand - TASK [postgresql : repmgr for PG 10 | Start repmgr service] ********************
2021-07-26T09:56:21.5055916Z[38;21m09:56:21 INFO cli.engine.ansible.AnsibleCommand - changed: [ci-10todevazceflannel-postgresql-vm-1]
[...]

PR #2459, opened 2021-07-22
Fix repmgr upgrade

to-bar:
`specification` variable cannot be used as task var because it is overridden by upgrade of `node-exporter`, which uses unnamed `include_vars` which unfortunately has higher [precedence](https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#ansible-variable-precedence). We *should not use* `include_vars` without the `name` parameter in `upgrade` role to limit top level vars with higher precedence than task vars. Refactoring is needed.

### DIFF
--- a/core/src/epicli/data/common/ansible/playbooks/roles/backup/tasks/postgresql.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/backup/tasks/postgresql.yml
@@ -10,7 +10,7 @@
 - name: Check if database is running on node0 database server
   become: true
   become_user: postgres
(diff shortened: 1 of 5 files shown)


atsikham (dismissed):

atsikham (dismissed):

przemyslavic (dismissed):

przemyslavic:
Please note my comment: https://github.com/epiphany-platform/epiphany/issues/2074#issuecomment-886575625

atsikham (dismissed):
LGTM

przemyslavic (approved):

atsikham (approved):

Issue #196, opened 2019-04-02
[FEATURE REQUEST] Advanced configuration type validation


toszo:
**Is your feature request related to a problem? Please describe.**
Currently there are is only basic type schema validations inside Epicli using jsonschema version 7.

**Describe the solution you'd like**
Add advanced type validations (enums, regex, ranges etc) inside the schema in the following folders:

- /workspaces/epiphany/core/src/epicli/data/azure/validation/infrastructure
- /workspaces/epiphany/core/src/epicli/data/aws/validation/infrastructure
- /workspaces/epiphany/core/src/epicli/data/azure/validation/infrastructure

**Describe alternatives you've considered**
*

**Additional context**
We probably want to break this down into smaller stories because of the amount of documents and fields to cover stories.

---

**DoD checklist**

- Changelog
  - [ ] updated
  - [ ] not needed
- COMPONENTS.md
  - [ ] updated
  - [ ] not needed
- Schema
  - [ ] updated
  - [ ] not needed
- Backport tasks
  - [ ] created
  - [ ] not needed
- Documentation
  - [ ] added
  - [ ] updated
  - [ ] not needed
- [ ] Feature has automated tests
- [ ] Automated tests passed (QA pipelines)
  - [ ] apply
  - [ ] upgrade
  - [ ] backup/restore
- [ ] Idempotency tested
- [ ] All conversations in PR resolved
- [ ] Solution meets requirements and is done according to design doc
- [ ] Usage compliant with license

seriva:
Not going todo this anymore for Epiphany.

PR #2470, opened 2021-07-28
Added basic configuration type and schema validation.

seriva:
Added basic document type validation to Epicli. This will be the scaffolding to add more advanced features later on ... maybe:

- Advanced type validations for example: enums, regex etc. (#196)
- Schema documentation + documentation generation. (#2471, #2472)
- Cross document rules and validation.  (#2473)

Steps done to add each documents:
1: Reviewed YAML default document if any structures needed to be uncommented
2: Convert YAML default to JSON default: https://www.json2yaml.com/convert-yaml-to-json
4: Convert JSON default to  JSON schema: https://www.liquid-technologies.com/online-json-to-schema-converter
4: Convert JSON schema to YAML schema: https://www.json2yaml.com/

So as most of this is generated, reviewing every document IMO is not needed and more for @przemyslavic  to test. 

Also added some minor changes to validate the actual schema we want to use to validate the document.


core/src/epicli/cli/engine/schema/SchemaValidator.py
```diff
@@ -46,8 +45,10 @@ def run_for_individual_documents(self):
...
 
             # Assert the schema
+            schema_dic = objdict_to_dict(schema)
             try:
-                validate(instance=objdict_to_dict(doc), schema=objdict_to_dict(schema))
+                Draft7Validator.check_schema(schema_dic)
+                validate(instance=objdict_to_dict(doc), schema=schema_dic)
             except Exception as e:
                 self.logger.error(f'Failed validating: {doc.kind}')
                 self.logger.error(e)
```
atsikham: Looks like a code duplication

core/src/epicli/cli/engine/schema/SchemaValidator.py
```diff
@@ -61,8 +62,10 @@ def run(self):
             if hasattr(schema['properties']["specification"], '$ref'):
                 if schema['properties']["specification"]['$ref'] == '#/definitions/unvalidated_specification':
                     self.logger.warn('No specification validation for ' + doc.kind)
+            schema_dic = objdict_to_dict(schema)
```
atsikham: Could you explain the purpose of ObjDict and this operation with the same `src_class` and `dst_class` values?

```python
def objdict_to_dict(something):
    """Recursively convert any ObjDict to python dictionary."""
    return _nested_dict_to_dict(something, src_class=dict, dst_class=dict)
```

core/src/epicli/data/common/validation/configuration/helm.yml
```diff
@@ -1 +1,8 @@
-$ref: '#/definitions/unvalidated_specification'
\ No newline at end of file
+"$id": "#/specification"
+title: "Helm specification schema"
+description: "Helm specification schema"
+type: object
+properties:
+  apache_epirepo_path:
+    type: string
+
```
erzetpe: Do the needful. :)

atsikham: need to change

atsikham (changes requested):

core/src/epicli/cli/engine/schema/SchemaValidator.py
seriva: ObjDict is just a helper class with some extension methods to query a dictionary like so:

```python
dict.value.nested_value.etc ....
```

Instead of like this with a normal dictionary :

```python
dict['value']['nested_value']['etc'] ...
```

Mostly used to make some of the operations in the infrastructure builders a bit more readable:

image

We use objdict_to_dict and dict_to_objdict function to switch between them as some external libraries for loading and validating dont work with ObjDict`s but only with plain python Dict`s

As for the src_class and dst_class values, that is just from a helper method to indicate what to convert to what, ObjDict->Dict or Dict ->ObjDict

dict_to_objdict will use: 

```python
return _nested_dict_to_dict(something, src_class=dict, dst_class=ObjDict)
```

And as ObjDict is a subclass of dict for objdict_to_dict we can use 

```python
return _nested_dict_to_dict(something, src_class=dict, dst_class=dict)
```

core/src/epicli/data/common/validation/configuration/helm.yml
seriva: Not sure what you want me to change, as the default configuration looks like this now:

```yaml
kind: configuration/helm
title: "Helm"
name: default
specification:
  apache_epirepo_path: "/var/www/html/epirepo" 
```

erzetpe (changes requested):

core/src/epicli/data/common/validation/configuration/helm.yml
atsikham: Whitespaces and newline, it's not correctly displayed here 

image

seriva: Nevermind, I see now;)

atsikham (dismissed):

atsikham (approved):

core/src/epicli/cli/engine/schema/SchemaValidator.py
seriva: Optimized.

erzetpe (approved):

Issue #805, opened 2019-12-30
[FEATURE REQUEST] Optimize downloading requirements introducing file checksums, use Python instead of Bash


to-bar:
**Is your feature request related to a problem? Please describe.**
1) Downloading all requirements takes about 12 min (Azure RHEL), sometimes longer.
2) Each time Epiphany downloads all requirements, there is no option to download only required ones - e.g. if there is no Ignite component enabled, its heavy binary is downloaded unnecessarily.
3) requirements.txt file doesn't contain files' checksums so we are not able to check if existing file was fully downloaded or it's corrupted. For this reason, files are re-downloaded each time (for RHEL) or may be corrupted (for Ubuntu).
4) requirements.txt file in current format doesn't allow to have flexible (structured) configuration (for checksums or dependencies).
5) We have redundant code so if you want to make a change in download-requirements.sh you may need to do it for every OS directory.

**Describe the solution you'd like**
1) Download any file only once. If exists, compare chechsums. If checksums differs, then re-download.
2) Switch from txt files with custom format to YAML.
3) Support 2 download modes: `full` (current solution) and `adaptive` (download only files required by enabled components).
4) Support multiple repositories for the same package.
5) Use the same code for all OS distribution when possible.
6) Support retries for network related tasks which may succeed on another try.
7) Ansible seems to meet the above requirements.

---

**DoD checklist**

- Changelog
  - [x] updated
  - [ ] not needed
- COMPONENTS.md
  - [ ] updated
  - [x] not needed
- Schema
  - [ ] updated
  - [x] not needed
- Backport tasks
  - [ ] created
  - [x] not needed
- Documentation
  - [ ] added
  - [ ] updated
  - [x] not needed
- [x] Feature has automated tests
- [x] Automated tests passed (QA pipelines)
  - [x] apply
  - [ ] upgrade
  - [ ] backup/restore
- [x] Idempotency tested
- [x] All conversations in PR resolved

rafzei:
I've check how messed this role is during #127...

We should keep consistency of the repository role for supported OS's. 

Proposed structure of files directory in repository role:

```
repository role
- files (dir)
  - client (dir)
    - common (dir contain common files with functions)
      - common_functions.sh (example file with functions for all OSs)
      - ...
    - debian.sh (this file should source common functions)
    - redhat.sh (same)
  - download-requirements (dir)
    - common (dir contain common files with functions)
      - common_functions.sh (example file with functions for all OSs)
    - centos-7
[...]

PR #2493, opened 2021-08-06
Add Internet connection test to download-requirements.sh (2124)

sbbroot:


erzetpe (dismissed):

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
...
+    err_msg=$(yum clean all) || exit_with_error "(yum) failed, reason: ($err_msg)"
+
+    err_msg=$(yum --quiet --disablerepo=* --enablerepo=$repo list available 2>&1 >/dev/null) \
+            || exit_with_error "(yum) failed, reason: ($err_msg)"
+}
+
+
+__check_crane() {
+#
+# TODO crane
```
rafzei: How about Crane? If we are going to use simple wget <URL> checking I would remove this section.

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/centos-7/download-requirements.sh
```diff
@@ -7,6 +7,11 @@
 
 set -euo pipefail
 
+# set variables needed by common_functions
+readonly CONNECTION_CHECK_ENABLED="yes"
```
rafzei: I personally like lowercase variables in the case of one shell scope.

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
...
+    [[ $# -lt 1 ]] && exit_with_error "(curl), no url provided"
+    local url=$1
+
+    echol "Testing connection to: ($url) using: (curl)..."
+
+    err_msg=$(curl -S --silent $url 2>&1 >/dev/null) || exit_with_error "(curl) failed, reason: ($err_msg)"
+}
+
+
+__check_wget() {
```
rafzei, line 93: Do we need __ prefix?

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/redhat-7/download-requirements.sh
```diff
@@ -653,6 +647,9 @@ else
         echol "Crane binary already exists"
 	else
 		file_url=$(head -n 1 <<< "${CRANE}")
+
+        check_connection wget $file_url
```
rafzei: Pay attention to indentation - you have used space instead of tab in this particular line. I see for this doc is set "Tab Size: 4"

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
...
+declare -A tools=(
+[curl]=__check_curl
+[wget]=__check_wget
+[apt]=__check_apt
+[yum]=__check_yum
+[crane]=__check_crane
+)
+
+
+check_connection() {
```
rafzei, line 211: I would add echo with a message at the beginning of this function

rafzei (commented):
Hmm.. I think we shouldn't even do this task before #805.
Anyways since it's already there @sbbroot please use [shellcheck extension for vscode](https://marketplace.visualstudio.com/items?itemName=timonwong.shellcheck). I see a bunch of underscores in your code. Please check them :v:

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/centos-7/download-requirements.sh
to-bar: I would use more explicit (narrower) name, e.g. `internet_access_checks_enabled`

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/centos-7/download-requirements.sh
```diff
@@ -599,6 +591,8 @@ get_requirements_from_group 'IMAGES'               'images'                "$REQ
 
 # --- Backup yum repositories ---
 
+check_connection yum
```
to-bar: The function is called just bellow the `--- Backup yum repositories ---` section (comment) but it tests connectivity what is not related to the backup. Please move it above the `=== Packages ===` section.

IMO, it would be nice to add new section, for example:
```
# --- Run internet access tests ---
check_connection yum
check_connection wget $first_url $second_url
```
For `$first_url` and `$second_url` we need any valid URLs (theoretically available files from two different domains) just to test if wget works (behind a proxy it may require extra settings in /etc/wgetrc or environment variables, such as `https_proxy`).
Why 2 URLs? The test should fail only if both files are not available since it's more likely that a single file may be not available (temporarily) on the remote server than two. When a file is not available, it doesn't necessarily mean a problem on the client side, thus I would suggest the double checking to avoid false/misleading results.

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/centos-7/download-requirements.sh
```diff
@@ -631,6 +625,9 @@ else
         echol "Crane binary already exists"
 	else
 		file_url=$(head -n 1 <<< "${CRANE}")
+
+        check_connection wget $file_url
```
to-bar: I would like to run all connection tests as prerequisites (close to the beginning, under the `--- Checks ---` section) in order to fail early if not met (to save time).

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
...
+    echol $@
+    exit 1
+}
+
+
+__check_curl() {
+#
+# Use curl in silent mode to check if target `url` is available.
+#
+# param $1: url which will be tested
```
to-bar:
```suggestion
# param $1: url to be tested
```

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
+echol() {
+#
+# Print to stdout, optionally to a log file.
+# Requires $CREATE_LOGFILE and $LOG_FILE_PATH to be defined.
+#
+# param $@: args which will be printed
```
to-bar:
```suggestion
# param $@: args to be printed
```

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
...
+}
+
+
+__check_curl() {
+#
+# Use curl in silent mode to check if target `url` is available.
+#
+# param $1: url which will be tested
+#
+    [[ $# -lt 1 ]] && exit_with_error "(curl), no url provided"
```
to-bar:
```suggestion
    (( $# > 0 )) || exit_with_error "__check_curl: no url provided"
```
IMO having real function name would be more informative.

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
...
+    local url=$1
+
+    echol "Testing connection to: ($url) using: (curl)..."
```
to-bar:
```suggestion
    echol "Testing connection to \"${url}\" using curl..."
```

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
...
+
+    err_msg=$(curl -S --silent $url 2>&1 >/dev/null) || exit_with_error "(curl) failed, reason: ($err_msg)"
```
to-bar:
```suggestion
    err_msg=$(curl --show-error --silent $url 2>&1 >/dev/null) || exit_with_error "curl failed, reason: ($err_msg)"
```
IMO in scripts it's better to use options in long form unless they are commonly known (intuitive/popular).

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
...
+}
+
+
+__check_wget() {
+#
+# Use wget in spider mode (without downloading resources) to check if target `url`
+# is available.
+#
+# param $1: url which will be tested
```
to-bar:
```suggestion
# param $1: url to be tested
```

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
...
+#
+    [[ $# -lt 1 ]] && exit_with_error "(wget), no url provided"
```
to-bar:
```suggestion
    (( $# > 0 )) || exit_with_error "__check_wget: no url provided"
```

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
to-bar: I guess this is to mark function as "private"/internal/helper - not called in external file.

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
...
+    local url=$1
+
+    echol "Testing connection to: ($url) using: (wget)..."
```
to-bar:
```suggestion
    echol "Testing connection to \"${url}\" using wget..."
```

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
...
+
+    err_msg=$(wget --spider -nd $url 2>&1 >/dev/null) || exit_with_error "(wget) failed, reason: ($err_msg)"
```
to-bar:
```suggestion
    err_msg=$(wget --spider --no-directories $url 2>&1 >/dev/null) || exit_with_error "wget failed, reason: ($err_msg)"
```
Do we need `--no-directories` option assuming that we will use only links to a single file?

I would prefer to return test result instead of exiting. In old Bash (as in RHEL 7) returning array is very limited but we could for example print array like: `printf '%s\n' "${array[@]}"` and have exit code in the first line - 0 - tests passed, 1 - failed. If failed, then have in next lines error message. Such approach would allow to run all tests in parent function then interpret results, for example print "report"/summary for all tools.

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
...
+}
+
+
+__check_apt() {
+#
+# Use `apt update` to make sure that there is connection to repositories.
```
to-bar, line 123: It looks like testing all repositories. IMO we may need to test repos one by one (in a loop) and finish testing (exit from the loop) on the first succeeded. I'm not sure about exit code when at least one repo is not available - but it looks like we may get false result using this command. We need to test whether **any** single repo is available instead of **all** of them.
This link may be helpful: https://askubuntu.com/questions/65245/apt-get-update-only-for-a-specific-repository

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
...
+#
+    echol "Testing (apt) connection..."
```
to-bar:
```suggestion
    echol "Testing apt connection..."
```

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
...
+
+    [[ $UID -ne 0 ]] && exit_with_error "(apt) needs to be run as a root"
```
to-bar:
```suggestion
    (( $UID == 0 )) || exit_with_error "apt needs to be run as root"
```

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
...
+
+    err_msg=$(apt update 2>&1 >/dev/null) || exit_with_error "(apt) failed, reason: ($err_msg)"
```
to-bar:
```suggestion
    err_msg=$(apt update 2>&1 >/dev/null) || exit_with_error "\"apt update\" failed, reason: ($err_msg)"
```

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
...
+}
+
+
+__check_yum() {
+#
+# Use `yum list` to make sure that there is connection to repositories.
+# Pick first available repo, clean the cache and then query available packages.
```
to-bar: Similar case to apt. Please see my comment there.
Probably we don't need to clean cache for all repos.
I would try something like this (in a loop):
`yum makecache fast --disablerepo="*" --enablerepo=<repo>`

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,122 @@
...
+    err_msg=$(yum --quiet --disablerepo=* --enablerepo=$repo list available 2>&1 >/dev/null) \
+            || exit_with_error "(yum) failed, reason: ($err_msg)"
+}
+
+
+__check_crane() {
+#
+# TODO crane
+#
+# param $1: url which will be tested
```
to-bar:
```suggestion
# param $1: repo/image to be tested
```
`repo` **or** `image` should be chosen depending on crane command (`crane ls` vs `crane pull`). I guess `crane ls` can be faster.

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
to-bar: @rafzei This check is to test `crane` (`crane ls` or `crane pull`), not `wget`.
Someone may need extra setup behind a proxy: https://github.com/google/go-containerregistry/issues/861

to-bar (changes requested):

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
sbbroot: All of those functions have proper echo messages, I'm not sure if another one is needed.

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
sbbroot: Crane ls will be used.

to-bar:
How about not exiting in a child function but in the parent? For example:

```
__check_curl() {
    (( $# > 0 )) || exit_with_error "__check_curl: no url provided"
    local url=$1

    echol "Testing connection to: \"${url}\" using curl..."

    err_msg=$(curl --show-error --silent "$url" 2>&1 >/dev/null); rc=$?
    if (( rc > 0)); then
        echol "curl failed, reason: [$err_msg], exit code: $rc"
    fi
    return $rc
}

check_connection() {
    if ${tools[$tool]} $@; then
        echol "PASSED"
    else
        # here we can decide what to do, e.g. run `exit_with_error` or return exit code from child function
        echol "FAILED"
    fi
}

check_connection curl https://github.fake
check_connection curl https://github.com

# output:

Testing connection to: "https://github.fake" using curl...
curl failed, reason: [curl: (6) Could not resolve host: github.fake], exit code: 6
FAILED
Testing connection to: "https://github.com" using curl...
PASSED
```

erzetpe (dismissed):

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/centos-7/download-requirements.sh
```diff
@@ -597,6 +589,8 @@ get_requirements_from_group 'IMAGES'               'images'                "$REQ
 
 # === Packages ===
 
+check_connection yum $(yum repolist --quiet | tail -n +2 | cut -d' ' -f1 | cut -d'/' -f1 | sed 's/!//')
```
to-bar:
```suggestion
check_connection yum $(yum repolist --quiet | tail -n +2 | cut -d' ' -f1 | cut -d'/' -f1 | sed 's/^!//')
```
yum repolist output's first column as ! if the repository has expired metadata.

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,238 @@
...
+    shift
+
+    local args=$@
+    local total_count=$#
+    local failed_count=0
+
+    for arg in $args; do
+        echol "- $arg..."
+        $function $arg
+        if (( $? != 0 )); then
```
to-bar: How about return on the first success to save time?

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,238 @@
...
+#
+# Run connection test for target `tool` with optional `url` parameter.
+# Requires $internet_access_checks_enabled to be defined.
+#
+# :param $1: which `tool` to test
+# :param $@: optional parameters used by some tools such as `url`
+#
+    [[ $internet_access_checks_enabled == "no" ]] && return 0
+
+    [[ $# -lt 1 ]] && exit_with_error "(tool) argument not provided"
```
to-bar:
```suggestion
    [[ $# -lt 1 ]] && exit_with_error '"tool" argument not provided'
```

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,238 @@
...
+    local tool=$1
+
+    shift  # discard tool variable
+
+    [[ ! -n ${tools[$tool]} ]] && exit_with_error "no such tool ($tool)"
```
to-bar:
```suggestion
    [[ ! -n ${tools[$tool]} ]] && exit_with_error "no such tool: \"$tool\""
```

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/common/common_functions.sh
```diff
@@ -0,0 +1,238 @@
...
+
+    (  # disable -e in order to handle non-zero return values
+        set +e
+
+        ${tools[$tool]} $@
+
+        if (( $? == 0 )); then
+            echol "Connection successful."
+        else
+            exit_with_error "Connection failure, reason: ($last_error)"
```
to-bar:
```suggestion
            exit_with_error "Connection failure, reason: [$last_error]"
```
() are probably used more often than [] in error message. 
IMO `[curl: (6) Could not resolve host: google.com; Name or service not known]` looks better than `(curl: (6) Could not resolve host: google.com; Name or service not known)`

core/src/epicli/data/common/ansible/playbooks/roles/repository/files/download-requirements/redhat-7/download-requirements.sh
```diff
@@ -616,6 +608,8 @@ get_requirements_from_group 'IMAGES'               'images'                "$REQ
 
 # --- Backup yum repositories ---
 
+check_connection yum $(yum repolist --quiet | tail -n +2 | cut -d' ' -f1 | cut -d'/' -f1 | sed 's/!//')
```
to-bar:
```suggestion
check_connection yum $(yum repolist --quiet | tail -n +2 | cut -d' ' -f1 | cut -d'/' -f1 | sed 's/^!//')
```

to-bar (changes requested):

to-bar (dismissed):

to-bar (approved):

erzetpe (approved):

PR #2500, opened 2021-08-18
Refactor upgrade role to not include "specification" at top level

to-bar:


core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/node-exporter.yml
```diff
@@ -106,8 +109,8 @@
       unarchive:
         remote_src: true
         src: "{{ download_directory }}/{{ exporter_file_name }}"
-        dest: "/opt/node_exporter"
-        creates: "/opt/node_exporter/node_exporter"
+        dest: /opt/node_exporter
+        creates: /opt/node_exporter/node_exporter
         extra_opts: [--strip-components=1]
         mode: '0755'
```
atsikham: What do you think about changing mode format in this PR?

to-bar: Applied

erzetpe (approved):

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/node-exporter.yml
```diff
@@ -44,6 +45,8 @@
           include_role:
             name: node_exporter
             tasks_from: install-node-exporter-as-daemonset.yml
+          vars:
+            specification: "{{ exporter_vars.specification }}"
```
atsikham, line 49: Don't we want to name it as `_specification`? This is a general question for all task vars we define.

atsikham (changes requested):

core/src/epicli/data/common/ansible/playbooks/roles/upgrade/tasks/node-exporter.yml
to-bar: Yes, we want but not in this PR. This would require additional changes (in `node_exporter` role) and tests for apply mode since `specification` is used there in many places and upgrade mode re-uses code from `node_exporter` role. I believe we can do that when changing `node_exporter` role.

atsikham (approved):

Issue #1351, opened 2020-06-15
Refactor Keycloak to work with Postgresql HA configuration


plirglo:
Currently keycloak is configured to work with only one database node (vm-0). We need to add setup to connect to database using ha tools (pgbouncer, pgpool) when HA is enabled.

mkyc:
Shouldn't we also have Keycloak itself in HA mode? Or is that in separate issue?

plirglo:
Keycloak is deployed in Kubernetes cluster, and yes it's replicable, but uses postgres database, and now kc configuration delivered by epiphany doesn't support HA postgres configuration. We might have problems because of this.

mkyc:
All clear now. I just noticed that kc uses ha configuration.

przemyslavic:
:heavy_check_mark: `Keycloak` deployment with 1 PostgreSQL node + `reapply`
:heavy_check_mark: `Keycloak` deployment with 2 clustered PostgreSQL nodes  + `reapply`
:heavy_check_mark: `Keycloak` deployment with 3 clustered PostgreSQL nodes  + `reapply`
:heavy_check_mark: `Keycloak` deployment with custom database config provided  + `reapply`

Tested both `develop` and `v1.2`.

PR #2501, opened 2021-08-23
Upgrade PostgreSQL to v13

atsikham:
Merging feature branch develop - contains all changes related to PostgreSQL epic.


to-bar (dismissed):

core/src/epicli/cli/epicli.py
```diff
@@ -290,6 +290,8 @@ def splitarg(arg):
     sub_parser._action_groups.append(optional)
 
     def run_upgrade(args):
+        if not query_yes_no('Has backup been done?', default='no'):
```
seriva, line 293: Maybe something more here to give some insight what/how to backup?

seriva (changes requested):
Are all validations for this properly updated? I see a bunch of defaults changed but only a small change in: 

core/src/epicli/data/common/validation/configuration/applications.yml

Lets at least make sure we keep basic type checking for required fields in order moving forward.

przemyslavic:
I encountered issues with determining the master during running `re-apply` command. Easily reproduced in Azure CentOS environment. The build was successful, however I ran into a problem during testing.
Inventory after re-applying:
```
[postgresql]
ci-devazurcentflannel-postgresql-vm-0 ansible_host=x
ci-devazurcentflannel-postgresql-vm-1 ansible_host=y
```
Repmgr shows that the primary node is `ci-devazurcentflannel-postgresql-vm-1`.
```
[postgres@ci-devazurcentflannel-postgresql-vm-0 ~]$ repmgr cluster show
 ID | Name                                  | Role    | Status    | Upstream                              | Location | Priority | Timeline | Connection string
----+---------------------------------------+---------+-----------+---------------------------------------+----------+----------+----------+-------------------------------------------------------------------
 1  | ci-devazurcentflannel-postgresql-vm-1 | primary | * running |                                       | default  | 100      | 1        | host=10.1.6.4 user=epi_repmgr dbname=epi_repmgr connect_timeout=2
 2  | ci-devazurcentflannel-postgresql-vm-0 | standby |   running | ci-devazurcentflannel-postgresql-vm-1 | default  | 100      | 1        | host=10.1.6.5 user=epi_repmgr dbname=epi_repmgr connect_timeout=2
 
```
In many places it is set that primary is `groups.postgresql[0]` which should be `ci-devazurcentflannel-postgresql-vm-0` according to the inventory file. It looks like the order changed on the second run.
This causes problems starting keycloak application trying to access `10.1.6.5` host that actually points to a replica.
```
   Environment:
      KEYCLOAK_USER:                 auth-service-username
      KEYCLOAK_PASSWORD:             <set to the key 'password' in secret 'as-testauthdb-http'>  Optional: false
      JGROUPS_DISCOVERY_PROTOCOL:    dns.DNS_PING
      JGROUPS_DISCOVERY_PROPERTIES:  dns_query=as-testauthdb-headless.namespace-for-auth.svc.cluster.local
      DB_VENDOR:                     postgres
      DB_ADDR:                       10.1.6.5
      DB_PORT:                       5432
```
```
13:40:30,732 WARN  [org.keycloak.connections.jpa.updater.liquibase.lock.CustomLockService] (ServerService Thread Pool -- 70) Lock didn't yet acquired. Will possibly retry to acquire lock. Details: ERROR: cannot execute SELECT FOR UPDATE in a read-only transaction [Failed SQL: SELECT ID FROM public.databasechangeloglock WHERE ID=1000 FOR UPDATE]
13:40:30,747 WARN  [org.keycloak.connections.jpa.updater.liquibase.lock.CustomLockService] (ServerService Thread Pool -- 70) Lock didn't yet acquired. Will possibly retry to acquire lock. Details: ERROR: cannot execute SELECT FOR UPDATE in a read-only transaction [Failed SQL: SELECT ID FROM public.databasechangeloglock WHERE ID=1000 FOR UPDATE]
```

plirglo:
>... 
> In many places it is set that primary is `groups.postgresql[0]` which should be `ci-devazurcentflannel-postgresql-vm-0` according to the inventory file. It looks like the order changed on the second run.
> This causes problems starting keycloak application trying to access `10.1.6.5` host that actually points to a replica.
> 
> ```
>    Environment:
>       KEYCLOAK_USER:                 auth-service-username
>       KEYCLOAK_PASSWORD:             <set to the key 'password' in secret 'as-testauthdb-http'>  Optional: false
>       JGROUPS_DISCOVERY_PROTOCOL:    dns.DNS_PING
>       JGROUPS_DISCOVERY_PROPERTIES:  dns_query=as-testauthdb-headless.namespace-for-auth.svc.cluster.local
>       DB_VENDOR:                     postgres
>       DB_ADDR:                       10.1.6.5
>       DB_PORT:                       5432
> ```
> 
> ```
> 13:40:30,732 WARN  [org.keycloak.connections.jpa.updater.liquibase.lock.CustomLockService] (ServerService Thread Pool -- 70) Lock didn't yet acquired. Will possibly retry to acquire lock. Details: ERROR: cannot execute SELECT FOR UPDATE in a read-only transaction [Failed SQL: SELECT ID FROM public.databasechangeloglock WHERE ID=1000 FOR UPDATE]
> 13:40:30,747 WARN  [org.keycloak.connections.jpa.updater.liquibase.lock.CustomLockService] (ServerService Thread Pool -- 70) Lock didn't yet acquired. Will possibly retry to acquire lock. Details: ERROR: cannot execute SELECT FOR UPDATE in a read-only transaction [Failed SQL: SELECT ID FROM public.databasechangeloglock WHERE ID=1000 FOR UPDATE]
> ```

I think it's connected to this issue and should be fixed with keycloak
https://github.com/epiphany-platform/epiphany/issues/1351

przemyslavic:
@plirglo agree. The issue is that we assume is some places ([playbooks, templates, backup and recovery tasks](https://github.com/epiphany-platform/epiphany/search?q=groups%5B%27postgresql%27%5D%5B0%5D)) that the first node from inventory is a master (which is not always true, especially after re-applying). Examples:
```
    when:
      - groups['postgresql'][0] == inventory_hostname
```
```
  delegate_to: "{{ groups['postgresql'] | first }}"
```
```
    auth_service_db_address: "{{ data.database.address | default(hostvars[groups.postgresql[0]].ansible_default_ipv4.address) }}"
```

core/src/epicli/cli/epicli.py
atsikham: Do you suggest something?

to-bar: This prompt is for all components (not only for PostgreSQL) and backup can be done in different way than using `epicli backup`, thus the prompt is so generic. For PostgreSQL, so far we don't remove anything (old data dir is preserved) so a backup may be even not required.
Does `Has backup of components to be upgraded been done?` is a better choice?

to-bar:
> 
> 
> Are all validations for this properly updated? I see a bunch of defaults changed but only a small change in:
> 
> core/src/epicli/data/common/validation/configuration/applications.yml
> 
> Lets at least make sure we keep basic type checking for required fields in order moving forward.

@seriva Good point. I've checked and added `PGPOOL_CHILD_LIFE_TIME` property.

core/src/epicli/cli/epicli.py
seriva: Gotcha, then I am ok with this.

seriva (approved):

to-bar (approved):

Issue #1294, opened 2020-05-25
Implement proper merging of lists of dictionaries for epicli yaml docs


sk4zuzu:
Epiphany's default yaml merger does not take into account that any document can be a mix of dictionaries and lists. There are two issues with that:
1. When dictionary is defined inside a list, then it's not properly converted to internal dictionary implementation `ObjDict`.
2. Merging user config with default config (in case of lists of dictionaries present inside) does not set default values at all, it means that full config structure must be always provided by the user.

sk4zuzu:
> 2. Merging user config with default config (in case of lists of dictionaries present inside) does not set default values at all, it means that full config structure must be always provided by the user.

This could be implemented for structures like:
```yaml
  available_roles:
    - name: repository
      enabled: yes
    - name: firewall
      enabled: yes
    - name: image-registry
      enabled: yes
    - name: kubernetes-master
      enabled: yes
    - name: kubernetes-node
      enabled: yes
```
which is basically a dictionary implemented with lists, where the key is the value of then `name:`.

There is a existing counter-example though:
```yaml
  supported_os:
    - name: RedHat
      comparison_operator: "=="
      version: 7.6
    - name: RedHat
      comparison_operator: "=="
      version: 7.7
    - name: RedHat
      comparison_operator: "=="
      version: 7.8
    - name: RedHat
      comparison_operator: "=="
      version: 7.9
    - name: CentOS
      comparison_operator: "=="
      version: 7.6
    - name: CentOS
```
both of these structures cannot be megred in the same way.

IMO, in the end, it seems to be better to leave the default behavior of replacing lists (like for example docker-compose does) and convert existing dictionary-like structures (implemented with lists) to proper dictionaries.

sk4zuzu:
This issue needs to be reconsidered. We cannot fix this problem by introducing another way to merge yaml docs. The correct approach is to fix yaml document schema, but this is not a backward-compatible operation.

mkyc:
@sk4zuzu can you please confirm that this is still valid task?

seriva:
Yes its a problem, no I dont think we should spend time on fixing this in legacy Epiphany. @toszo @sk4zuzu @mkyc what do you think? Close it?

sk4zuzu:
If the general consensus would be that we focus on new things, then considering this is not so easy to implement it in a generic way and we have been living without it since he beginning of time (:D), we can close it IMO.

seriva:
[...]

PR #2504, opened 2021-08-24
Introduced a _merge parameter for merging list

seriva:
Introduced a _merge (true|false) parameter for merging list based on the comments [here](https://github.com/epiphany-platform/epiphany/issues/1294#issuecomment-900194372)

In short this allows for adding a directive on how named list should be processed, or **merge** or **overwrite**, where **overwrite** is the default when nothing is defined. For example:

```yaml
kind: configuration/applications
title: "Kubernetes Applications Config"
name: default
provider: azure
specification:
  applications:
  - _merge: true
  - name: auth-service # requires PostgreSQL to be installed in cluster
    enabled: true
```

This will merge the defined **auth-service** entry with the **defaults** definition.


docs/home/HOWTO.md
```diff
@@ -135,3 +135,6 @@
     - [Downloading requirements progression and logging](./howto/REPOSITORY.md#downloading-requirements-progression-and-logging)
     - [Re-downloading requirements](./howto/REPOSITORY.md#re-downloading-requirements)
     - [Restoring system repositories](./howto/REPOSITORY.md#restoring-system-repositories)
+
+- [Schema](./howto/SCHEMA.md)
```
to-bar: IMO `Schema` may be too generic. It's rather related to validations. How about `Configuration file` or something like that?
And `Elasticsearch` below looks like a mistake 😉

docs/home/howto/SCHEMA.md
```diff
@@ -0,0 +1,42 @@
+## Schema
```
to-bar:
```suggestion
## Inheritance
```

to-bar: We could add info similar to [this one](https://docs.rubocop.org/rubocop/configuration.html#inheritance) - unless we already have in another doc.

docs/home/howto/SCHEMA.md
```diff
@@ -0,0 +1,42 @@
...
+
+### Named lists
+
+Epiphany uses a concept called **named lists** in the configuration YAML. Every entry in a **named lists**  list will have a ```name``` field to identify it and make it unique for merge operations:
```
to-bar:
```suggestion
Epiphany uses a concept called **named lists** in the configuration YAML. Every item in a **named list**  has the ```name``` key to identify it and make it unique for merge operation:
```

docs/home/howto/SCHEMA.md
```diff
@@ -0,0 +1,42 @@
...
+  list:
+  - name: entry1
+    property1: property1
+    property2: property2
+  - name: entry2
+    property1: property1
+    property2: property2
+  - name: entry3
+    property1: property1
+    property2: property2
```
to-bar:
```suggestion
  - name: item1
    property1: value1
    property2: value2
  - name: item2
    property1: value3
    property2: value4
```

docs/home/howto/SCHEMA.md
```diff
@@ -0,0 +1,42 @@
...
+...
+```
+
+By default, a **named list** inside your configuration it will completely overwrite the defaults that Epiphany provides. This behaviour is on purpose so when you, for example, define a list of users for Kafka inside your configuration it completely overwrite the users defined in the [Kafka defaults.](https://github.com/epiphany-platform/epiphany/blob/9ff9bb266cd6addf309059a8a5e7a48835daafc3/core/src/epicli/data/common/defaults/configuration/kafka.yml#L34).
```
to-bar:
```suggestion
By default, a **named list** in your configuration file will completely overwrite the defaults that Epiphany provides. This behaviour is on purpose so when you, for example, define a list of users for Kafka inside your configuration it completely overwrites the users defined in the [Kafka defaults](https://github.com/epiphany-platform/epiphany/blob/9ff9bb266cd6addf309059a8a5e7a48835daafc3/core/src/epicli/data/common/defaults/configuration/kafka.yml#L34).
```

docs/home/howto/SCHEMA.md
```diff
@@ -0,0 +1,42 @@
...
+
+In some cases however you don't want to overwrite a **named list**. A good example would be the [application configurations.](https://github.com/epiphany-platform/epiphany/blob/develop/core/src/epicli/data/common/defaults/configuration/applications.yml)
```
to-bar:
```suggestion
In some cases, however, you don't want to overwrite a **named list**. A good example would be the [application configurations](https://github.com/epiphany-platform/epiphany/blob/v1.0.1/core/src/epicli/data/common/defaults/configuration/applications.yml).
```
IMO it's better to avoid linking to develop branch since things are changing there 😉

docs/home/howto/SCHEMA.md
```diff
@@ -0,0 +1,42 @@
...
+
+You don't want to re-define every entry just to make sure Epiphany has all default entries needed by the Ansible automation. That is where the ```_merge``` metadata tag comes in. It will let you define if you want to ```overwrite``` or ```merge``` sa **named list** by setting it to ```true``` or ```false```.
```
to-bar:
```suggestion
You don't want to re-define every item just to make sure Epiphany has all default items needed by the Ansible automation. That is where the ```_merge``` metadata tag comes in. It will let you define whether you want to ```overwrite``` or ```merge``` a **named list** by setting it to ```true``` or ```false```.
```

docs/home/howto/SCHEMA.md
```diff
@@ -0,0 +1,42 @@
...
+name: default
+provider: azure
+specification:
+  applications:
+  - _merge: true
+  - name: auth-service
+    enabled: true
+```
+
+The ```_merge``` entry with ```true``` will tell Epicli to merge the application list and only change the ```enabled: true``` entry inside the ```auth-service``` and take the rests of the [configuration/applications]((https://github.com/epiphany-platform/epiphany/blob/develop/core/src/epicli/data/common/defaults/configuration/applications.yml)) configuration from the defaults.
```
to-bar:
```suggestion
The ```_merge``` item with ```true``` will tell epicli to merge the application list and only change the ```enabled: true``` setting inside the ```auth-service``` and take the rests of the [configuration/applications]((https://github.com/epiphany-platform/epiphany/blob/v1.0.1/core/src/epicli/data/common/defaults/configuration/applications.yml)) configuration from the defaults.
```

to-bar (changes requested):

docs/home/howto/SCHEMA.md
seriva: Good point

docs/home/HOWTO.md
seriva: Yeah agreed, Configuration file is better choice.

to-bar (dismissed):

erzetpe (dismissed):

rafzei (dismissed):

to-bar (dismissed):

rafzei (approved):

przemyslavic (approved):

PR #2518, opened 2021-08-26
Set up default value for child_life_time parameter

plirglo:
Set up default value for child_life_time parameter. Default = 300s (5 minutes)


erzetpe (approved):

erzetpe (changes requested):
Please fix git history

plirglo:
> Please fix git history

It will be squashed anyway.

PR #2543, opened 2021-09-02
Fix/log path issue

seriva:
- Fixed error in epicli if no 'path' is provided. Causes issue when running commands like: epicli --help
- Also added alias for epicli to .bashrc so you run epicli easily without debugger in the devcontainer

### DIFF
--- a/core/src/epicli/.devcontainer/Dockerfile
+++ b/core/src/epicli/.devcontainer/Dockerfile
@@ -66,4 +66,4 @@ RUN : SETUP USER, CERTS AND OTHERS \
  && setcap 'cap_net_bind_service=+ep' /usr/bin/ssh \
  && /bin/bash /config-post.sh
 
-RUN echo alias epicli='"cd /workspaces/epiphany/core/src/epicli && PYTHONPATH=$(pwd) python3 cli/epicli.py"' >> /etc/bash.bashrc
+RUN echo alias epicli='"PYTHONPATH=/workspaces/epiphany/core/src/epicli && python3 -m cli.epicli"' >> /etc/bash.bashrc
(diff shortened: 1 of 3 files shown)


core/src/epicli/.devcontainer/Dockerfile
```diff
@@ -65,3 +65,5 @@ RUN : SETUP USER, CERTS AND OTHERS \
  && chmod ug=r,o= /etc/sudoers.d/$USERNAME \
  && setcap 'cap_net_bind_service=+ep' /usr/bin/ssh \
  && /bin/bash /config-post.sh
+
+RUN echo alias epicli='"cd /workspaces/epiphany/core/src/epicli && PYTHONPATH=$(pwd) python3 cli/epicli.py"' >> /etc/bash.bashrc
```
rafzei: Its fine, but this could be even better:
alias epicli='PYTHONPATH=$PYTHONPATH:/workspaces/epiphany/core/src/epicli/; python3 /workspaces/epiphany/core/src/epicli/cli/epicli.py'
1- do not override global env pythonpath,
2- do not change user pwd (can be annoying)

seriva: As you could see after some testing, pwd needs to be /workspaces/epiphany/core/src/epicli. Also the PYTHONPATH is only for that session so should not matter.

rafzei: alias epicli="export PYTHONPATH=/workspaces/epiphany/core/src/epicli && python3 -m cli.epicli"
should do the job, can you verify?

to-bar: Or alternative approach (if we want to run `epicli` also as root):
```
cat /etc/profile.d/epicli.sh
export PYTHONPATH=${PYTHONPATH}:/workspaces/epiphany/core/src/epicli
alias epicli='python3 -m cli.epicli'
```
If we want only for vscode user, then we may create `~/.bash_aliases`

Btw maybe also worth to add `alias ll='ls -alF'` 😉

seriva: That fixed the issue. For now lets leave as is.

rafzei (approved):

plirglo (approved):

przemyslavic (approved):

PR #2552, opened 2021-09-06
Added interpretation of alternative yaml boolean values as true booleans

seriva:
This was already partially fixed in a previous issue. In this pull we add all acceptable values for Yes/No:

https://yaml.org/type/bool.html

Note that we don't add on/off etc as we need these as strings in for example the Postgres role. Also added unit tests for the functionality.


rafzei (dismissed):

erzetpe (approved):

mkyc:
Please agree with @przemyslavic that it's ok to merge.

rafzei (approved):

przemyslavic (approved):

PR #2555, opened 2021-09-07
Added ansible-lint and rubocop to the devcontainer

seriva:
- Added ansible-lint to the devcontainer
- Added rubocop to the devcontainer
- Added ansible-lint task to devcontainer to produce junit output
- Added rubocop task to devcontainer to produce junit output
- Moved running of Pytest and Pylint to tasks instead of debug configuration

**Notes:**  
- Running ansible-lint is pretty slow, running over all playbooks takes about half an hour. I checked and this seems to be a known issue which is being worked on by the creators. Thi is the reason it does not play will with showing results in the problems tab of VSCode for open  source files and only work when running it as a task.
- Rubocop works fine as tasks or from the problems tab:
image

**Task running:**
To run tasks to **ctrl+shift+p** and search for run tasks

image

Select that option and then select the task you want to run:

image


rafzei (approved):

erzetpe (approved):

seriva:
Merging as this does not require QA testing from a product perspective.

Issue #2497, opened 2021-08-11
[BUG] epicli init --full -p azure generates not correct region value


plirglo:
**Describe the bug**
Epicli apply --full -p azure generates not correct region value - it generates eu-west-2 which is value for AWS.

**How to reproduce**
1. execute `epicli init --full -p azure
2. see config file
3. execution with default region value will fail

**Expected behavior**
Region value should be compatible with Azure if we generate file for Azure

**Environment**
- Cloud provider: Azure
- OS: all

**epicli version**: all

**Additional context**
Probably default value in file https://github.com/epiphany-platform/epiphany/blob/develop/core/src/epicli/data/common/defaults/epiphany-cluster.yml should be changed.

---

**DoD checklist**

* [x] Changelog updated (if affected version was released)
* [x] COMPONENTS.md updated / doesn't need to be updated
* [x] Automated tests passed (QA pipelines)
  * [ ] apply
  * [ ] upgrade
* [ ] Case covered by automated test (if possible)
* [ ] Idempotency tested
* [x] Documentation updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Backport tasks created / doesn't need to be backported

sbbroot:
Bug is in this function:
```python
def load_all_yaml_objs(file_type, provider, kind):
    script_dir = os.path.dirname(__file__)
    path_to_file = os.path.join(script_dir, DATA_FOLDER_PATH, provider, file_type, kind+'.yml')
    return load_file_from_path(script_dir, path_to_file, file_type, kind)
```

__path_to_file__ will always have a value of:
```"epiphany/core/src/epicli/cli/helpers/../../data/azure/defaults/epiphany-cluster.yml"```
such file does not exist.

Therefore in __load_file_from_path__ function __else__ statement will always be executed.
```python
def load_file_from_path(script_dir, path_to_file, file_type, kind):
    if os.path.isfile(path_to_file):
        return load_yamls_file(path_to_file)
    else:
        path_to_file = os.path.join(script_dir, DATA_FOLDER_PATH, 'common', file_type, kind + '.yml')
        return load_yamls_file(path_to_file)
```
This leads to getting values from the:
```"epiphany/core/src/epicli/cli/helpers/../../data/common/defaults/epiphany-cluster.yml"```
which is the default file for all providers that contains __aws__ values.

przemyslavic:
PR #2652 is still in progress so this is definitely not ready for testing.
[...]

PR #2599, opened 2021-09-21
Fix epicli apply --full region values (#2497)

sbbroot:
* Add separate epiphany-cluster.yml for Azure

* Move epiphany-cluster.yml from common to aws


plirglo:
Please add task to changelog.

rafzei (dismissed):

rafzei (dismissed):

atsikham (changes requested):
_Point for discussion:_ shouldn't we consider having still a common configuration and merging that with a specific one for each cloud provider?

seriva:
> _Point for discussion:_ shouldn't we consider having still a common configuration and merging that with a specific one for each cloud provider?

Issue would be that this would require some different processes just for the cluster-config. So after discussion we decided to have dedicated document for each provider as we already have with the minimal cluster config. Only thing I think that is missing is the default config for ```any``` provider.

seriva (commented):
- Does the any provider has a dedicated document?
- Did you check if their where any changes needed in merging code when doing apply etc?

core/src/epicli/data/any/defaults/epiphany-cluster.yml
```diff
@@ -0,0 +1,71 @@
...
+title: "Epiphany cluster Config"
+provider: any
+name: "default"
+specification:
+  prefix: default
+  name: epiphanycluster
+  admin_user:
+    name: operations # YOUR-ADMIN-USERNAME
+    key_path: /root/.ssh/epiphany-operations/id_rsa # YOUR-SSH-KEY-PATH
+  cloud:
```
seriva: Cloud can be removed.

core/src/epicli/data/azure/defaults/epiphany-cluster.yml
```diff
@@ -0,0 +1,96 @@
...
+    name: operations # YOUR-ADMIN-USERNAME
+    key_path: /root/.ssh/epiphany-operations/id_rsa # YOUR-SSH-KEY-PATH
+  cloud:
+    k8s_as_cloud_service: False
+    subscription_name: YOUR-SUB-NAME
+    vnet_address_pool: 10.1.0.0/20
+    use_public_ips: False # When not using public IPs you have to provide connectivity via private IPs (VPN)
+    use_service_principal: False
+    region: West Europe
+    credentials: # todo change it to get credentials from vault
```
seriva: credentials is AWS only I think.

core/src/epicli/data/azure/defaults/epiphany-cluster.yml
```diff
@@ -0,0 +1,96 @@
...
+  name: epiphanycluster
+  admin_user:
+    name: operations # YOUR-ADMIN-USERNAME
+    key_path: /root/.ssh/epiphany-operations/id_rsa # YOUR-SSH-KEY-PATH
+  cloud:
+    k8s_as_cloud_service: False
+    subscription_name: YOUR-SUB-NAME
+    vnet_address_pool: 10.1.0.0/20
+    use_public_ips: False # When not using public IPs you have to provide connectivity via private IPs (VPN)
+    use_service_principal: False
```
seriva, line 16: use_service_principal is Azure only so it can be removed from the AWS config if present.

seriva (commented):
Also make sure that the count for each component are the same across all providers.

core/src/epicli/data/aws/defaults/epiphany-cluster.yml
```diff
@@ -13,7 +13,6 @@ specification:
     subscription_name: YOUR-SUB-NAME
```
seriva: subscription_name is also for Azure only so can be removed.

seriva (dismissed):

atsikham (dismissed):

rafzei (dismissed):

seriva (dismissed):

seriva:
Merged as per @przemyslavic  permission as we want todo #2643 ASAP. He will test it from develop.

seriva (approved):

plirglo (approved):

PR #2601, opened 2021-09-21
Logging output changes

seriva:
Previous:

- Stream (what you see in the console): Can be outputted in plain text with color or JSON output. Can be controlled with the ```--log-type plain|json``` flag.
- File: Only outputs in JSON and is unaffected by the ```--log-type plain|json``` flag.

New:
- Stream (what you see in the console):  Only outputs in plain text with color  now and is unaffected by the ```--log-type plain|json``` flag.
- File: Outputs in plain text by default now, but can be set to json with ```--log-type plain|json``` flag.

Output format was left unchanged but can be set by user with: ```--log-format```


atsikham (dismissed):

erzetpe (dismissed):

rafzei (dismissed):
Works pretty sweet now. 
One thing I've noticed is even you provide epicli --log-format the _Stream_ remains in old format. But I can live with that :)

erzetpe (dismissed):

seriva:
> Works pretty sweet now.
> One thing I've noticed is even you provide epicli --log-format the _Stream_ remains in old format. But I can live with that :)

I cant;) Pushed a fix for that.

rafzei (approved):
Tested, now is perfect :star2:

erzetpe (approved):

Issue #2573, opened 2021-09-10
[BUG] [PostgreSQL] Upgrade fails when there is user-defined tablespace


to-bar:
**Describe the bug**
Upgrade fails when there is user-defined tablespace:
```
TASK [postgresql : PostgreSQL | Extensions | repmgr | Clone standby node with repmgr] ********************************************************************************
[...]

PR #2614, opened 2021-09-22
Fixed upgrade with user defined tablespaces

atsikham:
Task #2573. Changelog is not updated as PostgreSQL upgrade hasn't been released yet.

### DIFF
--- a/core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/helpers/rm-dir-content.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/helpers/rm-dir-content.yml
@@ -1,0 +1,16 @@
+---
+- name: PostgreSQL | Get {{ directory_path }} directory listing
+  find:
+    path: "{{ directory_path }}"
+    file_type: any
+    hidden: true
+  register: directory_content
+
+- name: PostgreSQL | Remove {{ directory_path }} directory content
+  file:
+    path: "{{ file.path }}"
+    state: absent
+  loop: "{{ directory_content.files }}"
+  loop_control:
+    loop_var: file
+    label: "{{ file.path }}"
--- a/core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/helpers/rm-dir-content.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/helpers/rm-dir-content.yml
@@ -2,6 +2,7 @@
 - name: PostgreSQL | Get {{ directory_path }} directory listing
   find:
     path: "{{ directory_path }}"
+    patterns: "{{ pattern }}"
     file_type: any
     hidden: true
   register: directory_content
--- a/core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/upgrade/upgrade.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/upgrade/upgrade.yml
@@ -35,6 +35,7 @@
   set_fact:
     pg_primary_node: "{{ pg_primary_node_found.stdout | default('') }}"
 
+# postgresql_info module doesn't display tablespace locations, so it's requested
 - name: PostgreSQL | Query custom tablespaces
   become_user: postgres
   postgresql_query:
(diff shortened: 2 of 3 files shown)


core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/helpers/rm-dir-content.yml
```diff
@@ -0,0 +1,17 @@
...
+  register: directory_content
+
+- name: PostgreSQL | Remove {{ directory_path }} directory content
+  file:
+    path: "{{ file.path }}"
+    state: absent
+  loop: "{{ directory_content.files }}"
+  loop_control:
+    loop_var: file
+    label: "{{ file.path }}"
```
atsikham, line 17: https://docs.ansible.com/ansible/latest/user_guide/playbooks_loops.html#limiting-loop-output-with-label

seriva (dismissed):

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/upgrade/upgrade.yml
```diff
@@ -35,6 +35,27 @@
   set_fact:
     pg_primary_node: "{{ pg_primary_node_found.stdout | default('') }}"
 
+- name: PostgreSQL | Query custom tablespaces
```
to-bar, line 39: How about using `postgresql_info` module?
Please check https://docs.ansible.com/ansible/latest/collections/community/postgresql/postgresql_info_module.html#return-tablespaces

to-bar (changes requested):

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/upgrade/upgrade.yml
atsikham: It was my thought to do so initially, but this module doesn't display tablespaces' filesystem location and it should be requested somehow.

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/upgrade/upgrade.yml
```diff
@@ -35,6 +35,27 @@
...
+
+- name: PostgreSQL | Set 'pg_custom_tablespace_locations' fact
+  block:
+    - name: PostgreSQL | Initialize 'pg_custom_tablespace_locations' fact
+      set_fact:
+        pg_custom_tablespace_locations: [ ]
+
+    - name: PostgreSQL | Set 'pg_custom_tablespace_locations' fact
+      set_fact:
+        pg_custom_tablespace_locations: "{{ pg_custom_tablespace_locations + [pg_tablespace.location] }}"
```
to-bar: You can consider combining these two tasks into one using  `default([])`:
```suggestion
        pg_custom_tablespace_locations: "{{ pg_custom_tablespace_locations | default([]) + [pg_tablespace.location] }}"
```

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/upgrade/upgrade.yml
to-bar: Oh, didn't know, sorry then :flushed:
You can consider adding  comment about that.

core/src/epicli/data/common/ansible/playbooks/roles/postgresql/tasks/upgrade/upgrade.yml
atsikham: If the query result contains 0 elements, this fact will be undefined and this will require to add a default in a few other places

to-bar: Then `pg_custom_tablespaces.query_result | map(attribute='location') | reject('==', '')` is an option instead of `loop`.

atsikham: modified

to-bar (approved):

przemyslavic (approved):

Issue #1306, opened 2020-05-29
[FEATURE REQUEST] Allow to check on VMs which epicli version was used to deploy/upgrade components


rafzei:
**Is your feature request related to a problem? Please describe.**
Right now we can recognize the version of epicli by specific k8s version on masters/nodes.

**Describe the solution you'd like**
We want to dump the version number of executed epicli on cluster components (VMs) to easily recognize which version is deployed. We can also think about feature to add a history of upgrades if they were any. 

**Describe alternatives you've considered**
None.

**Additional context**
None.

---

**DoD checklist**

* [x] Changelog updated
* [x] COMPONENTS.md updated / doesn't need to be updated
* [x] Schema updated / doesn't need to be updated
* [ ] Feature has automated tests
* [x] Automated tests passed (QA pipelines)
  * [x] apply
  * [x] upgrade
* [x] Idempotency tested
* [ ] Documentation added / updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Solution meets requirements and is done according to design doc
* [x] Usage compliant with license
* [x] Backport tasks created / doesn't need to be backported

mkyc:
@rafzei do you mean that: 

```
*****EPICLI VERSION******
0.6.0
```

?
Those lines are currently at the top of generated dump.

rafzei:
@mkyc I mean 'target' VMs not a epicli one. I've edit the issue description - I hope it has more sense right now.

to-bar:
Today I was asked about such functionality so it's really needed.

atsikham:
Optional, but with that feature we can skip following step in `data/common/ansible/playbooks/roles/image_registry/tasks/load-image.yml` if there is an upgrade to the same version:

```yaml
- name: Load legacy version images to registry when upgrading
   include_tasks: load-image.yml
   vars:
     docker_image: "{{ item }}"
   loop: "{{ specification.images_to_load.legacy }}"
   when: is_upgrade_run
```

mkyc:
Such information is presented in MOTD now (see #2628)

atsikham:
> Such information is presented in MOTD now (see #2628)

Right, but only during login and there is no any upgrade history that @rafzei mentioned in description.

rafzei:
Yup, I think adding some ~/.version or ~/.epicli file containing that information would be a good solution here.

.epicli:
```
# type - date - version
Apply - 11:30:00 27-09-2021 - 1.3.0dev
```

[...]

Issue #2643, opened 2021-09-29
[FEATURE REQUEST] Restructure Epicli project folder


seriva:
**Is your feature request related to a problem? Please describe.**
Currently the core of the project "Epicli" lives in /core/src/epicli with the idea that this would extended with different additional projects. However as we now know we will transition to modules down the line. This setup does not make it very intuitive to work with as the devcontainer needs to be opened in /core/src/epicli and for documentation an extra windows needs to be opened.

**Describe the solution you'd like**
Because of this we want to restructure the project:
- Move the content of /core/src/epicli to the root of the project so that we can open the project in one single workspace/devcontainer. The structure of what is inside  /core/src/epicli will be left in tact for now.
- Merge /core/src/epicli/.gitignore into /.gitignore
- Fix paths for docker build inside:
   -  /core/src/epicli/setup.py
   - /.dockerignore
   - /Dockerfile
- Remove old certificate code in /core/src/epicli/.devcontainer/cert. This has not been maintained/tested properly when COVID started.
- Move CHANGELOG-x.x.md files to a subfolder inside docs and update CHANGELOG.md paths to these version
- Update /docs/home/DEVELOPMENT.md
- Add markdownlint plugin to devcontainer
- Adjust build pipelines for development branch

**Describe alternatives you've considered**
None

**Additional context**
https://docs.python-guide.org/writing/structure/

---

**DoD checklist**

* [ ] Changelog updated
* [x] COMPONENTS.md updated / doesn't need to be updated
* [x] Schema updated / doesn't need to be updated
* [ ] Feature has automated tests
* [ ] Automated tests passed (QA pipelines)
  * [x] apply
  * [ ] upgrade
* [ ] Idempotency tested
* [x] Documentation added / updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Solution meets requirements and is done according to design doc
* [x] Usage compliant with license
* [x] Backport tasks created / doesn't need to be backported

przemyslavic:
Tested together with #2666 (part 2).

PR #2650, opened 2021-10-01
Add epicli versioning (#1306)

sbbroot:
* Add version info file creation for each VM used.


seriva:
First we need to merge #2643.

core/src/epicli/data/common/ansible/playbooks/roles/versioning/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
+- name: Create versioning directory if it does not exist.
```
atsikham:
```suggestion
- name: Create versioning directory if it does not exist
```

core/src/epicli/data/common/ansible/playbooks/roles/versioning/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+  file:
+    path: /etc/epicli
+    state: directory
+    mode: '0755'
```
atsikham:
```suggestion
    mode: u=rwx,go=rx
```

core/src/epicli/data/common/ansible/playbooks/roles/versioning/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+
+- name: Get versioning file status.
```
atsikham:
```suggestion
- name: Get versioning file status
```

core/src/epicli/data/common/ansible/playbooks/roles/versioning/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+  stat:
+    path: /etc/epicli/version
+  register: versioning_file
+
+- name: Create versioning file.
```
atsikham:
```suggestion
- name: Create versioning file
```

core/src/epicli/data/common/ansible/playbooks/roles/versioning/tasks/main.yml
```diff
@@ -0,0 +1,22 @@
...
+  copy:
+    dest: /etc/epicli/version
+    content: |
+      #mode | date | version
+  when: not versioning_file.stat.exists
+
+- name: Append versioning info.
```
atsikham:
```suggestion
- name: Append versioning info
```

core/src/epicli/data/common/ansible/playbooks/roles/versioning/tasks/main.yml
atsikham: Has it been checked by linter?

core/src/epicli/cli/engine/ansible/AnsibleRunner.py
```diff
@@ -152,6 +161,12 @@ def upgrade(self):
         # pre-flight to prepare machines
         self.pre_flight(inventory_path)
 
+        self.ansible_command.run_playbook(inventory=inventory_path,
+                                          playbook_path=self.playbook_path(to_role_name('versioning')),
+                                          extra_vars={'mode': 'upgrade',
+                                                      'date': datetime.datetime.now().ctime(),
+                                                      'version': VERSION})
```
atsikham: There is `is_upgrade_run` fact and `ansible_date_time` can be used from ansible. Is it necessary to pass these parameters and modify `run_playbook ` for that?

atsikham (changes requested):

core/src/epicli/cli/engine/ansible/AnsibleRunner.py
to-bar: Yes, there can be also time differences, remote time (on VM) `ansible_date_time` can be different than local time (on Ansible controller).

to-bar (changes requested):
IMO:
1) We may no need separate playbook. `postflight` playbook can be used if we want to log only successful runs.
2) We can use parseable file format, for example YAML (Ansible has `from_yaml` and `to_nice_yaml` filters).
Sample structure:
```
# ordered list of successful epicli runs - the latest at the top
deployments:
  - version: 1.2.0
    date: 2021-08-17 15:24
    mode: apply
    state: started
  - version: 1.2.0
    date: 2021-08-17 15:24
    mode: upgrade
    state: completed
  - version: 1.1.0
    date: 2021-05-17 15:24
    mode: apply
    state: completed
```
3) `/etc/epicli/version` should be replaced with something like `/var/lib/epiphany/history.yml`. `\etc` is mainly intended for configuration files, see [here](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s07.html).
4) If we implement deployment history instead of just the latest version, then role can be renamed to e.g. `deployment_history`.

atsikham:
What about adding the status for each run? We can add info in preflight about started run and update state in postflight. This way if there is a failure and partial installation is done, we will have an information about that.

core/src/epicli/cli/engine/ansible/AnsibleRunner.py
sbbroot: I do not agree with the `is_upgrade_run` change. If we want to add something later we will probably end up with a bunch of `if` `else`s. Here the script doesn't care about which mode it is, so there is no place for conditionals.

sbbroot: About `ansible_date_time`, maybe we should add both local and remote time?

atsikham: We don't need to implement `is_upgrade_run`, it's already in place and used in Ansible configuration.

sbbroot: I meant that it's a bad design, and if there is a better way of doing it, we should do it. Scripts should not be aware of the mode they are in, this leads to putting conditionals everywhere. In our python code we have distinct paths for both apply and upgrade modes.

atsikham: Ok, before talking about design, let’s change the code not to use a separate role as @to-bar suggested. Then we will see what’s necessary in addition.

cli/engine/ansible/AnsibleRunner.py
```diff
@@ -1,5 +1,7 @@
+import datetime
 import os
 import time
+import shutil
```
atsikham: Is this change necessary without changes in code?

ansible/playbooks/postflight.yml
```diff
@@ -5,3 +5,18 @@
   become_method: sudo
   roles:
     - postflight
+
+
```
atsikham: Remove one empty line.

ansible/playbooks/postflight.yml
```diff
@@ -5,3 +5,18 @@
...
+- name: Ansible playbook for changing the state of the current running version info
```
atsikham: - It's not necessary to mention that this is Ansible playbook.
- `current running version` - looks not good? Whoose version?

May be something like, just a suggestion: `Change the state of the previous Epicli execution`

ansible/playbooks/postflight.yml
```diff
@@ -5,3 +5,18 @@
...
+  hosts: all
+  gather_facts: yes
```
atsikham:
```suggestion
  gather_facts: true
```

ansible/playbooks/postflight.yml
```diff
@@ -5,3 +5,18 @@
...
+  become: true
+  become_method: sudo
+  tasks:
+    - name: Change current running status to complete
+      lineinfile:
+        path: /var/lib/epiphany/history.yml
+        firstmatch: yes  # only first occurrence
```
atsikham:
```suggestion
        firstmatch: true  # only first occurrence
```

ansible/playbooks/postflight.yml
```diff
@@ -5,3 +5,18 @@
...
+        backrefs: yes  # allows to use regexp groups
```
atsikham:
```suggestion
        backrefs: true  # allows to use regexp groups
```

ansible/playbooks/postflight.yml
atsikham: May be there a situation when such update will change not the last status? Is there a way tostart  search from the end?

ansible/playbooks/preflight.yml
```diff
@@ -11,3 +11,50 @@
     - preflight
   environment:
     KUBECONFIG: "{{ kubeconfig.remote }}"
+
+
+- name: Ansible playbook for adding epicli version info for each component used
```
atsikham: Same as previous.

ansible/playbooks/preflight.yml
```diff
@@ -11,3 +11,50 @@
...
+  hosts: all
+  gather_facts: yes
```
atsikham: Change all `yes` values to `true`

cli/engine/ansible/AnsibleVarsGenerator.py
```diff
@@ -142,6 +142,7 @@ def populate_group_vars(self, ansible_dir):
         main_vars['is_upgrade_run'] = self.is_upgrade_run
         main_vars['roles_with_generated_vars'] = sorted(self.roles_with_generated_vars)
         main_vars['upgrade_components'] = Config().upgrade_components
+        main_vars['version'] = VERSION
```
atsikham: Looks like not required thing - there is a version in generated defaults. @to-bar what do you think?

atsikham (dismissed):

ansible/playbooks/postflight.yml
sbbroot: I don't think that there is a possibility that such thing could happen.
Preflight task always puts newest execution on top.

cli/engine/ansible/AnsibleVarsGenerator.py
sbbroot: Version value is not visible at preflight stage without this change.

to-bar: For me this is ok since VERSION is the single source of true.  But I would change `main_vars['version']` to `main_vars['epiphany_version']` for two reasons: 1) more intuitive, 2) lower risk that it will be overridden for example by `set_fact` task.

ansible/playbooks/postflight.yml
```diff
@@ -5,3 +5,17 @@
   become_method: sudo
   roles:
     - postflight
+
+- name: Change the state info of the previous epicli execution
+  hosts: all
+  gather_facts: true
```
to-bar:
```suggestion
  gather_facts: false
```
It looks like facts are not used here but gathering them takes time.

ansible/playbooks/postflight.yml
```diff
@@ -5,3 +5,17 @@
...
+  become: true
+  become_method: sudo
+  tasks:
+    - name: Change current running status to complete
```
to-bar: You can consider more robust (but longer) approach:
1) Load file using `slurp` module
2) Parse yaml using Ansible filter (`from_yaml`)
3) Update value using Ansible filter(s) - probably `combine`
4) Save updated contents to file (using `copy` module)
Example: https://github.com/epiphany-platform/epiphany/blob/7e7987ab6d3dfd047040c487b4d4a1c230887ba4/ansible/playbooks/roles/kubernetes_common/tasks/extend-kubeadm-config.yml#L19

ansible/playbooks/preflight.yml
```diff
@@ -11,3 +11,50 @@
...
+      lineinfile:
+        path: /var/lib/epiphany/history.yml
+        firstmatch: true  # only first occurrence
+        backrefs: true  # allows to use regexp groups
+        regexp: '(.*)state\: started'  # assuming that current status is always on top
+        line: '\1state: failed'
+      when: history_file.stat.exists and history_file.stat.size > 0
+
+    - name: Add history info
+      lineinfile:
```
to-bar: Operations on lists - like prepend are easier to understand than using regexp that replaces single line string with multi-line.

ansible/playbooks/preflight.yml
```diff
@@ -11,3 +11,50 @@
...
+        path: /var/lib/epiphany/history.yml
+        regexp: 'deployments:'
+        line: "{{ history_file_content }}"
+      when: history_file.stat.exists and history_file.stat.size > 0
+
+    - name: Create history file
```
to-bar: Without `lineinfile` you would have full content stored in variable (`set_fact`) and the when condition will be unneeded.

to-bar (changes requested):

ansible/playbooks/postflight.yml
to-bar: This would make sense if we update not only the `state` but the `date` as well.

ansible/playbooks/postflight.yml
sbbroot: I find it overengineered. I don't see robustness through putting more code (and possibly more bugs).

ansible/playbooks/preflight.yml
sbbroot: Current solution uses only single variable for each task. Changing it would make it more complex in my opinion.

ansible/playbooks/postflight.yml
to-bar: My opinion is that files that have a standardized format (like yml, json, xml) should be modified using parsers. And Ansible has built-in utils to work with YAML. So I will not approve regexp based solution.

erzetpe (dismissed):

erzetpe (dismissed):

ansible/playbooks/preflight.yml
```diff
@@ -11,3 +11,57 @@
     - preflight
   environment:
     KUBECONFIG: "{{ kubeconfig.remote }}"
+
+- name: Add epicli version info for each component used
```
to-bar:
```suggestion
- name: Add entry to epicli execution history on each host
```

ansible/playbooks/preflight.yml
```diff
@@ -11,3 +11,57 @@
...
+  become: true
+  become_method: sudo
+  vars:
+    history_last_entry: {}
+    history_other_entries: []
+    history_newest_entry:
+        version: "{{ epiphany_version }}"
+        date: "{{ ansible_date_time.date }} {{ ansible_date_time.time }}"
+        mode: "{{ is_upgrade_run | ternary('upgrade', 'apply') }}"
+        state: started
```
to-bar:
```suggestion
      version: "{{ epiphany_version }}"
      date: "{{ ansible_date_time.date }} {{ ansible_date_time.time }}"
      mode: "{{ is_upgrade_run | ternary('upgrade', 'apply') }}"
      status: started
```

ansible/playbooks/preflight.yml
```diff
@@ -11,3 +11,57 @@
...
+  tasks:
+    - name: Create epiphany directory if it does not exist
+      file:
+        path: /var/lib/epiphany/
```
to-bar:
```suggestion
        path: /var/lib/epiphany
```

ansible/playbooks/preflight.yml
```diff
@@ -11,3 +11,57 @@
...
+      register: history_file_handler
+      when: history_file.stat.exists and history_file.stat.size > 0
+
+    - name: Parse history_file_handler content to yaml format
+      set_fact:
+        history_last_entry: "{{ history_file_yaml.deployments[0] }}"
+        history_other_entries: "{{ history_file_yaml.deployments[1:] }}"
+      vars:
+        history_file_yaml: "{{ history_file_handler.content | b64decode | from_yaml }}"
+      when: history_file.stat.exists and history_file.stat.size > 0
```
to-bar:
```suggestion
    - name: Parse history file
      when:
        - history_file.stat.exists
        - history_file.stat.size > 0
      block:
        - name: Load history file
          slurp:
            src: /var/lib/epiphany/history.yml
          register: history_file_handler

        - name: Parse history file
          set_fact:
            history_latest_entry: "{{ _history_file_yaml.deployments[0] }}"
            history_other_entries: "{{ _history_file_yaml.deployments[1:] }}"
          vars:
            _history_file_yaml: "{{ history_file_handler.content | b64decode | from_yaml }}"
```

ansible/playbooks/preflight.yml
```diff
@@ -11,3 +11,57 @@
...
+
+    - name: Construct new history file
+      set_fact:
+        new_history_file: >-
+          {{ {'deployments': [history_newest_entry] + history_last_entry | ternary([history_last_entry], []) + history_other_entries} }}
+
+    - name: Save the history file content
+      copy:
+        content: "{{ new_history_file | to_nice_yaml }}"
+        dest: /var/lib/epiphany/history.yml
```
to-bar:
```suggestion
    - name: Check previous status and update if needed
      set_fact:
        history_latest_entry: "{{ history_latest_entry | combine({'status': 'failed'}) }}"
      when:
        - history_latest_entry.status is defined
        - history_latest_entry.status != 'completed'

    - name: Update history and save to file
      copy:
        content: "{{ _new_content | to_nice_yaml(indent=2) }}"
        dest: /var/lib/epiphany/history.yml
      vars:
        _new_content:
          deployments: >-
            {{ [history_new_entry] + history_latest_entry | ternary([history_latest_entry], []) + history_other_entries }}
```

ansible/playbooks/postflight.yml
```diff
@@ -5,3 +5,33 @@
   become_method: sudo
   roles:
     - postflight
+
+- name: Change the state info of the previous epicli execution
```
to-bar:
```suggestion
- name: Update status in epicli execution history file
```

ansible/playbooks/postflight.yml
```diff
@@ -5,3 +5,33 @@
...
+  gather_facts: false
+  become: true
+  become_method: sudo
+  tasks:
+    - name: Load the history file
+      slurp:
+        src: /var/lib/epiphany/history.yml
+      register: history_file
+
+    - name: Parse history_file to yaml format and change the state
```
to-bar:
```suggestion
    - name: Parse history file and update the status
```

ansible/playbooks/postflight.yml
```diff
@@ -5,3 +5,33 @@
...
+
+    - name: Construct history file
+      set_fact:
+        new_history_file: >-
+          {{ {'deployments': [history_first_entry] + history_other_entries} }}
+
+    - name: Save newly constructed history file
+      copy:
+        content: "{{ new_history_file | to_nice_yaml }}"
+        dest: /var/lib/epiphany/history.yml
```
to-bar:
```suggestion
    - name: Save updated history to file
      copy:
        content: "{{ _new_content | to_nice_yaml(indent=2) }}"
        dest: /var/lib/epiphany/history.yml
      vars:
        _new_content:
          deployments: "{{ [history_latest_entry] + history_other_entries }}"
```

ansible/playbooks/postflight.yml
```diff
@@ -5,3 +5,33 @@
...
+  become_method: sudo
+  tasks:
+    - name: Load the history file
+      slurp:
+        src: /var/lib/epiphany/history.yml
+      register: history_file
+
+    - name: Parse history_file to yaml format and change the state
+      set_fact:
+        history_first_entry: >-
```
to-bar:
```suggestion
        history_latest_entry: >-
```

ansible/playbooks/preflight.yml
```diff
@@ -11,3 +11,57 @@
...
   environment:
     KUBECONFIG: "{{ kubeconfig.remote }}"
+
+- name: Add epicli version info for each component used
+  hosts: all
+  gather_facts: true
+  become: true
+  become_method: sudo
+  vars:
+    history_last_entry: {}
```
to-bar:
```suggestion
    history_latest_entry: {}
```

ansible/playbooks/preflight.yml
```diff
@@ -11,3 +11,57 @@
...
+    history_other_entries: []
+    history_newest_entry:
```
to-bar:
```suggestion
    history_new_entry:
```

ansible/playbooks/postflight.yml
```diff
@@ -5,3 +5,33 @@
...
+          {{ history_file_yaml.deployments[0] | combine({'state': 'completed'})}}
```
to-bar:
```suggestion
          {{ history_file_yaml.deployments[0] | combine({'status': 'completed'})}}
```

to-bar (changes requested):

to-bar (approved):

erzetpe (approved):

przemyslavic (approved):

PR #2652, opened 2021-10-01
Fix upgrade mode (#2497) (#2599)

sbbroot:
* Fix bug introduced in 1f9fb7fe

### DIFF
--- a/core/src/epicli/cli/engine/ansible/AnsibleInventoryUpgrade.py
+++ b/core/src/epicli/cli/engine/ansible/AnsibleInventoryUpgrade.py
@@ -104,7 +104,7 @@ class AnsibleInventoryUpgrade(Step):
             self.cluster_model = select_single(self.manifest_docs, lambda x: x.kind == 'epiphany-cluster')
 
         # Merge manifest cluster config with newer defaults
-        default_cluster_model = load_yaml_obj(data_types.DEFAULT, 'common', 'epiphany-cluster')
+        default_cluster_model = load_yaml_obj(data_types.DEFAULT, 'any', 'epiphany-cluster')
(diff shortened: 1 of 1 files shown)


seriva (approved):

core/src/epicli/cli/engine/ansible/AnsibleInventoryUpgrade.py
```diff
@@ -104,7 +104,7 @@ def upgrade(self):
             self.cluster_model = select_single(self.manifest_docs, lambda x: x.kind == 'epiphany-cluster')
 
         # Merge manifest cluster config with newer defaults
-        default_cluster_model = load_yaml_obj(data_types.DEFAULT, 'common', 'epiphany-cluster')
+        default_cluster_model = load_yaml_obj(data_types.DEFAULT, 'any', 'epiphany-cluster')
```
atsikham, line 107: epiphany-cluster.yml is located in different folders. Why `any` is used in this place? Should usage of other providers such as `aws` and `azure` be considered?

seriva: @atsikham Good point. @sbbroot You probably want to use the self.cluster_model.provider here. Should be present during upgrade. This would allow proper merging of any, azure or aws providers now as we work with separate models for each now after the previous fixes you did.

przemyslavic:
The previous PR (#2599, already merged) introduced a bug. Please see my comment https://github.com/epiphany-platform/epiphany/issues/2497#issuecomment-938599380.

seriva:
Closing this as I added the fix to #2677 to avoid issues with merging later on.

PR #2708, opened 2021-10-14
Refactor PostgreSQL upgrade

to-bar:

Issue #2656, opened 2021-10-01
[BUG] [PostgreSQL] WAL files are not removed from $PGDATA/pg_wal directory


to-bar:
**Describe the bug**
Before version 1.2.0 (PostgreSQL 13) we had set `archive_command` parameter to `'test ! -f /dbbackup/{{ inventory_hostname }}/backup/%f && gzip -c < %p > /dbbackup/{{ inventory_hostname }}/backup/%f'`.

Till version 1.1.0 this was OK since we had the task for creating target directory for WAL files to be archived:
```
# Placed here to avoid errors in log due to nonexistent directory
- name: Create directory for archiving WAL files
  file:
    path: /dbbackup/{{ inventory_hostname }}/backup/
    state: directory
    owner: postgres
    group: postgres
```

In version 1.1.0 the above task has been removed but without changing the `archive_command`.

There are two consequences of such mismatched configuration:
1) WAL files are not removed because they are not considered as archived.
Example after importing a database with size of 13 GB:
```
ls -la /var/lib/postgresql/10/main/pg_wal | wc -l

780

du -sh /var/lib/postgresql/10/main/pg_wal

13G
```

With correct configuration it would be:
```
ls -la /var/lib/postgresql/10/main/pg_wal | wc -l

69

du -sh /var/lib/postgresql/10/main/pg_wal

1.1G
```

2) A lot of warnings in PostgreSQL log file (~1/min), for example:
```
grep -c "archive command failed with exit code 2" /var/log/postgresql/postgresql-10-main.log
891
```

This bug was found during testing PostgreSQL upgrade after importing ~ 6 GB dump file.

**How to reproduce**
1. Deploy PostgreSQL cluster using epicli 1.1.0
2.  Create new database, e.g. `psql template1 -c 'create database test'`
3. Import roles, e.g. `psql test < /tmp/cicd17-autotest17-202108191758UTC.pgdumpall-roles.sql`
4. Import data, e.g. `psql test < /tmp/cicd17-autotest17-202108191758UTC.pgdump.sql`
5. Run `sudo find /var/lib/postgresql/10/main/pg_wal -maxdepth 1 -type f | wc -l`

**Expected behavior**
~ 65 files (or less) in /var/lib/postgresql/10/main/pg_wal/

```
$ repmgr node check | grep WAL
        WAL archiving: OK (0 pending archive ready files)
```

**Config files**
Defaults with replication enabled.

**Environment**
- Cloud provider: All
[...]

PR #2709, opened 2021-10-15
Fixed wal files removal

atsikham:
Temporary workaround for those who will upgrade from 1.1 version. Task #2656.
Depends on #2708.


ansible/playbooks/roles/postgresql/tasks/preflight/upgrade.yml
```diff
@@ -28,6 +28,27 @@
   set_fact:
     pg_upgrade_already_run: "{{ stat_file_in_pg_new_data_dir.stat.exists }}"
 
+# This task is a temporary workaround for issue #2656
+# Can be removed when Epiphany v1.1 is out of support
+- name: Check if there is broken archive command from Epiphany v1.1
```
to-bar, line 47: This block requires running PostgreSQL service. In PR #2708 I added skipping such checks on rerun since after a failure service might be down. I suggest to merge PR #2708 first, then adjust this check to run conditionally i.e. `when not pg_upgrade_failed`.

rafzei (approved):

erzetpe (approved):

Issue #2703, opened 2021-10-13
[FEATURE REQUEST] Add tests for docker and kubelet cgroup driver


atsikham:
**Is your feature request related to a problem? Please describe.**
Epicli sets `systemd` cgroup driver for docker and kubelet after fix #1988. But it's not covered by tests.

**Describe the solution you'd like**
Add tests for kubelet and docker cgroup driver configuration.

**Describe alternatives you've considered**

-

**Additional context**
https://github.com/epiphany-platform/epiphany/blob/develop/tests/spec/spec/kubernetes_master/kubernetes_master_spec.rb

---

**DoD checklist**

* [x] Changelog updated
* [x] COMPONENTS.md updated / doesn't need to be updated
* [x] Schema updated / doesn't need to be updated
* [ ] Feature has automated tests
* [ ] Automated tests passed (QA pipelines)
  * [x] apply
  * [x] upgrade
* [x] Idempotency tested
* [x] Documentation added / updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Solution meets requirements and is done according to design doc
* [x] Usage compliant with license
* [x] Backport tasks created / doesn't need to be backported

PR #2696, opened 2021-10-08
Kubernetes v1.19 upgrade

atsikham:
Tasks #2093 and #1587.
Check COMPONENTS.md for the list of updated components.

According to [release notes](https://v1-19.docs.kubernetes.io/docs/setup/release/notes/) there are changes for deprecated parts:
* https://github.com/kubernetes/kubernetes/pull/92610
* https://github.com/kubernetes/kubernetes/pull/89778
* https://github.com/kubernetes/kubernetes/pull/90513

PR #2711, opened 2021-10-15
Add tests for docker and kubelet cgroup driver

przemyslavic:
Task #2703.
Useful for testing Kubernetes upgrade #2696.

### DIFF
--- a/tests/spec/spec/kubernetes_master/kubernetes_master_spec.rb
+++ b/tests/spec/spec/kubernetes_master/kubernetes_master_spec.rb
@@ -1,223 +1,235 @@
 require 'spec_helper'
 require 'base64'
 
-kube_apiserver_secure_port = 6443
+# kube_apiserver_secure_port = 6443
 
-describe 'Waiting for all pods to be ready' do
-  describe command("for i in {1..1200}; do if [ $(kubectl get pods --all-namespaces -o json  | jq -r '.items[] | select(.status.phase != \"Running\" or ([ .status.conditions[] | select(.type == \"Ready\" and .status != \"True\") ] | length ) == 1 ) | .metadata.namespace + \"/\" + .metadata.name' | wc -l) -eq 0 ]; \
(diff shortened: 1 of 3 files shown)


to-bar (dismissed):

to-bar:
@przemyslavic You can consider adding this task to changelog :slightly_smiling_face:

przemyslavic:
Done ☺️

tests/spec/spec/kubernetes_master/kubernetes_master_spec.rb
```diff
@@ -186,7 +185,39 @@
     describe command('kubectl describe nodes | grep -i taints') do
       its(:stdout) { should match /<none>/ }
       its(:exit_status) { should eq 0 }
-    end    
+    end
   end
 
 end
+
+describe 'Check the container runtime cgroup driver' do
```
atsikham: In our case container runtime is Docker, and here different thing is checked, so I would change description.

tests/spec/spec/kubernetes_node/kubernetes_node_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'spec_helper'
+
+describe 'Check the container runtime cgroup driver' do
```
atsikham: Same here.

atsikham (changes requested):

atsikham:
In our [documentation](https://github.com/epiphany-platform/epiphany/blob/develop/docs/home/CERTIFICATES.md) we have such statement:

```
kubelet.conf is not renewed because kubelet is configured for automatic certificate renewal. To verify that, navigate to /var/lib/kubelet/ and check config.yaml file, where rotateCertificates setting is true by default.
```

We could check `rotateCertificates` here the same way as `cgroupDriver` (in ConfigMap and `/var/lib/kubelet/config.yaml`).

WDYT about adding it here without an additional task?

atsikham (approved):
Great to see this done so quick!

erzetpe (approved):

PR #2714, opened 2021-10-18
Truncate motd file on upgrade

atsikham:
Task #2494

### DIFF
--- a/ansible/playbooks/roles/common/tasks/main.yml
+++ b/ansible/playbooks/roles/common/tasks/main.yml
@@ -81,6 +81,14 @@
   changed_when: false
   when: ansible_selinux is defined and ansible_selinux != False and ansible_selinux.status == 'enabled'
 
+# Before version 1.3 Epiphany updated /etc/motd file, so this task is necessary for upgrades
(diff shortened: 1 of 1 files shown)


przemyslavic (approved):

przemyslavic:
Tested `apply` and `upgrade` on `Ubuntu`, `RHEL` and `CentOS`.
On a cluster upgraded from v1.1.0 to dev it shows `Epiphany version is 1.1.0`, is that correct? I assume it is because this is an upgrade, and not re-apply, but I want to make sure 😉

atsikham:
@przemyslavic I think it should be discussed as I also don't know requirements, but I would change that during upgrade 😄

przemyslavic:
@atsikham After discussion, we agreed that it would be good to have information about the original version that was installed and the version to which it was updated.

erzetpe (approved):

atsikham:
> @atsikham After discussion, we agreed that it would be good to have information about the original version that was installed and the version to which it was updated.

It was discussed internally that what's is displayed there needs to be discussed after #1306 (PR #2650) is completed.

Issue #1661, opened 2020-09-16
[BUG] Changing the default Kubernetes certificate location results in a cluster deployment error


przemyslavic:
**Describe the bug**
Changing the default Kubernetes certificate location results in a cluster deployment error - the installations fails trying to initialize the cluster while the kubelet service is not running.

**To Reproduce**
Steps to reproduce the bug:
1. Deploy a new cluster from develop branch with the configuration given below (changing the default certificate location `/etc/kubernetes/pki` to a different one):  `epicli apply -f test.yml`

**Config files**

```yaml
---
kind: configuration/kubernetes-master
title: "Kubernetes Master Config"
name: default
provider: azure
specification:
  advanced:
    certificates:
      location: /etc/kube-certs/pki
      expiration_days: 365
      renew: false
```

**Expected behavior**
The cluster has been deployed successfully.

**OS (please complete the following information):**
 - OS: [all]

**Cloud Environment (please complete the following information):**
 - Cloud Provider [all]

**Additional context**
Log:
```
2020-09-15T12:54:22.1590201Z    [38;21m12:54:22 INFO cli.engine.ansible.AnsibleCommand - TASK [kubernetes_master : Initialize Kubernetes cluster] ***********************    [0m
2020-09-15T12:58:27.4272444Z    [31;21m12:58:27 ERROR cli.engine.ansible.AnsibleCommand - fatal: [ec2-xx-xx-xx-xx.eu-west-1.compute.amazonaws.com]: FAILED
```
```
This error is likely caused by:\n\t\t- The kubelet is not running\n\t\t- The kubelet is unhealthy due to a misconfiguration of the node in some way (required cgroups disabled)\n\n\tIf you are on a systemd-powered system, you can try to troubleshoot the error with the following commands:\n\t\t- 'systemctl status kubelet'\n\t\t- 'journalctl -xeu kubelet'
```
The problem is that the path `/etc/kubernetes/pki` is hardcoded in many places and therefore we are not able to change the certificate locations in one place at the moment. In my opinion, we should control this parameter in one place.

przemyslavic:
:heavy_check_mark: `apply`
:heavy_check_mark: `re-apply`
:heavy_check_mark: `apply` HA cluster
:heavy_check_mark: `upgrade` from v 1.0.1 to 1.3.0dev

PR #2717, opened 2021-10-20
Moved K8s pki location to Ansible defaults

atsikham:
Task #1661.
Why it's moved to defaults instead of supporting specifying parameter:
- It previously hasn't been working so nobody needs this property as a parameter
- Backup relates to `/etc/kubernetes/pki` value


erzetpe (dismissed):

ansible/playbooks/roles/kubernetes_common/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+pki:
+  location: /etc/kubernetes/pki
```
rafzei: Can we use it as source for pki.location in kubernetes_master role?

to-bar: We do not have defaults in kubernetes_common. So it would required to include the new file in master and node roles. So IMO we can keep current approach since it's simpler 😉

rafzei (approved):

to-bar (approved):

Issue #1076, opened 2020-03-26
Add sorting entries in the inventory file


przemyslavic:
**Describe the bug**
Now for different clusters we have virtual machines in a disordered way.

**Expected behavior**
There should be always the same order of the hostnames based on for example timestamp of creation data.

**Config files**
```
[kubernetes_master]
ci-haazurrhelcanal-kubernetes-master-vm-2 ansible_host=
ci-haazurrhelcanal-kubernetes-master-vm-0 ansible_host=
ci-haazurrhelcanal-kubernetes-master-vm-1 ansible_host=

[kubernetes_master]
ci-haazurubucanal-kubernetes-master-vm-0 ansible_host=
ci-haazurubucanal-kubernetes-master-vm-1 ansible_host=
ci-haazurubucanal-kubernetes-master-vm-2 ansible_host=

[kubernetes_master]
ci-haazurubuflannel-kubernetes-master-vm-2 ansible_host=
ci-haazurubuflannel-kubernetes-master-vm-1 ansible_host=
ci-haazurubuflannel-kubernetes-master-vm-0 ansible_host=

```
**Provider:** 

Azure, AWS

mkyc:
Check if that is related to #1706

rpudlowski93:
Yes, it looks that the PR is releated to the task. Moreover as seriva metioned here : https://github.com/epiphany-platform/epiphany/issues/1702, we should removed ASG for AWS. So in my opinion the task can be connected to #1706, and a new task should be created for removing the ASG for AWS.

przemyslavic:
:heavy_check_mark: `apply` + `reapply` for `azure` and `any` providers
:heavy_check_mark:  `upgrade` from v1.0 to v1.3.0dev + `apply` v1.3.0dev
After executing the `apply` command, the hosts are sorted in inventory by hostname.
Tested all available components with clustering enabled (where possible).

IMO #1706 should be closed now.

PR #2733, opened 2021-10-29
Add sorting entries in the inventory file (#1076)

sbbroot:
* Add sorting hostnames for `Azure` and `Any` provider in the inventory
  file

### DIFF
--- a/tests/unit/engine/providers/data/APIProxy_data.py
+++ b/tests/unit/engine/providers/data/APIProxy_data.py
@@ -1,0 +1,193 @@
+from typing import Dict, List
+
+from cli.helpers.ObjDict import ObjDict
+from cli.helpers.objdict_helpers import dict_to_objdict
+
+
+def CONFIG_DOC() -> ObjDict:
+    return dict_to_objdict([
+        {
+            'kind': 'infrastructure/machine',
+            'title': 'Virtual Machine Infra',
+            'provider': 'any',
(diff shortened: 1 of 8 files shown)


tests/unit/engine/providers/data/APIProxy_data.py
```diff
@@ -0,0 +1,193 @@
...
+                                 'service-2',
+                                 'service-3',
+                                 'service-4']
+                }
+            }
+        },
+        'version': '1.3.0dev'
+    })
+
+RUNNING_INSTANCES: List[List[Dict]] = [
```
seriva: This would be specific to Azure, maybe name it something like RUNNING_INSTANCES_AZURE?

seriva (approved):
Looks good. Don't care for the TestCase wrapping but that's partially preferences ;)

seriva (changes requested):
Missed the use of unittest in from your imports as PawoG noted. We choose pytest over unittest so pls rewrite these tests in pytests as I dont think we should be using 2 testing frameworks mixed.

erzetpe (dismissed):

erzetpe (dismissed):

seriva (dismissed):

przemyslavic (dismissed):

przemyslavic (approved):

erzetpe (approved):

PR #2762, opened 2021-11-29
Implement new download requirements script (#805)

sbbroot:
* rsync added to the Dockerfile


seriva (commented):
- Maybe there was a different discussion with @to-bar  but my understanding was we wanted to do this in Ansible?
- IF we are using Python all code is compatible with Python 2.7>? Yes 2 is deprecated but RedHat is still using 2.7 that I think.
- rsync would also be added to the devcontainer.
- Dependency list would be nice to have in YAML and images and files might be shared across distros as they are the same. Platform would still have seperation between AMD/ARM.

sbbroot:
> * Maybe there was a different discussion with @to-bar  but my understanding was we wanted to do this in Ansible?
> * IF we are using Python all code is compatible with Python 2.7>? Yes 2 is deprecated but RedHat is still using 2.7 that I think.
> * rsync would also be added to the devcontainer.
> * Dependency list would be nice to have in YAML and images and files might be shared across distros as they are the same. Platform would still have seperation between AMD/ARM.

* I did research on my own and found out that Ansible is not suitable for such feature. Python solution is shorter, faster and more elastic than Ansible approach. I also considered Ansible modules but they were designed for other purposes, this is more of a standalone script. In my opinion you should look at the code and consider if it's a fine code and we're ok with it, or maybe there is something wrong and we cannot go on with this kind of approach.
* No need for python 2.7, both RedHat and Centos support Python 3.6+
* Thanks, I've added rsync to the devcontainer
* My fault, I did not look at those files and thought that each distro has it's own set of images and files. No need for YAML since python modules are very flexible and in this case much faster than loading and parsing YAML. I came out with a simple solution to share common files and images.

to-bar:
> I did research on my own and found out that Ansible is not suitable for such feature. Python solution is shorter, faster and more elastic than Ansible approach. I also considered Ansible modules but they were designed for other purposes, this is more of a standalone script. In my opinion you should look at the code and consider if it's a fine code and we're ok with it, or maybe there is something wrong and we cannot go on with this kind of approach.

Didn't research but for me Ansible solution seems to be shorter since it provides ready to use modules like `get_url`, `apt`, `yum`, `docker_image` (requires Docker daemon), and `shell`. Downloading requirements probably could be implemented by calling those in loops (for example iterate over list of URLs).  For tools like crane we could write our own module (if not exists in Ansible Galaxy). 

> No need for python 2.7, both RedHat and Centos support Python 3.6+
RHEL/CentOS 7 out of the box has only Python 2 but it is possible to install Python 3. So this will be prerequisite. Open question whether it should be automated on our side or provided (installed) by user on his own.

IMO the main goal of this task is not just get rid of Bash but speed up downloading on rerun. To achieve this we may need to rely on checksums in order to detect corrupted files (downloaded only partially) to re-download them. Are you going to add checksums?

>> Dependency list would be nice to have in YAML

Agree.  It's a good practice to store data/configuration in a format that can be parsed by other tools - for example in case we decide to download something using Ansible.

seriva:
I think for now we can close this pull as this is something we will tackle later from the branch that will be created from this.

Issue #2702, opened 2021-10-13
[FEATURE REQUEST] Use state flag file in K8s upgrades


atsikham:
**Is your feature request related to a problem? Please describe.**
At the moment kubernetes upgrade tasks are executed even if upgrade steps were already completed. For example, for Kafka we [use](https://github.com/epiphany-platform/epiphany/blob/b19ae4dbb0ddb52a1bbb00646f462045de79798e/ansible/playbooks/roles/upgrade/tasks/kafka.yml#L31) a flag file to achieve that, but for K8s there is only a [TODO comment](https://github.com/epiphany-platform/epiphany/blob/b19ae4dbb0ddb52a1bbb00646f462045de79798e/ansible/playbooks/roles/upgrade/tasks/kubernetes.yml#L45).

**Describe the solution you'd like**
Use the same approach in K8s upgrade as we have in Kafka.

**Describe alternatives you've considered**

-

**Additional context**

-

---

**DoD checklist**

* [x] Changelog updated
* [x] COMPONENTS.md updated / doesn't need to be updated
* [x] Schema updated / doesn't need to be updated
* [x] Feature has automated tests
* [ ] Automated tests passed (QA pipelines)
  * [x] apply
  * [x] upgrade
* [ ] Idempotency tested
* [x] Documentation added / updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Solution meets requirements and is done according to design doc
* [x] Usage compliant with license
* [ ] Backport tasks created / doesn't need to be backported

PR #2766, opened 2021-12-01
Add state flag file in K8s upgrades (#2702)

cicharka:
* it is based on solution under upgrade/tasks/kafka.yml
* added usage of update flag for masters and nodes - having single
  flag file for both was introducing malfunction (incomplete master
  upgrade flag was removed by tasks in nodes block)
* 'Switch apiserver address to local' task is now executed before the
  upgrade procedure in order to correctly fetch versions of particular
  masters - without it procedure for some masters could be skipped,
  because fetched version didn't match host it tried to upgrade
* separate _get_ and _set_ cluster version tasks
* introduce _initial_cluster_version_ used to evaluate need of
  an upgrade

Tested on cluster (Ubuntu machines) containing:
1. three k8s masters and three k8s nodes (HA) (upgrade from v1.0.1 to v1.3.0)

### DIFF
--- a/ansible/playbooks/roles/upgrade/defaults/main.yml
+++ b/ansible/playbooks/roles/upgrade/defaults/main.yml
@@ -19,3 +19,6 @@ opendistro_for_elasticsearch:
       filename: demo2epiphany-certs-migration-root-CAs.pem
 
   upgrade_state_file_path: /etc/elasticsearch/epicli-upgrade-started.state
+
+kubernetes:
+  upgrade_flag_path: /tmp/kubernetes-upgrade-in-progress{{ ver }}.flag
--- a/ansible/playbooks/roles/upgrade/defaults/main.yml
+++ b/ansible/playbooks/roles/upgrade/defaults/main.yml
@@ -21,4 +21,5 @@ opendistro_for_elasticsearch:
   upgrade_state_file_path: /etc/elasticsearch/epicli-upgrade-started.state
 
 kubernetes:
-  upgrade_state_file_path: /var/lib/epiphany/upgrade/state/kubernetes-{{ ver }}.uncompleted
+  upgrade_master_state_file_path: /var/lib/epiphany/upgrade/state/kubernetes-master-{{ ver }}.uncompleted
+  upgrade_node_state_file_path: /var/lib/epiphany/upgrade/state/kubernetes-node-{{ ver }}.uncompleted
--- a/ansible/playbooks/roles/upgrade/defaults/main.yml
+++ b/ansible/playbooks/roles/upgrade/defaults/main.yml
@@ -21,5 +21,4 @@ opendistro_for_elasticsearch:
   upgrade_state_file_path: /etc/elasticsearch/epicli-upgrade-started.state
 
 kubernetes:
-  upgrade_master_state_file_path: /var/lib/epiphany/upgrade/state/kubernetes-master-{{ ver }}.uncompleted
-  upgrade_node_state_file_path: /var/lib/epiphany/upgrade/state/kubernetes-node-{{ ver }}.uncompleted
+  upgrade_state_file_path: /var/lib/epiphany/upgrade/state/kubernetes-{{ ver }}.uncompleted
--- a/ansible/playbooks/roles/upgrade/defaults/main.yml
+++ b/ansible/playbooks/roles/upgrade/defaults/main.yml
@@ -21,4 +21,4 @@ opendistro_for_elasticsearch:
   upgrade_state_file_path: /etc/elasticsearch/epicli-upgrade-started.state
 
 kubernetes:
-  upgrade_flag_path: /tmp/kubernetes-upgrade-in-progress{{ ver }}.flag
+  upgrade_state_file_path: /var/lib/epiphany/upgrade/state/kubernetes-{{ ver }}.uncompleted
--- a/ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
+++ b/ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
@@ -1,6 +1,8 @@
 ---
 - name: Wait for kube-apiserver and get cluster version
-  delegate_to: "{{ groups.kubernetes_master[0] }}"
+  delegate_to: >-
+    {{ inventory_hostname if inventory_hostname in groups.kubernetes_master else
+      groups.kubernetes_master[0] }}
   block:
     - name: k8s | Include wait-for-kube-apiserver.yml
       import_tasks: kubernetes/utils/wait-for-kube-apiserver.yml
--- a/ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
+++ b/ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
@@ -12,10 +12,10 @@
   import_tasks: kubernetes/get-kubelet-version.yml # sets kubelet_version
   delegate_to: "{{ groups.kubernetes_master[0] }}"
 
-- name: Check for upgrade flag file
+- name: Check for upgrade state file
   stat:
     path: "{{ kubernetes.upgrade_state_file_path }}"
-  register: lock_file_status
+  register: upgrade_state_file_status
 
 - name: Upgrade masters then nodes
   vars:
@@ -24,13 +24,14 @@
   block:
     - name: Upgrade masters
       when: >
-        lock_file_status.stat.exists or
+        upgrade_state_file_status.stat.exists or
         cluster_version is version('v' + version, '<')
       block:
-        - name: Create upgrade flag file
-          file:
-            path: "{{ kubernetes.upgrade_state_file_path }}"
-            state: touch
+        - name: Create upgrade state file
+          copy:
+            dest: "{{ kubernetes.upgrade_state_file_path }}"
+            content: Upgrade started
+            mode: u=rw,g=r,o=
 
         - name: k8s | Upgrade first master to v{{ version }}
           include_tasks: kubernetes/upgrade-master0.yml
@@ -42,20 +43,21 @@
           when:
             - inventory_hostname in groups.kubernetes_master[1:]
 
-        - name: Remove k8s upgrade flag file
+        - name: Remove k8s upgrade state file
           file:
             path: "{{ kubernetes.upgrade_state_file_path }}"
             state: absent
 
     - name: Upgrade nodes
       when: >
-        lock_file_status.stat.exists or
+        upgrade_state_file_status.stat.exists or
         kubelet_version is version('v' + version, '<')
       block:
-        - name: Create upgrade flag file
-          file:
-            path: "{{ kubernetes.upgrade_state_file_path }}"
-            state: touch
+        - name: Create upgrade state file
+          copy:
+            dest: "{{ kubernetes.upgrade_state_file_path }}"
+            content: Upgrade started
+            mode: u=rw,g=r,o=
 
         - name: k8s | Upgrade node to v{{ version }}
           include_tasks: kubernetes/upgrade-node.yml
@@ -63,7 +65,7 @@
             - groups.kubernetes_node is defined
             - inventory_hostname in groups.kubernetes_node
 
-        - name: Remove k8s upgrade flag file
+        - name: Remove k8s upgrade state file
           file:
             path: "{{ kubernetes.upgrade_state_file_path }}"
             state: absent
--- a/ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
+++ b/ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
@@ -1,4 +1,14 @@
 ---
+# During HA control plane upgrade server address in kubeconfig is switched to local for
+#   * compatibility between client and server versions
+#   * identifying correct server version
+
+- name: k8s/master | Switch apiserver address to local
+  include_tasks: kubernetes/utils/set-local-apiserver.yml # sets kubectl_context_cluster
+  when:
+    - groups.kubernetes_master | length > 1
+    - inventory_hostname in groups.kubernetes_master
+
 - name: Wait for kube-apiserver and get cluster version
   delegate_to: >-
     {{ inventory_hostname if inventory_hostname in groups.kubernetes_master else
@@ -14,10 +24,15 @@
   import_tasks: kubernetes/get-kubelet-version.yml # sets kubelet_version
   delegate_to: "{{ groups.kubernetes_master[0] }}"
 
-- name: Check for upgrade state file
+- name: Check upgrade state file for master
   stat:
-    path: "{{ kubernetes.upgrade_state_file_path }}"
-  register: upgrade_state_file_status
+    path: "{{ kubernetes.upgrade_master_state_file_path }}"
+  register: upgrade_master_state_file_status
+
+- name: Check upgrade state file for node
+  stat:
+    path: "{{ kubernetes.upgrade_node_state_file_path }}"
+  register: upgrade_node_state_file_status
 
 - name: Upgrade masters then nodes
   vars:
@@ -26,12 +41,12 @@
   block:
     - name: Upgrade masters
       when: >
-        upgrade_state_file_status.stat.exists or
+        upgrade_master_state_file_status.stat.exists or
         cluster_version is version('v' + version, '<')
       block:
-        - name: Create upgrade state file
+        - name: Create k8s upgrade state file for masters
           copy:
-            dest: "{{ kubernetes.upgrade_state_file_path }}"
+            dest: "{{ kubernetes.upgrade_master_state_file_path }}"
             content: Upgrade started
             mode: u=rw,g=r,o=
 
@@ -45,19 +60,19 @@
           when:
             - inventory_hostname in groups.kubernetes_master[1:]
 
-        - name: Remove k8s upgrade state file
+        - name: Remove k8s upgrade state file for masters
           file:
-            path: "{{ kubernetes.upgrade_state_file_path }}"
+            path: "{{ kubernetes.upgrade_master_state_file_path }}"
             state: absent
 
     - name: Upgrade nodes
       when: >
-        upgrade_state_file_status.stat.exists or
+        upgrade_node_state_file_status.stat.exists or
         kubelet_version is version('v' + version, '<')
       block:
-        - name: Create upgrade state file
+        - name: Create k8s upgrade state file for nodes
           copy:
-            dest: "{{ kubernetes.upgrade_state_file_path }}"
+            dest: "{{ kubernetes.upgrade_node_state_file_path }}"
             content: Upgrade started
             mode: u=rw,g=r,o=
 
@@ -67,10 +82,19 @@
             - groups.kubernetes_node is defined
             - inventory_hostname in groups.kubernetes_node
 
-        - name: Remove k8s upgrade state file
+        - name: Remove k8s upgrade state file for nodes
           file:
-            path: "{{ kubernetes.upgrade_state_file_path }}"
+            path: "{{ kubernetes.upgrade_node_state_file_path }}"
             state: absent
+
+
+- name: k8s/master | Switch apiserver address to HAProxy
+  command: |-
+    kubectl config set-cluster {{ kubectl_context_cluster.stdout }} --server=https://localhost:3446
+  when:
+    - groups.kubernetes_master | length > 1
+    - inventory_hostname in groups.kubernetes_master
+  changed_when: true
 
 - name: k8s | Upgrade internal haproxy load-balancer
   import_tasks: kubernetes/upgrade-haproxy.yml
--- a/ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
+++ b/ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
@@ -18,21 +18,30 @@
       import_tasks: kubernetes/utils/wait-for-kube-apiserver.yml
 
     - name: k8s | Include get-cluster-version.yml
-      import_tasks: kubernetes/get-cluster-version.yml # sets cluster_version
+      import_tasks: kubernetes/get-cluster-version.yml
+
+    - name: k8s | Set cluster version facts
+      set_fact:
+        initial_cluster_version: "{{ _cluster_version }}"
+        cluster_version: "{{ _cluster_version }}"
+      vars:
+        _cluster_version: "{{ (kubectl_cluster_version.stdout | from_yaml).serverVersion.gitVersion }}"
 
 - name: k8s | Include get-kubelet-version.yml
-  import_tasks: kubernetes/get-kubelet-version.yml # sets kubelet_version
+  import_tasks: kubernetes/get-kubelet-version.yml
   delegate_to: "{{ groups.kubernetes_master[0] }}"
 
-- name: Check upgrade state file for master
-  stat:
-    path: "{{ kubernetes.upgrade_master_state_file_path }}"
-  register: upgrade_master_state_file_status
+- name: k8s | Set kubelet version as fact
+  set_fact:
+    initial_kubelet_version: "{{ kubelet_version.stdout }}"
 
-- name: Check upgrade state file for node
+- name: Check if upgrade state file exists
   stat:
-    path: "{{ kubernetes.upgrade_node_state_file_path }}"
-  register: upgrade_node_state_file_status
+    path: "{{ kubernetes.upgrade_state_file_path }}"
+    get_attributes: false
+    get_checksum: false
+    get_mime: false
+  register: k8s_upgrade_state_file_status
 
 - name: Upgrade masters then nodes
   vars:
@@ -40,13 +49,14 @@
     cni_version: "{{ cni_ver }}"
   block:
     - name: Upgrade masters
-      when: >
-        upgrade_master_state_file_status.stat.exists or
-        cluster_version is version('v' + version, '<')
+      when:
+        - inventory_hostname in groups.kubernetes_master
+        - k8s_upgrade_state_file_status.stat.exists
+          or initial_cluster_version is version('v' + version, '<')
       block:
-        - name: Create k8s upgrade state file for masters
+        - name: Create K8s upgrade state file on master node
           copy:
-            dest: "{{ kubernetes.upgrade_master_state_file_path }}"
+            dest: "{{ kubernetes.upgrade_state_file_path }}"
             content: Upgrade started
             mode: u=rw,g=r,o=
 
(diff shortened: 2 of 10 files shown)


seriva (dismissed):

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -12,13 +12,22 @@
...
     version: "{{ ver }}"
     cni_version: "{{ cni_ver }}"
   block:
+    - name: Create upgrade flag file
+      file:
+        path: "{{ kubernetes.upgrade_flag_path }}"
+        state: touch
+
     - name: Upgrade masters
-      when: cluster_version is version('v' + version, '<=')
```
atsikham, line 21: Here and below we have single conditions, while common one for the whole block may be passed in different cases. Correct me if this is a wrong example/case: upgrade process at this point because of `kubelet_version`, but we don't need to execute this task.

atsikham (changes requested):
Have you checked creation of files for each incremental upgrade?

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
cicharka: > Have you checked creation of files for each incremental upgrade?

Yes, I did, works correctly.

> Here and below we have single conditions, while common one for the whole block may be passed in different cases. Correct me if this is a wrong example/case: upgrade process at this point because of kubelet_version, but we don't need to execute this task

Do you mean that we should separate **Upgrde masters** and **Upgrade nodes** tasks, so if only _kubelet_version_ is lower (_cluster_version_ is already desired) we will execute only **Upgrade nodes** task, and we will not execute **Upgrade masters**?

I was thinking about it and:
- based on fact that kubeadm clusters upgrade has to follow the workflow (primary cp node, additional cp nodes, worker nodes) and
- idempotency character of `kubeadm upgrade` command

I assumed that they can be coupled together. But I can easily change it so we will have previous conditions (I will revert lines 21 and 34 and I will just add flag condition to them)

atsikham: @to-bar what is your opinion here? May be we could approve these changes with additional comments?

ansible/playbooks/roles/upgrade/defaults/main.yml
```diff
@@ -19,3 +19,6 @@ opendistro_for_elasticsearch:
       filename: demo2epiphany-certs-migration-root-CAs.pem
 
   upgrade_state_file_path: /etc/elasticsearch/epicli-upgrade-started.state
+
+kubernetes:
+  upgrade_flag_path: /tmp/kubernetes-upgrade-in-progress{{ ver }}.flag
```
to-bar:
```suggestion
  upgrade_state_file_path: /var/lib/epiphany/upgrade/state/kubernetes-{{ ver }}.uncompleted
```
Since `/tmp` is cleaned up after reboot I would use `/var/lib/epiphany/upgrade/state` instead. This location is ready to use since it's created by a task from `common` role which is run before `upgrade` playbook.

Changing name to `state_file` is for naming consistency, previously we used `lock_file` or `flag_file` but IMO `state_file` is more informative/intuitive 😉

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
to-bar: We run upgrade in serial mode (host by host). Let's consider a case when the first master had been upgraded successfully (the state flag file was removed) but upgrade failed on next master or on any node. The question is whether we may need to run upgrade tasks again on the first master (or 2nd master when failure was on 3rd one)? If we do not need to rerun upgrade tasks on any successfully upgraded node, I would vote for saving time and having separate conditions 😉

to-bar (changes requested):

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
cicharka: Cool, I will change it then :)

atsikham:
Just FYI, this PR should fix such issue that I met recently.

```
TASK [upgrade : k8s/install | Install kubeadm 1.20.12 package for Debian family] *************************************************************************
fatal: [ec2-18-212-2-237.compute-1.amazonaws.com]: FAILED! => {"cache_update_time": 1639014589, "cache_updated": true, "changed": false, "msg": "'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\"      install 'kubeadm=1.20.12-00' 'kubelet=1.20.12-00'' failed: E: Packages were downgraded and -y was used without --allow-downgrades.\n", "rc": 100, "stderr": "E: Packages were downgraded and -y was used without --allow-downgrades.\n", "stderr_lines": ["E: Packages were downgraded and -y was used without --allow-downgrades."], "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\nThe following packages will be DOWNGRADED:\n  kubeadm kubelet\n0 upgraded, 0 newly installed, 2 downgraded, 0 to remove and 56 not upgraded.\n", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information...", "The following packages will be DOWNGRADED:", "  kubeadm kubelet", "0 upgraded, 0 newly installed, 2 downgraded, 0 to remove and 56 not upgraded."]}
```

cicharka:
> Just FYI, this PR should fix such issue that I met recently.
> 
> ```
> TASK [upgrade : k8s/install | Install kubeadm 1.20.12 package for Debian family] *************************************************************************
> fatal: [ec2-18-212-2-237.compute-1.amazonaws.com]: FAILED! => {"cache_update_time": 1639014589, "cache_updated": true, "changed": false, "msg": "'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\"      install 'kubeadm=1.20.12-00' 'kubelet=1.20.12-00'' failed: E: Packages were downgraded and -y was used without --allow-downgrades.\n", "rc": 100, "stderr": "E: Packages were downgraded and -y was used without --allow-downgrades.\n", "stderr_lines": ["E: Packages were downgraded and -y was used without --allow-downgrades."], "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\nThe following packages will be DOWNGRADED:\n  kubeadm kubelet\n0 upgraded, 0 newly installed, 2 downgraded, 0 to remove and 56 not upgraded.\n", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information...", "The following packages will be DOWNGRADED:", "  kubeadm kubelet", "0 upgraded, 0 newly installed, 2 downgraded, 0 to remove and 56 not upgraded."]}
> ```

@atsikham I have not faced issue like that before, do you remember the way to reproduce it so I will be able to verify?

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
cicharka: I moved PR to draft since I am not happy yet with solution provided for this part.

atsikham:
> > Just FYI, this PR should fix such issue that I met recently.
> > ```
> > TASK [upgrade : k8s/install | Install kubeadm 1.20.12 package for Debian family] *************************************************************************
> > fatal: [ec2-18-212-2-237.compute-1.amazonaws.com]: FAILED! => {"cache_update_time": 1639014589, "cache_updated": true, "changed": false, "msg": "'/usr/bin/apt-get -y -o \"Dpkg::Options::=--force-confdef\" -o \"Dpkg::Options::=--force-confold\"      install 'kubeadm=1.20.12-00' 'kubelet=1.20.12-00'' failed: E: Packages were downgraded and -y was used without --allow-downgrades.\n", "rc": 100, "stderr": "E: Packages were downgraded and -y was used without --allow-downgrades.\n", "stderr_lines": ["E: Packages were downgraded and -y was used without --allow-downgrades."], "stdout": "Reading package lists...\nBuilding dependency tree...\nReading state information...\nThe following packages will be DOWNGRADED:\n  kubeadm kubelet\n0 upgraded, 0 newly installed, 2 downgraded, 0 to remove and 56 not upgraded.\n", "stdout_lines": ["Reading package lists...", "Building dependency tree...", "Reading state information...", "The following packages will be DOWNGRADED:", "  kubeadm kubelet", "0 upgraded, 0 newly installed, 2 downgraded, 0 to remove and 56 not upgraded."]}
> > ```
> 
> @atsikham I have not faced issue like that before, do you remember the way to reproduce it so I will be able to verify?

To reproduce that you need to run K8s upgrade on Ubuntu, interrupt that after the new kubeadm package is installed, and then run again.

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
cicharka: So finally I just moved state file check for masters and nodes. Inside upgrade of masters, the upgrade of first master and next master will be based on inventory check.

ansible/playbooks/roles/upgrade/defaults/main.yml
cicharka: Done!

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -12,14 +12,26 @@
   import_tasks: kubernetes/get-kubelet-version.yml # sets kubelet_version
   delegate_to: "{{ groups.kubernetes_master[0] }}"
 
+- name: Check for upgrade flag file
+  stat:
+    path: "{{ kubernetes.upgrade_state_file_path }}"
+  register: lock_file_status
```
to-bar:
```suggestion
- name: Check for upgrade state flag file
  stat:
    path: "{{ kubernetes.upgrade_state_file_path }}"
  register: upgrade_state_file_status
```
I would use upgrade state file for consistency and avoid using `lock` term since it is often used for preventing running multiple instances at the same time (e.g. by yum or apt).

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -12,14 +12,26 @@
...
+
 - name: Upgrade masters then nodes
   vars:
     version: "{{ ver }}"
     cni_version: "{{ cni_ver }}"
   block:
     - name: Upgrade masters
-      when: cluster_version is version('v' + version, '<=')
+      when: >
+        lock_file_status.stat.exists or
```
to-bar:
```suggestion
        upgrade_state_file_status.stat.exists or
```

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -12,14 +12,26 @@
...
+        cluster_version is version('v' + version, '<')
       block:
+        - name: Create upgrade flag file
+          file:
+            path: "{{ kubernetes.upgrade_state_file_path }}"
+            state: touch
```
to-bar:
```suggestion
        - name: Create upgrade state file
          copy:
            dest: "{{ kubernetes.upgrade_state_file_path }}"
            content: Upgrade started
            mode: u=rw,g=r,o=
```
`touch` will update existing file each run so Ansible will report state as `changed`. `copy` will report `ok` when file already exists.

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -30,16 +42,31 @@
...
 
+        - name: Remove k8s upgrade flag file
+          file:
+            path: "{{ kubernetes.upgrade_state_file_path }}"
+            state: absent
+
     - name: Upgrade nodes
-      when: kubelet_version is version('v' + version, '<=')
+      when: >
+        lock_file_status.stat.exists or
```
to-bar:
```suggestion
        upgrade_state_file_status.stat.exists or
```

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
cicharka: nice one, I will remember that!

atsikham (dismissed):

docs/changelogs/CHANGELOG-1.3.md
```diff
@@ -15,10 +15,14 @@
...
 - [#2703](https://github.com/epiphany-platform/epiphany/issues/2703) - Add tests for docker and kubelet cgroup driver
 - [#1076](https://github.com/epiphany-platform/epiphany/issues/1076) - Add sorting entries in the inventory file
+<<<<<<< HEAD
 - [#2768](https://github.com/epiphany-platform/epiphany/issues/2768) - Add posibility to provide custom hostnames
 - [#2785](https://github.com/epiphany-platform/epiphany/issues/2785) - Add configuration option to Keycloak for PROXY_ADDRESS_FORWARDING env. variable
 - [#2814](https://github.com/epiphany-platform/epiphany/issues/2814) - Add description how to enable TLS in Kibana
 - [#1076](https://github.com/epiphany-platform/epiphany/issues/2595) - Document connection protocols and ciphers
+=======
+- [#2702](https://github.com/epiphany-platform/epiphany/issues/2702) - Use state flag file in K8s upgrades
+>>>>>>> 743a2549 (Changelog updated)
```
atsikham: Requires change

cicharka: Already fixed

ansible/playbooks/roles/upgrade/defaults/main.yml
```diff
@@ -19,3 +19,7 @@ opendistro_for_elasticsearch:
       filename: demo2epiphany-certs-migration-root-CAs.pem
 
   upgrade_state_file_path: /etc/elasticsearch/epicli-upgrade-started.state
+
+kubernetes:
+  upgrade_master_state_file_path: /var/lib/epiphany/upgrade/state/kubernetes-master-{{ ver }}.uncompleted
+  upgrade_node_state_file_path: /var/lib/epiphany/upgrade/state/kubernetes-node-{{ ver }}.uncompleted
```
to-bar:
```suggestion
kubernetes:
  upgrade_state_file_path: /var/lib/epiphany/upgrade/state/kubernetes-{{ ver }}.uncompleted
```
kubernetes_master and kubernetes_node roles are exclusive so how about common file?

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -1,25 +1,67 @@
...
+      groups.kubernetes_master[0] }}
   block:
     - name: k8s | Include wait-for-kube-apiserver.yml
       import_tasks: kubernetes/utils/wait-for-kube-apiserver.yml
 
     - name: k8s | Include get-cluster-version.yml
-      import_tasks: kubernetes/get-cluster-version.yml # sets cluster_version
+      import_tasks: kubernetes/get-cluster-version.yml
+
+    - name: Set initial_cluster_version fact # sets initial_cluster_version
```
to-bar:
```suggestion
    - name: Set initial_cluster_version fact
```
The fact name is just few lines below

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -1,25 +1,67 @@
...
+
+- name: Check upgrade state file for master
+  stat:
+    path: "{{ kubernetes.upgrade_master_state_file_path }}"
+  register: upgrade_master_state_file_status
+
+- name: Check upgrade state file for node
+  stat:
+    path: "{{ kubernetes.upgrade_node_state_file_path }}"
+  register: upgrade_node_state_file_status
```
to-bar:
```suggestion
```

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -1,25 +1,67 @@
...
   delegate_to: "{{ groups.kubernetes_master[0] }}"
 
+- name: Set kubelet version as fact
+  set_fact:
+    initial_kubelet_version: "{{ kubelet_version.stdout }}"
+
+- name: Check upgrade state file for master
+  stat:
+    path: "{{ kubernetes.upgrade_master_state_file_path }}"
+  register: upgrade_master_state_file_status
```
to-bar:
```suggestion
- name: Check if upgrade state file exists
  stat:
    path: "{{ kubernetes.upgrade_state_file_path }}"
    get_attributes: false
    get_checksum: false
    get_mime: false
  register: k8s_upgrade_state_file_status
```

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -1,25 +1,67 @@
...
+
 - name: Upgrade masters then nodes
   vars:
     version: "{{ ver }}"
     cni_version: "{{ cni_ver }}"
   block:
     - name: Upgrade masters
-      when: cluster_version is version('v' + version, '<=')
+      when: >
+        upgrade_master_state_file_status.stat.exists or
```
to-bar:
```suggestion
        k8s_upgrade_state_file_status.stat.exists or
```

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -1,25 +1,67 @@
...
+        initial_cluster_version is version('v' + version, '<')
       block:
+        - name: Create k8s upgrade state file for masters
+          copy:
+            dest: "{{ kubernetes.upgrade_master_state_file_path }}"
```
to-bar:
```suggestion
        - name: Create K8s upgrade state file on master node
          copy:
            dest: "{{ kubernetes.upgrade_state_file_path }}"
```

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -30,16 +72,41 @@
           when:
             - inventory_hostname in groups.kubernetes_master[1:]
 
+        - name: Remove k8s upgrade state file for masters
+          file:
+            path: "{{ kubernetes.upgrade_master_state_file_path }}"
+            state: absent
```
to-bar:
```suggestion
        - name: Remove K8s upgrade state file on master node
          file:
            path: "{{ kubernetes.upgrade_state_file_path }}"
            state: absent
```

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -30,16 +72,41 @@
...
+
     - name: Upgrade nodes
-      when: kubelet_version is version('v' + version, '<=')
+      when: >
+        upgrade_node_state_file_status.stat.exists or
```
to-bar:
```suggestion
        k8s_upgrade_state_file_status.stat.exists or
```

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -30,16 +72,41 @@
...
+        initial_kubelet_version is version('v' + version, '<')
       block:
+        - name: Create k8s upgrade state file for nodes
+          copy:
+            dest: "{{ kubernetes.upgrade_node_state_file_path }}"
```
to-bar:
```suggestion
        - name: Create K8s upgrade state file on node
          copy:
            dest: "{{ kubernetes.upgrade_state_file_path }}"
```

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -30,16 +72,41 @@
...
           include_tasks: kubernetes/upgrade-node.yml
           when:
             - groups.kubernetes_node is defined
             - inventory_hostname in groups.kubernetes_node
 
+        - name: Remove k8s upgrade state file for nodes
+          file:
+            path: "{{ kubernetes.upgrade_node_state_file_path }}"
+            state: absent
+
```
to-bar:
```suggestion
        - name: Remove K8s upgrade state file on node
          file:
            path: "{{ kubernetes.upgrade_state_file_path }}"
            state: absent
```

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -1,25 +1,67 @@
...
+    - name: k8s/kubelet-cm | Include set-cluster-version.yml
+      include_tasks: kubernetes/set-cluster-version.yml  # sets cluster_version
 
 - name: k8s | Include get-kubelet-version.yml
-  import_tasks: kubernetes/get-kubelet-version.yml # sets kubelet_version
+  import_tasks: kubernetes/get-kubelet-version.yml
   delegate_to: "{{ groups.kubernetes_master[0] }}"
 
+- name: Set kubelet version as fact
```
atsikham: As I see some tasks start with 'k8s |', and others aren't in the same place. Should be changed?

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -1,25 +1,67 @@
...
     - name: k8s | Include get-cluster-version.yml
-      import_tasks: kubernetes/get-cluster-version.yml # sets cluster_version
+      import_tasks: kubernetes/get-cluster-version.yml
+
+    - name: Set initial_cluster_version fact # sets initial_cluster_version
+      set_fact:
+        initial_cluster_version: >-
+          {{ (kubectl_cluster_version.stdout | from_yaml).serverVersion.gitVersion }}
+
+    - name: k8s/kubelet-cm | Include set-cluster-version.yml
```
atsikham: `k8s/kubelet-cm` - is that name correct?

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -30,16 +72,41 @@
...
           when:
             - groups.kubernetes_node is defined
             - inventory_hostname in groups.kubernetes_node
 
+        - name: Remove k8s upgrade state file for nodes
+          file:
+            path: "{{ kubernetes.upgrade_node_state_file_path }}"
+            state: absent
+
+
```
atsikham: Unnecessary newline. Could you apply formatting to be sure there is no such places and additional whitespaces?

atsikham (changes requested):

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
to-bar: Not sure if we need this task here. If `cluster_version` is used somewhere before `patch-kubelet-cm.yml`, then I would just add the fact to the previous task.

to-bar (changes requested):

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
cicharka: That name comes from [ansible/playbooks/roles/upgrade/tasks/kubernetes/patch-kubelet-cm.yml](https://github.com/epiphany-platform/epiphany/pull/2766/files/8039af8429e265e518dd817ad02b362ccca93d0f#diff-ee268bede56a4d59831f4e37a661445606a5f6669f046921cc627e5b0bbd1b51) so it is my mistake of not changing it here. I will fix it

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
cicharka: Honestly I am not aware that's the convention here, need to clarify.

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
cicharka: Yes, I will fix it.

ansible/playbooks/roles/upgrade/defaults/main.yml
cicharka: Condition for the block **Upgrade nodes** [here](https://github.com/epiphany-platform/epiphany/pull/2766/files#diff-b642cd0642ad945c5f7f283861c694a6c5fb4218ed56c80eabf9df27fecd615bR81-R83) doesn't check if we are operating on master or node inventory_hostname, therefore it always enters block, creates flag file, skips upgrade, and then removes flag file again -> so even if the flag was present because of the master failure, it was removed inside upgrade **node block**. 

I will add all of your suggested changes, along with the additional change in [this condition](https://github.com/epiphany-platform/epiphany/pull/2766/files#diff-b642cd0642ad945c5f7f283861c694a6c5fb4218ed56c80eabf9df27fecd615bR81-R83), where we will check based on inventory hostname if we have to enter **node upgrade** block.

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
cicharka: We are using all of the facts (`version`, `major_version` and `minor_version`) that we are setting in `set-cluster-version.yml` -> Therefore it would be a code duplication if we will set it manually in previous task.

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
cicharka: I will change some of the following schema:
- if task includes or import another task for specific component I will add k8s
- if task is performing action on specific component/service I will add k8s
- if task is responsible for keeping the upgrade procedure ongoing, traces it and is a helper, I will leave it without k8s prefix
Do you agree with the proposed form?

atsikham (dismissed):

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -1,25 +1,64 @@
...
+  register: k8s_upgrade_state_file_status
+
 - name: Upgrade masters then nodes
   vars:
     version: "{{ ver }}"
     cni_version: "{{ cni_ver }}"
   block:
     - name: Upgrade masters
-      when: cluster_version is version('v' + version, '<=')
+      when: >
```
to-bar: Since now we have common flag file I would extend condition to run this block only on master nodes:
```
      when:
        - inventory_hostname in groups.kubernetes_master
        - k8s_upgrade_state_file_status.stat.exists
          or initial_cluster_version is version('v' + version, '<')
```

ansible/playbooks/roles/upgrade/tasks/kubernetes/gather-static-facts.yml
```diff
@@ -8,6 +8,8 @@
         '.items[].metadata | select (.labels."k8s-app" == "canal"
                                   or .labels."k8s-app" == "calico-node"
                                   or .labels.app == "flannel") | .name'
+      args:
```
to-bar: This fix is already in develop.

to-bar (changes requested):

atsikham (dismissed):

seriva (dismissed):

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -1,25 +1,65 @@
 ---
+# During HA control plane upgrade server address in kubeconfig is switched to local for
+#   * compatibility between client and server versions
+#   * identifying correct server version
+
+- name: k8s/master | Switch apiserver address to local
+  include_tasks: kubernetes/utils/set-local-apiserver.yml # sets kubectl_context_cluster
```
to-bar, line 12: This file uses `kubectl` so to be on the safe side I would consider placing `k8s | Include wait-for-kube-apiserver.yml` task at the beginning, as we had before (it was the first task).

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
```diff
@@ -1,25 +1,65 @@
...
 - name: Wait for kube-apiserver and get cluster version
-  delegate_to: "{{ groups.kubernetes_master[0] }}"
+  delegate_to: >-
+    {{ inventory_hostname if inventory_hostname in groups.kubernetes_master else
+      groups.kubernetes_master[0] }}
   block:
     - name: k8s | Include wait-for-kube-apiserver.yml
       import_tasks: kubernetes/utils/wait-for-kube-apiserver.yml
 
     - name: k8s | Include get-cluster-version.yml
```
to-bar, line 22: I think we can simplify this flow. To decide about skipping upgrade we need only two common things (doesn't matter if it is master or worker node):
1) flag file existence
2) kubelet version

This means that probably we do not have to get cluster version here but only later, when condition is met and upgrade is not skipped.

Tasks `Create K8s upgrade state file on master node` and `Create K8s upgrade state file on node` have the same content so we can have one common task even not using YAML anchors. We could set fact node_role { master | worker }, then have:
```
- name: k8s/{{ node_role  }} | Create K8s upgrade state file
   copy:
      dest: "{{ kubernetes.upgrade_state_file_path }}"
      content: Upgrade started
      mode: u=rw,g=r,o=
```
The same with removal tasks.

to-bar (changes requested):

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
cicharka: As discussed with @to-bar and @seriva, we will not refactor that part for current release as it is more serious code interference with the upgrade of K8s.
Ticket for refactoring: #2844

ansible/playbooks/roles/upgrade/tasks/kubernetes.yml
cicharka: I moved `Wait for kube-apiserver` task at the beginning of the upgrade.

to-bar (approved):

erzetpe:
Tests passed on Ubuntu 18.04 LTS and Red Hat 7.8.

erzetpe (approved):

Issue #2699, opened 2021-10-11
[UPGRADE] Upgrade Grafana to the latest version


rafzei:
**Is your feature request related to a problem? Please describe.**
Grafana's team doesn't directly specify EOL but there are few CVE's in our current version:

- [CVE-2021-39226](https://nvd.nist.gov/vuln/detail/CVE-2021-39226)
- [CVE-2020-27846](https://nvd.nist.gov/vuln/detail/CVE-2020-27846)
- [CVE-2021-27358](https://nvd.nist.gov/vuln/detail/CVE-2021-27358)
- [CVE-2021-28148](https://nvd.nist.gov/vuln/detail/CVE-2021-28148)
- [CVE-2021-27962](https://nvd.nist.gov/vuln/detail/CVE-2021-27962)
- [CVE-2021-28147](https://nvd.nist.gov/vuln/detail/CVE-2021-28147)

The latest released version is [8.2.4](https://github.com/grafana/grafana/releases/tag/v8.2.4)
Check if there are any breaking changes we should take care of when upgrading.

Additional context
Check if this fixes #2420 or if we can include this in this issue. Also check if we can do this refactoring when taking care of the upgrade: #2028

---

**DoD checklist**

* [x] Changelog updated
* [x] COMPONENTS.md updated / doesn't need to be updated
* [x] Schema updated / doesn't need to be updated
* [x] Feature has automated tests
* [x] Automated tests passed (QA pipelines)
  * [x] apply
  * [x] upgrade
* [x] Idempotency tested
* [x] Documentation added / updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Solution meets requirements and is done according to design doc
* [x] Usage compliant with license
* [x] Backport tasks created / doesn't need to be backported

mkyc:
Spike about versions was in #2719.

sbbroot:
Upgraded to the newest version: [8.2.5](https://github.com/grafana/grafana/releases/tag/v8.2.5)

przemyslavic:
✔️ Tested `apply` and `upgrade` from v1.0 (x86_64) and from v1.1 (aarch64).

Issue #2028, opened 2021-02-09
[FEATURE REQUEST] Refactor Grafana upgrade role


plirglo:
**Is your feature request related to a problem? Please describe.**
We keep grafana_version in two places (Grafana role defaults and hardcoded in grafana upgrade task)

**Describe the solution you'd like**
We need to refactor these roles to use in upgrade role grafana_version from grafana defaults. This should not be hardcoded in upgrade task.

przemyslavic:
Tested together with #2699.

PR #2767, opened 2021-12-01
Upgrade Grafana to 8.3.2 (#2699)

sbbroot:
* Upgrade Grafana to 8.2.5

* Upgrade dashboard 179 (Docker and Host Monitoring w/ Prometheus) to revision 7


atsikham (dismissed):

plirglo:
How do we handle upgrade of exisiting installation?

rafzei:
@sbbroot see here: https://github.com/epiphany-platform/epiphany/blob/develop/ansible/playbooks/roles/upgrade/tasks/grafana.yml

Please, test the upgrade scenario as well.

sbbroot:
Added upgrade refactor [#2028 ](https://app.zenhub.com/workspaces/epiphany-platform-5ef9d3c749afc20019782391/issues/epiphany-platform/epiphany/2028)

ansible/playbooks/roles/grafana/defaults/versions.yml
```diff
@@ -0,0 +1,11 @@
+---
```
seriva, line 1: Is this extra file really needed? I don't think there is any other place we actually have this and it only adds more code as we need to load this separately.

sbbroot: This is needed because main.yml expects variables from elsewhere (probably a bad design).
I had discussion with @to-bar and we agreed on that.

atsikham (dismissed):
I would probably use different names, but these changes look good, so approved.

seriva (dismissed):

ansible/playbooks/roles/upgrade/tasks/grafana.yml
```diff
@@ -6,17 +6,21 @@
...
-      grafana_version: 7.3.5
   block:
+    - name: Get Grafana version vars
+      include_vars:
+        file: roles/grafana/defaults/versions.yml
+        name: grafana_version_vars
+
+    - name: Set Grafana version fact
+      set_fact:
+        grafana_version: "{{ grafana_version_vars.grafana_version }}"
```
to-bar:
```suggestion
    - name: Set Grafana version facts
      set_fact:
        grafana_installed_version: "{{ ansible_facts.packages['grafana'][0].version }}"
        grafana_target_version: "{{ grafana_version_vars.grafana_version }}"
```

ansible/playbooks/roles/upgrade/tasks/grafana.yml
```diff
@@ -6,17 +6,21 @@
...
+
     - name: Grafana | Print versions
       debug:
         msg:
           - "Installed version: {{ ansible_facts.packages['grafana'][0].version }}"
```
to-bar:
```suggestion
          - "Installed version: {{ grafana_installed_version }}"
```

ansible/playbooks/roles/upgrade/tasks/grafana.yml
```diff
@@ -6,17 +6,21 @@
...
-          - "Target version: {{ grafana_defaults.grafana_version }}"
+          - "Target version: {{ grafana_version }}"
```
to-bar:
```suggestion
          - "Target version: {{ grafana_target_version }}"
```

ansible/playbooks/roles/upgrade/tasks/grafana.yml
```diff
@@ -25,4 +29,4 @@
         name: grafana
         tasks_from: install
       when:
-        - grafana_defaults.grafana_version is version(ansible_facts.packages['grafana'][0].version, '>')
+        - grafana_version is version(ansible_facts.packages['grafana'][0].version, '>')
```
to-bar:
```suggestion
        - grafana_target_version is version(grafana_installed_version, '>')
```
I would also introduce upgrade state file, like we have here:
https://github.com/epiphany-platform/epiphany/blob/9fd3e99b50bf73920480798b8bd336688f3746a3/ansible/playbooks/roles/upgrade/tasks/opendistro_for_elasticsearch-01.yml#L35-L37

ansible/playbooks/roles/grafana/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,97 @@
...
+    name: grafana_versions
+
+- name: Include upgrade defaults
+  include_vars:
+    file: defaults/upgrade.yml
+    name: upgrade_defaults
+
+- name: Populate service facts
+  service_facts:
+  when: ansible_facts.service is undefined
```
to-bar:
```suggestion
- name: Get information about installed packages as facts
  package_facts:
    manager: auto
  when: ansible_facts.packages is undefined
```
Grafana is installed as package so IMO it should be enough just to check whether it exists (without checking binary file and service - so we could remove `Check Grafana binary existance` task as well).

ansible/playbooks/roles/grafana/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,97 @@
...
+# If state file exists it means the previous run failed
+- name: Check for upgrade state file
+  stat:
+    path: "{{ upgrade_defaults.state_file_path }}"
+    get_attributes: false
+    get_checksum: false
+    get_mime: false
+  register: state_file_status
+
+- name: Create upgrade state file
```
to-bar: I would move this task to a place where we already know that upgrade is required (after `Check versions and set grafana_will_upgrade` task). If upgrade is already performed (rerun) there is no need to create the state file since upgrade is skipped.

ansible/playbooks/roles/grafana/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,97 @@
...
+
+        - name: Grafana | Upgrade
+          when:
+            - grafana_will_upgrade
+          block:
+            - name: Stop Grafana service
+              systemd:
+                state: stopped
+                name: grafana
+              when:
```
to-bar: If Grafana package is installed, then service should exists.

to-bar (changes requested):

ansible/playbooks/roles/grafana/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,85 @@
...
+  include_vars:
+    file: defaults/versions.yml
+    name: grafana_versions
+
+- name: Include upgrade defaults
+  include_vars:
+    file: defaults/upgrade.yml
+    name: upgrade_defaults
+
+- name: Populate service facts
```
to-bar: We need `ansible_facts.packages['grafana']` to get Grafana version so we need to use `package_facts` module. Hence there is no need to use `service_facts` since having grafana package present we will have `grafana.service` as well.

to-bar (changes requested):

erzetpe (dismissed):

ansible/playbooks/roles/grafana/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,81 @@
...
+- name: Include new version of defaults
+  include_vars:
+    file: defaults/versions.yml
+    name: grafana_versions
+
+- name: Include upgrade defaults
+  include_vars:
+    file: defaults/upgrade.yml
+    name: upgrade_defaults
+
```
to-bar, line 16: For consistency, I would use the same approach as we have for `opendistro_for_elasticsearch` or `kafka_exporter` adding such task:

```
- name: Assert that grafana package is installed
  assert:
    that: ansible_facts.packages['grafana'] is defined
    fail_msg:
      - grafana package not found, nothing to upgrade.
      - Re-apply your configuration or remove grafana group from Ansible inventory.
    quiet: true
```

Then some logic below could be simplify.

ansible/playbooks/roles/grafana/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,81 @@
...
+  register: state_file_status
+
+- name: Set grafana_will_upgrade fact
+  set_fact:
+    grafana_will_upgrade: false
+
+- name: Start Grafana upgrade
+  when:
+    - ansible_facts.services["grafana.service"] is defined
+    - grafana_stat.stat.exists
```
to-bar: The condition is incorrect since upgrade state file is created inside the nested block. I would simplify this to have such condition for main block:

```
  when: grafana_target_version is version(grafana_installed_version, '>')
     or state_file_status.stat.exists
```

Then `grafana_will_upgrade` would be unneeded.

to-bar (changes requested):

ansible/playbooks/roles/grafana/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,88 @@
...
+    grafana_will_upgrade: false
+
+- name: Start Grafana upgrade
+  when: grafana_target_version is version(grafana_installed_version, '>')
+     or state_file_status.stat.exists
+  block:
+    - name: Upgrade Grafana
+      when: ansible_facts.packages['grafana'] is defined
+      block:
+        - name: Set Grafana version facts
```
to-bar: We need this task before the block

to-bar (changes requested):

ansible/playbooks/roles/grafana/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,76 @@
...
+
+    # Since we do not manage custom resources like plugins (it's up to the user), upgrading grafana by just installing new binary is considered complete here.
+    # https://grafana.com/docs/grafana/latest/installation/upgrading/
+    - name: Run installation process
+      import_role:
+        name: grafana
+        tasks_from: install
+
+    # Start services again if needed
+    - name: start Grafana service
```
to-bar:
```suggestion
    - name: Start Grafana service
```

ansible/playbooks/roles/grafana/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,76 @@
...
+    get_checksum: false
+    get_mime: false
+  register: state_file_status
+
+- name: Set Grafana version facts
+  set_fact:
+    grafana_installed_version: "{{ ansible_facts.packages['grafana'][0].version }}"
+    grafana_target_version: "{{ grafana_versions.grafana_version }}"
+
+- name: Grafana | Print versions
```
to-bar:
```suggestion
- name: Print versions
```

ansible/playbooks/roles/grafana/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,76 @@
...
+  debug:
+    msg:
+      - "Installed version: {{ grafana_installed_version }}"
+      - "Target version: {{ grafana_target_version }}"
+
+- name: Start Grafana upgrade
```
to-bar:
```suggestion
- name: Upgrade Grafana
```

ansible/playbooks/roles/grafana/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,76 @@
...
+      systemd:
+        state: stopped
+        name: grafana
+
+    # Since we do not manage custom resources like plugins (it's up to the user), upgrading grafana by just installing new binary is considered complete here.
+    # https://grafana.com/docs/grafana/latest/installation/upgrading/
+    - name: Run installation process
+      import_role:
+        name: grafana
+        tasks_from: install
```
to-bar:
```suggestion
    - name: Run installation tasks
      include_tasks: install.yml
```

to-bar (changes requested):

ansible/playbooks/roles/grafana/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,74 @@
...
+      systemd:
+        state: stopped
+        name: grafana-server
+
+    # Since we do not manage custom resources like plugins (it's up to the user), upgrading grafana by just installing new binary is considered complete here.
+    # https://grafana.com/docs/grafana/latest/installation/upgrading/
+    - name: Run installation tasks
+      include_tasks: install.yml
+
+    # Start services again if needed
```
to-bar:
```suggestion
```

to-bar (approved):

atsikham (approved):

seriva:
LGTT&M!

seriva (approved):

przemyslavic:
v8.3.3 has already been released 🙈

atsikham:
> v8.3.3 has already been released 🙈

If this is not a fast change, I vote for keeping it as is and upgrade in future releases.

przemyslavic (approved):

przemyslavic:
> > v8.3.3 has already been released 🙈
> 
> If this is not a fast change, I vote for keeping it as is and upgrade in future releases.

Now we deploy/upgrade to v8.3.2. So this should be a simple change I think. But it's already tested so let's merge it if there are no significant changes in v8.3.3 👍

seriva:
Lets keep as it is for now, otherwise we can keep bumping versions

PR #2769, opened 2021-12-02
Switch to containerd

rafzei:


rafzei:
Docker is still needed in image-registry role. Probably there is also docker-related documentation in k8s context so I will update it tomorrow.

rafzei:
Im pushing it to feature/containerd, It has to be implemented in epicli v2.0.0. Not tested.

Issue #2380, opened 2021-06-15
[BUG] Unable to drain nodes with Istio application enabled due to PodDisruptionBudgets


przemyslavic:
**Describe the bug**
There is an issue with upgrading kubernetes cluster with Istio application enabled. It hangs for hours while trying to evict Istio pod because of a violation of `PodDisruptionBudgets`.
```
evicting pod istio-system/istio-ingressgateway-6c4468dd97-cc5tp
error when evicting pod "istio-ingressgateway-6c4468dd97-cc5tp" (will retry after 5s): Cannot evict pod as it would violate the pod's disruption budget.
```
```
root@ec2-x-x-x-x:~# kubectl describe PodDisruptionBudget -n istio-system --kubeconfig /etc/kubernetes/admin.conf
Name:           istio-ingressgateway
Namespace:      istio-system
Min available:  1
Selector:       app=istio-ingressgateway,istio=ingressgateway
Status:
    Allowed disruptions:  0
    Current:              1
    Desired:              1
    Total:                1
Events:                   <none>
Name:           istiod
Namespace:      istio-system
Min available:  1
Selector:       app=istiod,istio=pilot
Status:
    Allowed disruptions:  0
    Current:              1
    Desired:              1
    Total:                1
Events:                   <none>
```

We should have at least 2 replicas for `istio-ingressgateway` and `istiod` applications. Otherwise we won't be able to perform node drain/upgrade when PodDisruptionBudgets is enabled (here: `Allowed disruptions:  0`).

**How to reproduce**
1. Deploy a v0.9/v1.0 cluster with Istio deployment enabled
2. Perform upgrade to develop.

**Expected behavior**
The cluster has been successfully upgraded.

**Environment**
- Cloud provider: [all]
- OS: [all]

**epicli version**:
v0.9
v1.0
develop

**Additional context**
To fix this we have to scale up Istio to 2 replicas when:
`   - groups.kubernetes_node | length > 0 or k8s_as_cloud_service`
 
For single machine installation 1 replica probably should be ok.

Additional docs:
https://kubernetes.io/docs/tasks/administer-cluster/safely-drain-node/

There is also a `--disable-eviction` flag for `kubectl drain` command but I don't think we want to use this
```
      --disable-eviction=false: Force drain to use delete, even if eviction is supported. This will bypass checking
PodDisruptionBudgets, use with caution.
```

---

**DoD checklist**

* [ ] Changelog updated (if affected version was released)
[...]

PR #2771, opened 2021-12-06
Upgrade K8s to v1.21

atsikham:
* Update versions
* Changes for https://github.com/kubernetes/kubeadm/issues/2525
* Remove deprecated admission plugin


seriva:
@atsikham Do we need any preflight checks for the repository machines disk space when we are pulling in so many new images? I remember @to-bar was doing somethings for that before. Maybe it needs to be adjusted?

atsikham:
>@atsikham Do we need any preflight checks for the repository machines disk space when we are pulling in so many new images? I remember @to-bar was doing somethings for that before. Maybe it needs to be adjusted?

I don't see anything related to K8s/repository disk space in preflight checks, but it could be investigated in additional task. In fact, it would be nice to have such check independently on the number images we have, according to the currently consumed disk space on machines.
@przemyslavic could you check amount of used/free disk space for repository and K8s control plane node after pipelines execution? That way we can understand if there is something critical.

ansible/playbooks/roles/upgrade/tasks/kubernetes/patch-kubeadm-apiserver.yml
```diff
@@ -16,25 +16,29 @@
...
+      register: command_kubeadm_configmap
+      changed_when: false
+
     # The following procedure ensures that etcd encryption is always enabled during subsequent kubeadm executions
-    - name: k8s/master | Patch kubeadm-config configmap (patch-kubeadm-etcd-encryption.yml)
-      when:
-        - _kubeadm_api_server_extra_args['encryption-provider-config'] is undefined
+    # Starting from K8s v1.21 support for DenyEscalatingExec admission plugin is removed,
+    # but it was specified in Epiphany's default values
+    - name: k8s/master | Patch kubeadm-config configmap (patch-kubeadm-apiserver.yml)
```
to-bar, line 37: Here we have `k8s/master` prefix but in the previous task `k8s/utils`. How about using the same prefix across file for consistency?

ansible/playbooks/roles/upgrade/tasks/kubernetes/patch-kubeadm-apiserver.yml
```diff
@@ -16,25 +16,29 @@
...
+      when: _kubeadm_api_server_extra_args['encryption-provider-config'] is undefined or
+        (version is version('1.21', '>=')
+        and version is version('1.22', '<')
+        and _kubeadm_api_server_extra_args['enable-admission-plugins'] is defined)
```
to-bar:
```suggestion
      when: _kubeadm_api_server_extra_args['encryption-provider-config'] is undefined
         or (version is version('1.21', '>=')
             and version is version('1.22', '<')
             and _kubeadm_api_server_extra_args['enable-admission-plugins'] is defined)
```

ansible/playbooks/roles/upgrade/tasks/kubernetes/patch-kubeadm-image-repository.yml
```diff
@@ -2,7 +2,7 @@
 - name: k8s/master | Patch imageRepository in kubeadm-config ConfigMap
   run_once: true
   block:
-    - name: k8s/master | Get kubeadm-config configmap
+    - name: k8s/utils | Get kubeadm-config configmap
```
to-bar: The next task has different prefix. I would prefer to use the same in across the file (unless task is delegated).

docs/changelogs/CHANGELOG-1.3.md
```diff
@@ -29,12 +29,13 @@
...
-- Upgrade Calico and Canal to v3.20.2
-- Upgrade Coredns to v1.7.0
+- Upgrade Calico and Canal to v3.20.3
+- Upgrade Coredns to v1.8.0
 - Upgrade Kubernetes dashboard to v2.3.1
 - Upgrade Kubernetes metrics-scraper to v1.0.7
 - [#2093](https://github.com/epiphany-platform/epiphany/issues/2093) - Upgrade K8s to v1.19.15 (transitional version)
-- [#2658](https://github.com/epiphany-platform/epiphany/issues/2658) - Upgrade K8s to v1.20.12
+- [#2658](https://github.com/epiphany-platform/epiphany/issues/2658) - Upgrade K8s to v1.20.12 (transitional version)
+- [#2659](https://github.com/epiphany-platform/epiphany/issues/2659) - Upgrade K8s to v1.21.7
```
to-bar, line 36: How about having only the latest entry? IMO, from user perspective, there is no value in information about transitional versions.

to-bar (changes requested):

ansible/playbooks/roles/upgrade/tasks/kubernetes/patch-kubeadm-image-repository.yml
atsikham: This is a mistake left after trying moving that part into a separate file.

erzetpe (approved):

to-bar (approved):

to-bar:
> @atsikham Do we need any preflight checks for the repository machines disk space when we are pulling in so many new images? I remember @to-bar was doing somethings for that before. Maybe it needs to be adjusted?

I only added removing old image files (*.tar) from epirepo that no longer exist in requirements.txt to free up disk space. However old images are not removed from image registry so we may still have "snowball effect" and something to do with this. In the future we can add size attribute for each image and count how much space is needed to download them and load into registry.

przemyslavic:
~~:x: Upgrade of HA cluster does not work :cry: It's been hanging for hours on `TASK [upgrade : k8s/utils | Drain master or node in preparation for maintenance]`.~~

:heavy_check_mark: `apply` x86_64, aarch64
:heavy_check_mark: `apply` HA cluster
:heavy_check_mark: promote to HA
:heavy_check_mark: single machine installation
:heavy_check_mark: `upgrade` from v1.0 x86_64
:heavy_check_mark: `upgrade` from v1.1 aarch64
:heavy_check_mark: `upgrade` from v1.0 single machine installation
:heavy_check_mark: `upgrade` from v1.0 HA installation

The issue was caused by Istio deployment #2380.

przemyslavic:
As for disk usage, it is as follows:
- `apply` 
Control Plane: 30/32 GB by default
```
[ubuntu@ec2-1-1-1-1 ~]$ df -h /
Filesystem      Size  Used Avail Use% Mounted on
/dev/nvme0n1p1   30G  4.5G   25G  16% /
```
Repository: 64 GB by default
```
df -h[ubuntu@ec2-2-2-2-2 ~]$ df -h /
Filesystem      Size  Used Avail Use% Mounted on
/dev/xvda1       62G   17G   46G  28% /
```
Single machine installation: 30/32 GB by default
```
[ubuntu@ec2-1-1-1-1 ~]$ df -h /
Filesystem      Size  Used Avail Use% Mounted on
/dev/nvme0n1p1   30G   19G   11G  64% /
```

- `upgrade` 
Control Plane:
```
df -[ubuntu@ec2-1-1-1-1 ~]$ df -h /
Filesystem      Size  Used Avail Use% Mounted on
/dev/nvme0n1p1   30G  6.9G   23G  24% /
```
Repository:
```
[ubuntu@ec2-2-2-2-2 ~]$ df -h /
Filesystem      Size  Used Avail Use% Mounted on
/dev/xvda1       62G   23G   40G  36% /
```
Single machine:
```
[ubuntu@ec2-1-1-1-1 ~]$ df -h /
Filesystem      Size  Used Avail Use% Mounted on
/dev/nvme0n1p1   30G   25G  4.9G  84% /
```

It doesn't look bad. Only in the case of a single machine installation with a default 30 GB disk, the used space is around 80%.

przemyslavic (approved):

przemyslavic:
:heart:

Issue #1221, opened 2020-04-29
kafka-exporter service doesn't start after restarting kafka VMs


to-bar:
```
[root@ec2-15-236-37-69 ~]# systemctl status kafka-exporter
● kafka-exporter.service - Service that runs Kafka Exporter
   Loaded: loaded (/etc/systemd/system/kafka-exporter.service; enabled; vendor preset: disabled)
   Active: failed (Result: start-limit) since Tue 2020-04-28 13:32:02 UTC; 6h ago
  Process: 3532 ExecStart=/opt/kafka_exporter/kafka_exporter --kafka.server=ec2-15-236-37-69.eu-west-3.compute.amazonaws.com:9092 --kafka.server=ec2-35-180-51-166.eu-west-3.compute.amazonaws.com:9092 --web.listen-address=:9308 --web.telemetry-path=/metrics --log.level=info --topic.filter=.* --group.filter=.* --kafka.version=2.0.0 (code=exited, status=2)
 Main PID: 3532 (code=exited, status=2)

Apr 28 13:32:02 ec2-15-236-37-69.eu-west-3.compute.amazonaws.com systemd[1]: Unit kafka-exporter.service entered failed state.
Apr 28 13:32:02 ec2-15-236-37-69.eu-west-3.compute.amazonaws.com systemd[1]: kafka-exporter.service failed.
Apr 28 13:32:02 ec2-15-236-37-69.eu-west-3.compute.amazonaws.com systemd[1]: kafka-exporter.service holdoff time over, scheduling restart.
Apr 28 13:32:02 ec2-15-236-37-69.eu-west-3.compute.amazonaws.com systemd[1]: Stopped Service that runs Kafka Exporter.
Apr 28 13:32:02 ec2-15-236-37-69.eu-west-3.compute.amazonaws.com systemd[1]: start request repeated too quickly for kafka-exporter.service
Apr 28 13:32:02 ec2-15-236-37-69.eu-west-3.compute.amazonaws.com systemd[1]: Failed to start Service that runs Kafka Exporter.
Apr 28 13:32:02 ec2-15-236-37-69.eu-west-3.compute.amazonaws.com systemd[1]: Unit kafka-exporter.service entered failed state.
Apr 28 13:32:02 ec2-15-236-37-69.eu-west-3.compute.amazonaws.com systemd[1]: kafka-exporter.service failed.

[root@ec2-15-236-37-69 ~]# journalctl -u kafka-exporter
[...]

PR #2763, opened 2021-11-30
#2700 upgrade Prometheus to v2.31.1 + AlertManager to v0.23.0

seriva:
- Updated Prometheus  to 2.31.1
- Updated AlertManager to 0.23.0
- Added upgrade path for Prometheus  and AlertManager

PR #2772, opened 2021-12-06
Upgrade Kafka Exporter to v1.4.2

plirglo:
- Upgrade Kafka Exporter to version 1.4.2
- Fix for [#1221](https://github.com/epiphany-platform/epiphany/issues/1221)


cli/epicli.py
```diff
@@ -253,6 +253,7 @@ def upgrade_parser(subparsers):
         'ignite',
         'image_registry',
         'kafka',
+        'kafka-exporter',
```
cicharka: Minor comment - this is the only place where you are using dash instead of underscore (other components in cli are using underscores)

seriva: I would suggest to keep it consistent and use underscore.

ansible/playbooks/roles/kafka_exporter/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,96 @@
...
+
+  - name: Upgrade | Kafka Exporter | Avoid risk of downgrade
+    when: kafka_exporter_version_old is not version( kafka_exporter.version, '<' )
+    block:
+    - name: Upgrade | Kafka Exporter | Remove upgrade flag file
+      file:
+        path: "{{ lock_file }}"
+        state: absent
+
+    - name: Upgrade | Kafka Exporter | Fail task
```
cicharka: This will fail whole **upgrade** playbook - for me it is not necessary to fail and stop playbook execution (I mean, whole upgrade procedure) because we already have newer version.

cicharka: * I wanted to reference here part:
` - name: Upgrade | Kafka Exporter | Fail task
     fail:
        msg: Kafka Exporter in newer version already installed!
 `

ansible/playbooks/roles/kafka_exporter/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,96 @@
+---
+- name: Upgrade | Kafka Exporter | Check for upgrade flag file
```
seriva: **Upgrade | Kafka Exporter |** is not needed here. See @to-bar  explanation here: https://github.com/epiphany-platform/epiphany/pull/2763#discussion_r761328303

erzetpe (dismissed):

cli/epicli.py
plirglo: Changed

ansible/playbooks/roles/kafka_exporter/tasks/upgrade/main.yml
plirglo: Done

ansible/playbooks/roles/kafka_exporter/tasks/upgrade/main.yml
plirglo: done

cicharka (approved):
Looks good to me!

seriva (approved):

ansible/playbooks/roles/kafka_exporter/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,102 @@
...
+        src: "{{ download_directory }}/{{ kafka_exporter_file_name }}"
+        dest: "/opt/kafka_exporter"
+        creates: "/opt/kafka_exporter/kafka_exporter"
+        extra_opts: [--strip-components=1]
+        mode: 0755
+        owner: root
+        group: kafka_exporter
+      check_mode: false
+
+    - name: Update service
```
przemyslavic, line 80: From what I can see, we render the system service template from scratch during the upgrade. At the same time, we gave the user the opportunity to set his own configuration by overwriting the defaults (https://github.com/epiphany-platform/epiphany/blob/develop/schema/common/defaults/configuration/kafka-exporter.yml#L8). This means that when upgrading, we will clear all settings and restore it to defaults. Is this an intended action and are we sure we want to do so?

Issue #2750, opened 2021-11-22
[UPGRADE] Upgrade JMX exporter to the newest version


sbbroot:
**Describe the solution you'd like**
JMX exporter needs to be upgraded to the newest version: [0.16.1](https://github.com/prometheus/jmx_exporter/releases/tag/parent-0.16.1)

---

**DoD checklist**

* [x] Changelog updated
* [x] COMPONENTS.md updated / doesn't need to be updated
* [x] Schema updated / doesn't need to be updated
* [x] Feature has automated tests
* [x] Automated tests passed (QA pipelines)
  * [x] apply
  * [x] upgrade
* [x] Idempotency tested
* [x] Documentation added / updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Solution meets requirements and is done according to design doc
* [x] Usage compliant with license
* [x] Backport tasks created / doesn't need to be backported

mkyc:
Spike about versions was in #2719.

przemyslavic:
:heavy_check_mark: Tested `apply` and `upgrade` from v1.0 (x86_64) and from v1.1 (aarch64).

PR #2779, opened 2021-12-09
Upgrade JMX exporter to the newest version (#2750)

cicharka:
* Upgrade JMX exporter to version 0.16.1
* Move prometheus_jmx_path from schema to defaults:
  * prometheus_jmx_path removed from schema for roles: kafka and
    jmx-exporter and moved to defaults for these roles
  * unification of that name, which is holding symlink path for Prometheus JMX exporter
    binary, used by roles: kafka, jmx-exporter, zookeeper
  
  Justification:
  Path of symbolic link to JMX exporter binary should not be managed by
  the user since he is not responsible for binary download and
  configuration of prometheus scrape


ansible/playbooks/roles/zookeeper/defaults/main.yml
```diff
@@ -25,4 +25,4 @@ prometheus_kafka_jmx_file_sd_labels:
 
 
 jmx_exporter_user: jmx-exporter
-jmx_exporter_group: jmx-exporter 
\ No newline at end of file
+jmx_exporter_group: jmx-exporter
```
erzetpe: Add enter.

erzetpe (changes requested):

rafzei (approved):
LGTM

plirglo (approved):

erzetpe (approved):

ansible/playbooks/roles/jmx_exporter/tasks/upgrade/main.yml
```diff
@@ -0,0 +1,101 @@
...
+    - name: Set version facts
+      set_fact:
+        jmx_exporter_version:
+          old: "{{ installed_jmx_exporter_version.stdout }}"
+          new: "{{ jmx_exporter_version }}"
+
+    - name: Avoiding risk of downgrade
+      when: jmx_exporter_version.old is not version( jmx_exporter_version.new, '<' )
+      debug:
+        msg: "Skipping upgrade: JMX Exporter in newer version already installed!"
```
przemyslavic, line 41: @cicharka 
This can be a bit confusing when you have the same version.
After upgrading I have v0.16.1 and on re-run it tells me that `JMX Exporter in newer version is already installed!`.
Actually it's not newer, but the same.
Or the condition should be changed to `<=`.

przemyslavic: Fixed in #2794

Issue #2768, opened 2021-12-01
[FEATURE REQUEST] Add possibility to provide custom hostnames


rafzei:
**Is your feature request related to a problem? Please describe.**
Currently, there is no option to provide a custom VM hostname via epicli.
The hostname is set up referencing inventory hostname.

**Describe the solution you'd like**
Current: `[prefix][clustername]-kubernetes[master/node]vm[x]`
Needed: `[prefix][clustername]-kubernetes[master/node]vm[x].[domain.name]`
but it should allow to set up anything you can imagine :space_invader: 

**Additional context**
Internal team request, if possible prepare a workaround they could apply to their LTS deployment.

Please check `kind: infrastructure/virtual-machine`. Maybe `specification.name` should be the right place for that? Right now it's overwritten after manifests merge.

---

**DoD checklist**

* [ ] Changelog updated
* [ ] COMPONENTS.md updated / doesn't need to be updated
* [ ] Schema updated / doesn't need to be updated
* [ ] Feature has automated tests
* [ ] Automated tests passed (QA pipelines)
  * [ ] apply
  * [ ] upgrade
* [ ] Idempotency tested
* [ ] Documentation added / updated / doesn't need to be updated
* [ ] All conversations in PR resolved
* [ ] Solution meets requirements and is done according to design doc
* [ ] Usage compliant with license
* [ ] Backport tasks created / doesn't need to be backported

seriva:
So this is only for Azure? Or we want this also on AWS?

rafzei:
The fr/fix was requested for Azure but afaik the team also uses AWS.

seriva:
After some more research I found several issues with the task described above:

- For both AWS and Azure the `infrastructure/virtual-machine` document type represents a template for a group of VM`s of a certain component . This template is then being used to generate the Terraform VM definitions on Azure and auto-scaling groups on AWS.  Any VM specific config is generated by the providers:
![image.png](https://images.zenhubusercontent.com/5efee199ee690405efd9d38b/606225df-3931-4fe2-a09d-607cee9d2693)
Adding a mechanism to specify something like a hostname per VM would require some massive changes which will be hard to keep backwards compatible.
[...]

PR #2790, opened 2021-12-14
Added hostname_domain_extension.

seriva:
Implementation for the proposal here: https://github.com/epiphany-platform/epiphany/issues/2768#issuecomment-993822587

Can be reviewed and tested, some documentation will be added after @rafzei  confirms this is ok for the team that requested it.


cli/engine/providers/azure/APIProxy.py
```diff
@@ -59,7 +59,10 @@ def get_ips_for_feature(self, component_key):
         for instance in running_instances:
             if isinstance(instance, list):
                 instance = instance[0]
-            name = instance['virtualMachine']['name']
+            if self.cluster_model.specification.cloud.hostname_domain_extension != '':
+                name = instance['virtualMachine']['name'] + f'.{self.cluster_model.specification.cloud.hostname_domain_extension}'
+            else:
+                name = instance['virtualMachine']['name']
```
atsikham, line 65: Could be something like this, but up to you
```
name = instance['virtualMachine']['name']
if self.cluster_model.specification.cloud.hostname_domain_extension != '':
  name += f'.{self.cluster_model.specification.cloud.hostname_domain_extension}'
```

or

```
name = (instance['virtualMachine']['name'] + f'.{self.cluster_model.specification.cloud.hostname_domain_extension}').rstrip('.')
```

cli/engine/providers/azure/InfrastructureBuilder.py
```diff
@@ -24,6 +24,23 @@ def __init__(self, docs, manifest_docs=[]):
         self.docs = docs
         self.manifest_docs = manifest_docs
 
+        # Check if there is a hostname_domain_extension we want to apply.
+        self.hostname_domain_extension = None
+        if self.cluster_model.specification.cloud.hostname_domain_extension != '':
+            self.hostname_domain_extension = self.cluster_model.specification.cloud.hostname_domain_extension
```
atsikham: Seems to be not needed or just 

```
self.hostname_domain_extension = self.cluster_model.specification.cloud.hostname_domain_extension
```

cli/engine/providers/azure/InfrastructureBuilder.py
```diff
@@ -24,6 +24,23 @@ def __init__(self, docs, manifest_docs=[]):
...
+
+        # Check if there is a hostname_domain_extension we already applied and we want to retain.
+        # The same as VM images we want to preserve hostname_domain_extension over versions.
+        manifest_cluster_model = select_first(self.manifest_docs, lambda x: x.kind == 'epiphany-cluster')
+        if self.manifest_docs:
+            if 'hostname_domain_extension' in manifest_cluster_model.specification.cloud:
+                if manifest_cluster_model.specification.cloud.hostname_domain_extension != '':
```
atsikham: Connected to previous comments, could be removed

atsikham (changes requested):

rafzei:
Works as expected, will give approve after other reviews

cli/engine/providers/azure/APIProxy.py
seriva: Actually your first suggestion was my initial version, but as this for is a weird edge case in behavior I went with the normal if-statement to make it more readable and easier to understand. 

As for you second suggestion, yes I like it, but again not in the context;) 

IMO leave as is.

cli/engine/providers/azure/InfrastructureBuilder.py
seriva: Agreed, we can just use '' for the second check needed inside the vm construction.

cli/engine/providers/azure/InfrastructureBuilder.py
seriva: Cannot as we need to check for hostname preservation. Same as VM images we don't want people to change this value over different versions and/or applies. manifest_cluster_model is coming from the build folder. And as this can be on old cluster which might not have this setting, we need to check for presence, then value, then with the new input.

rafzei (dismissed):

cli/engine/providers/azure/InfrastructureBuilder.py
atsikham: So we can check for presence, but if old value is empty string, why can't we do following steps?

seriva: Yeah you are right, fixing it.

atsikham (dismissed):

rafzei (dismissed):

rafzei:
We have to backport it to LTS

atsikham (approved):

rafzei (approved):

rafzei:
@przemyslavic FYI: I've tested it on Azure/Ubuntu, same APM.

przemyslavic:
@rafzei Thanks a lot!
@seriva Let's merge it 👍

Issue #2748, opened 2021-11-22
[UPGRADE] Upgrade Kafka exporter to v1.4.0


sbbroot:
**Describe the solution you'd like**
Kafka exporter needs to be upgraded to the newest version: [1.4.2](https://github.com/danielqsj/kafka_exporter/releases/tag/v1.4.2)

Update: We have chosen version 1.4.0 since 1.4.1 and 1.4.2 contain critical bug.

**Additional context**
Check if this fixes #1221.

---

**DoD checklist**

* [x] Changelog updated
* [x] COMPONENTS.md updated / doesn't need to be updated
* [x] Schema updated / doesn't need to be updated
* [x] Feature has automated tests
* [x] Automated tests passed (QA pipelines)
  * [x] apply
  * [x] upgrade
* [x] Idempotency tested
* [x] Documentation added / updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Solution meets requirements and is done according to design doc
* [x] Usage compliant with license
* [x] Backport tasks created / doesn't need to be backported

mkyc:
Spike about versions was in #2719.

przemyslavic:
I found a problem with the `/metrics` endpoint not responding is some cases (Prometheus is not able to scrape any data). More investigation in progress.

The issue occurs with new deployments and also with upgraded clusters and seems to be nondeterministic.

to-bar:
> I found a problem with the `/metrics` endpoint not responding is some cases (Prometheus is not able to scrape any data). More investigation in progress.
> 
> The issue occurs with new deployments and also with upgraded clusters and seems to be nondeterministic.

Due to the issue we decided to upgrade to version 1.4.0 instead of 1.4.2 (1.4.1 also affected).

przemyslavic:
✔️ Tested `apply` and `upgrade` from v1.0 (x86_64) and from v1.1 (aarch64).

Additional PR #2796.

PR #2796, opened 2021-12-15
Downgrade kafka_exporter to v1.4.0

plirglo:
Related issue: #2748
Related PR: #2772

We use version 1.4.0 since 1.4.1 and 1.4.2 contain critical bug (see https://github.com/danielqsj/kafka_exporter/issues/273).

### DIFF
--- a/schema/common/defaults/configuration/kafka-exporter.yml
+++ b/schema/common/defaults/configuration/kafka-exporter.yml
@@ -8,10 +8,10 @@ specification:
   config_flags:
     - "--web.listen-address=:9308" # Address to listen on for web interface and telemetry.
     - '--web.telemetry-path=/metrics' # Path under which to expose metrics.
-    - '--verbosity=0'
     - '--topic.filter=.*' # Regex that determines which topics to collect.
     - '--group.filter=.*' # Regex that determines which consumer groups to collect.
     #- '--tls.insecure-skip-tls-verify' # If true, the server's certificate will not be checked for validity. This will make your HTTPS connections insecure.
+    #- '--log.enable-sarama' # Turn on Sarama logging
     - '--kafka.version=2.6.0'
     #- '--sasl.enabled' # Connect using SASL/PLAIN.
     #- '--sasl.handshake' # Only set this to false if using a non-Kafka SASL proxy
--- a/ansible/playbooks/roles/kafka/templates/kafka.service.j2
+++ b/ansible/playbooks/roles/kafka/templates/kafka.service.j2
@@ -1,7 +1,6 @@
 [Unit]
 Description=Kafka Daemon
 After=zookeeper.service
-Wants=kafka-exporter.service
 
 {% if specification.kafka_var.javax_net_debug is defined %}
 {%      set javax_debug = '-Djavax.net.debug=' ~ specification.kafka_var.javax_net_debug %}
--- a/ansible/playbooks/roles/kafka/templates/kafka.service.j2
+++ b/ansible/playbooks/roles/kafka/templates/kafka.service.j2
@@ -1,6 +1,7 @@
 [Unit]
 Description=Kafka Daemon
 After=zookeeper.service
+Wants=kafka-exporter.service
 
 {% if specification.kafka_var.javax_net_debug is defined %}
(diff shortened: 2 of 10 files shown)


przemyslavic:
It didn't help 😞. Metrics endpoint is still unreachable in some cases.
After executing `systemctl restart kafka-exporter.service` it's fine.
```
[ubuntu@ec2-1-1-1-1 ~]$ curl 10.1.5.212:9308/metrics
^C
[ubuntu@ec2-1-1-1-1 ~]$ sudo systemctl restart kafka-exporter.service
[ubuntu@ec2-1-1-1-1 ~]$ curl 10.1.5.212:9308/metrics
# HELP go_gc_duration_seconds A summary of the pause duration of garbage collection cycles.
# TYPE go_gc_duration_seconds summary
go_gc_duration_seconds{quantile="0"} 0
go_gc_duration_seconds{quantile="0.25"} 0
go_gc_duration_seconds{quantile="0.5"} 0
go_gc_duration_seconds{quantile="0.75"} 0
go_gc_duration_seconds{quantile="1"} 0
go_gc_duration_seconds_sum 0
go_gc_duration_seconds_count 0
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 13
# HELP go_info Information about the Go environment.
# TYPE go_info gauge
go_info{version="go1.17.1"} 1
```

schema/common/defaults/configuration/kafka-exporter.yml
```diff
@@ -8,10 +8,10 @@ specification:
   config_flags:
     - "--web.listen-address=:9308" # Address to listen on for web interface and telemetry.
     - '--web.telemetry-path=/metrics' # Path under which to expose metrics.
-    - '--verbosity=0'
```
cicharka, line 11: I saw that in previous version we have been using '--log.level=info', and in version 1.4.2 there was '--verbosity=0', so how about verbosity/log.level config_flag in version 1.4.0 ?

przemyslavic: @plirglo correct me if I'm wrong, but from what I remember, none of these flags works in version 1.4.0 😞

plirglo: Correct, those options don't work in v1.4.0 even if they are specified in documentation :(

atsikham: @plirglo WDYT about adding a comment describing that for future upgrades?

plirglo: You mean comment in this config file? Maybe, but i think we should consider rather to use alternative for kafka exporter since it looks like this project is not maintained very well. We've got issue to find alternative: [#2801](https://github.com/epiphany-platform/epiphany/issues/2801)

cicharka (approved):

schema/common/defaults/configuration/kafka-exporter.yml
atsikham: There is a high probability that it will replaced, but if no, we will forget that.

przemyslavic (approved):

schema/common/defaults/configuration/kafka-exporter.yml
to-bar: IMO `log.level` or `verbosity` isn't something important here. Logs go through systemd-journald so they are rotated. In normal conditions the default level should be fine. When we upgrade version in the future forgetting to add the `verbosity` flag, user will still be able to add it on they own (our validation only requires an array of strings).

atsikham: Adding a note would be faster than writing all these comments 😞

plirglo: @atsikham I'm not a supporter of leaving commented part of code because this might be used in future. W can add this as a known issue with Exporter, WDYT?

atsikham: I'm ok with any case, just see that some other options are already commented.

plirglo: Agree, but these options are possible to use in this version.

Issue #2785, opened 2021-12-13
[FEATURE REQUEST] Add configuration option to Keycloak for PROXY_ADDRESS_FORWARDING env. variable


GrzegorzKlankABB:
**Is your feature request related to a problem? Please describe.**
I would like Keycloak service to be aware of being behind HAPRoxy and inform it to use X-Forwarded-... headers for proper JWT issuer information. At the moment I manually edit keycloak stateful set to include PROXY_ADDRESS_FORWARDING environment variable.

**Describe the solution you'd like**
I would like Epiphany yml configuration file to have flag in auth-service application config that would translate to PROXY_ADDRESS_FORWARDING true/false value passed as environment variable to the keycloak pod. The default value can be set to false to maintain backward compatibility. 

**Describe alternatives you've considered**

**Additional context**

---

**DoD checklist**

* [x] Changelog updated
* [x] COMPONENTS.md updated / doesn't need to be updated
* [x] Schema updated / doesn't need to be updated
* [ ] Feature has automated tests
* [x] Automated tests passed (QA pipelines)
  * [x] apply
  * [ ] upgrade
* [x] Idempotency tested
* [ ] Documentation added / updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Solution meets requirements and is done according to design doc
* [x] Usage compliant with license
* [ ] Backport tasks created / doesn't need to be backported

plirglo:
OS does not matter during testing process since application works in k8s cluster.
Tested on Ubuntu with application setup:
```yaml
kind: configuration/applications
title: Kubernetes Applications Config
name: default
specification:
  applications:
  - name: auth-service
    enabled: true
    image_path: epiphanyplatform/keycloak:14.0.0
    use_local_image_registry: true
    service:
      proxy_address_forwarding: true
```
By default ```proxy_address_forwarding``` is set to ```false```

Configuration deployed:

![image.png](https://images.zenhubusercontent.com/5ef9f41844c19c2f7f1c83e8/e0fa6410-b40c-45b9-ab48-6c116b860931)

![image.png](https://images.zenhubusercontent.com/5ef9f41844c19c2f7f1c83e8/977425b3-d35e-40af-8ca3-9f849ad8ddab)

![image.png](https://images.zenhubusercontent.com/5ef9f41844c19c2f7f1c83e8/ffd836c8-5ae9-42f3-8b12-5a012761fffa)

plirglo:
[...]

PR #2797, opened 2021-12-16
Added PROXY_ADDRESS_FORWARDING for Keycloak

seriva:
Added PROXY_ADDRESS_FORWARDING for Keycloak


ansible/playbooks/roles/applications/templates/auth-service/00-auth-service.yml.j2
```diff
@@ -209,6 +209,10 @@ spec:
               name: "{{ auth_service_name }}-db"
         - name: X509_CA_BUNDLE
           value: "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+{% if data.service.proxy_address_forwarding is defined and data.service.proxy_address_forwarding %}
```
atsikham, line 212: `if data.service.proxy_address_forwarding is defined` - should it be checked if there is alvays a default value?

atsikham (changes requested):

ansible/playbooks/roles/applications/templates/auth-service/00-auth-service.yml.j2
seriva: From Epicli perspective this default should always be there but when we move to the collections as people might chose not to run through CLI.  So I like to cover the bases.

atsikham: default(false) filter could be used, but current version is of for me

atsikham (dismissed):

to-bar (dismissed):

przemyslavic:
@seriva I think it can be merged after the conflict is resolved as it has already been tested by @plirglo https://github.com/epiphany-platform/epiphany/issues/2785#issuecomment-1000101017.

erzetpe (approved):

atsikham (approved):

PR #2800, opened 2021-12-17
Fix postgres nodes assertion

pprach:


cli/engine/ApplyEngine.py
```diff
@@ -140,7 +140,21 @@ def assert_no_postgres_nodes_number_change(self):
...
+                if next_postgres_count == 0:
+                    postgres_role_mapping_exists = False
+
+                    for object in self.configuration_docs:
+                        if object.kind == 'configuration/feature-mapping':
+                            for available_role in object.specification.available_roles:
+                                if (available_role.name == 'postgresql') and (available_role.enabled == True):
+                                    postgres_role_mapping_exists = True
+                                    break;
+                    if postgres_role_mapping_exists:
```
seriva:
```suggestion
                    if not postgres_role_mapping_exists:
                        raise Exception("Postgresql scaling is not supported yet. Please revert your 'postgresql' count to previous value.")     
```

cli/engine/ApplyEngine.py
```diff
@@ -140,7 +140,21 @@ def assert_no_postgres_nodes_number_change(self):
...
+                        pass
+                    else:
+                        raise Exception("Postgresql scaling is not supported yet. Please revert your 'postgresql' count to previous value.")
```
seriva: Maybe put the exception string in a variable instead of twice the same.

cli/engine/ApplyEngine.py
```diff
@@ -123,8 +123,8 @@ def assert_no_master_downscale(self):
                     raise Exception("ControlPlane downscale is not supported yet. Please revert your 'kubernetes_master' count to previous value or increase it to scale up Kubernetes.")
 
     def assert_no_postgres_nodes_number_change(self):
+        self.process_configuration_docs()
```
seriva: We should not process al configuration documents at this point here jet. Check input documents if 'configuration/feature-mapping' is present, if not take from default and maybe merge.

cli/engine/ApplyEngine.py
```diff
@@ -140,7 +140,21 @@ def assert_no_postgres_nodes_number_change(self):
                 prev_postgres_count = len(existing_inventory.list_hosts(pattern='postgresql'))
                 next_postgres_count = int(components['postgresql']['count'])
 
-                if prev_postgres_count != next_postgres_count:
+                # To avoid situation where the postgres is configured on some other host different then postgres
+                if next_postgres_count == 0:
+                    postgres_role_mapping_exists = False
+
+                    for object in self.configuration_docs:
```
seriva: Check comment on previous line, there should be no need to loop over the configuration docs for the 'configuration/feature-mapping' if its taken from default or input docs straight.

seriva:
Closing this one as this was a bit more complicated then just not allowing adding of postgresql role to different components definitions. New pull here:

https://github.com/epiphany-platform/epiphany/pull/2811

Issue #2782, opened 2021-12-10
[BUG] The target namespace for node-exporter is not coherent


romsok24:
**Describe the bug**
The component node-exporter within the k8s daemonset  installation variant ( k8s as cloud service )
is not coherent in the context of target namespace:
 - on one one hand [the component defaults](https://github.com/epiphany-platform/epiphany/blob/develop/ansible/playbooks/roles/node_exporter/defaults/main.yml) are targeting it to the dedicated ns (_epi-monitoring_) 
 - on the other hand the component upgrade procedure (ie. [check for existence task](https://github.com/epiphany-platform/epiphany/blob/develop/ansible/playbooks/roles/upgrade/tasks/node-exporter.yml#:~:text=command%3A%20helm%20list%20%2D%2Doutput%20json) or [uninstall task](https://github.com/epiphany-platform/epiphany/blob/develop/ansible/playbooks/roles/node_exporter/tasks/uninstall-node-exporter-as-daemonset.yml) ) will try to operate on the _default_ ns only 

This will cause the node-exporter component to not being updated upon the epiphany upgrade process.

The bug perceived by @przemyslavic .

**How to reproduce**
A ) install epiphany in the `k8s_as_cloud_service: True` mode
B ) upgrade epiphany 

**Expected behavior**
The node-exporter component should be operated (installed, upgraded) from within the one and only k8s namespace 

**Environment**
Epiphany installation with  k8s daemonset installation enabled ( `k8s_as_cloud_service` configuration parameter set to True )

**epicli version**: [`epicli --version`]
1.3.0dev

---

**DoD checklist**

* [x] Changelog updated (if affected version was released)
* [x] COMPONENTS.md updated / doesn't need to be updated
* [ ] Automated tests passed (QA pipelines)
  * [x] apply
  * [x] upgrade
* [ ] Case covered by automated test (if possible)
* [x] Idempotency tested
* [x] Documentation updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [ ] Backport tasks created / doesn't need to be backported

przemyslavic:
The purpose of these tasks was to uninstall node exporter from the default namespace (it was installed there at the beginning) and to install it in dedicated `epi-monitoring`. It was a kind of fix. Now, when upgrading, we should take into account both variants, that node exporter can be in default namespace or dedicated.
This is exactly what task name says:
[...]

Issue #2747, opened 2021-11-22
[UPGRADE] Upgrade Node exporter to v1.3.1


sbbroot:
**Describe the solution you'd like**
Node exporter needs to be upgraded to the newest version: [1.3.1](https://github.com/prometheus/node_exporter/releases/tag/v1.3.1)

---

**DoD checklist**

* [ ] Changelog updated
* [x] COMPONENTS.md updated / doesn't need to be updated
* [x] Schema updated / doesn't need to be updated
* [x] Feature has automated tests
* [x] Automated tests passed (QA pipelines)
  * [x] apply
  * [x] upgrade
* [x] Idempotency tested
* [x] Documentation added / updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Solution meets requirements and is done according to design doc
* [x] Usage compliant with license
* [x] Backport tasks created / doesn't need to be backported

mkyc:
Spike about versions was in #2719.

romsok24:
What was done:

- requrements* file updated for every distro  ( details in [the branch](https://github.com/epiphany-platform/epiphany/tree/feature/upgr-node-eporter-2747) )
- changes tested with epicli on ubu18
  - run the: epicli test -b ....... -g <component ie. node_exporter>
  - tested epicli upgrade -b build/dir/ (node_exporter is not in the scope of epicli upgrade command )
- [the PR](https://github.com/epiphany-platform/epiphany/pull/2761) was created ans passed for running the test build pipelines of updated code.

romsok24:
@sbbroot Are we ok to upgrade to the currently newest version which is already not the one described in this issue and [here](https://github.com/epiphany-platform/epiphany/issues/2719)?

sbbroot:
@romsok24 Yes.

przemyslavic:
✔️ Tested `apply` and `upgrade` from v1.0 (x86_64) and from v1.1 (aarch64).

Upgrade of Node Exporter as a DaemonSet is tracked in #2782.

PR #2807, opened 2021-12-21
Node Exporter in daemonset mode is now upgraded in relevant NS #2782

romsok24:
### PR scope
Node Exporter ( when installed in daemonset mode) is now being upgraded in its relevant kubernetes name space.
Related to [#2782](https://github.com/epiphany-platform/epiphany/issues/2782) issue.

 ### Additional remarks
Additionally [CHANGELOG.md file](https://github.com/epiphany-platform/epiphany/blob/feature/upgr-node-eporter-inanyns-2782/docs/changelogs/CHANGELOG-1.3.md#:~:text=Upgrade%20Node%2DExporter%20to%20v1.3.1) was updated by the info about the changes implemented within [#2747](https://github.com/epiphany-platform/epiphany/issues/2747) as was reminded [here](https://github.com/epiphany-platform/epiphany/issues/2782#issuecomment-998596068).


seriva (dismissed):

docs/changelogs/CHANGELOG-1.3.md
```diff
@@ -4,6 +4,7 @@
 
 ### Added
 
+- [#2782](https://github.com/epiphany-platform/epiphany/issues/2782#issuecomment-998596068) - Assure Node-Exporter is upgraded in its relevant namespace
```
atsikham: Correct link

docs/changelogs/CHANGELOG-1.3.md
```diff
@@ -32,6 +33,7 @@
 
 ### Updated
 
+- Upgrade Node-Exporter to v1.3.1
```
atsikham: Was there a task for that? Some components are listed here without a task as they were updated in the scope of k8s upgrade and there was no a separate task

ansible/playbooks/roles/upgrade/tasks/node-exporter.yml
```diff
@@ -18,30 +18,37 @@
...
         - node_exporter_release_exists
       block:
-        - name: Include uninstall task for Node Exporter as DaemonSet in default namespace for "k8s as cloud service"
+        - name: "Include uninstall task for Node Exporter as DaemonSet within {{ node_exporter_release_ns }} ns for k8s as cloud service mode"
           include_role:
             name: node_exporter
             tasks_from: uninstall-node-exporter-as-daemonset.yml
 
-        - name: Include install task for Node Exporter as DaemonSet in dedicated namespace for "k8s as cloud service"
+        - name: "Include install task for Node Exporter as DaemonSet in {{ node_exporter_release_ns }} ns for k8s as cloud service mode"
```
atsikham: here and above - capital letter for K8s

ansible/playbooks/roles/upgrade/tasks/node-exporter.yml
```diff
@@ -18,30 +18,37 @@
     - k8s_as_cloud_service
   block:
     - name: Get information about helm releases
-      command: helm list --output json
+      command: helm list -A --output json
```
atsikham: Change to --all-namespaces

ansible/playbooks/roles/upgrade/tasks/node-exporter.yml
```diff
@@ -18,30 +18,37 @@
...
         _names: >-
           {{ helm_list.stdout | from_json
                               | map(attribute='name')
                               | select('==', exporter_defaults.node_exporter_helm_chart_name)
                               | list }}
+        _namespace: >-
+          {{ helm_list.stdout | from_json
+                              | map(attribute='namespace')
+                              | list
+                              | first }}
```
atsikham: Is it ok here not to search for necessary namespace but just to take the first one? What's about removal `_names` var, getting `_namespace` the same way with select and then use `_namespace` in `node_exporter_release_exists`?

atsikham (changes requested):

ansible/playbooks/roles/upgrade/tasks/node-exporter.yml
romsok24: ad 1) Seems to be efficient enough as the returned value is of list type and taking into consideration that we install node-exporter only in single ns.

ad2 )  We need to set both variables as the `_names` var ( so consequently `node_exporter_release_exists` ) allows us to find out the helm installation entity which is not possible when do the same but using `_namespace' var ( as You've proposed ) in the situation of upgrade combined with the ns move. It would only work in the scenario of just version upgrade.

docs/changelogs/CHANGELOG-1.3.md
romsok24: Yeap - exactly as it was stated in Additional remarks chapter in this PR.
I didn't put the task no along with the task name (_Upgrade Node-Exporter to v1.3.1_) because I just wanted to stay aligned with the other elements in this section.

atsikham: Add a task id/link and move a bit below to have a common format

ansible/playbooks/roles/upgrade/tasks/node-exporter.yml
atsikham: 1. Have you tested the case when not only node-exporter is installed by helm and there are a few releases?
2. If a namespace found, doesn't it mean that node-exporter release exists and vice versa?

atsikham (changes requested):

ansible/playbooks/roles/upgrade/tasks/node-exporter.yml
romsok24: The remarks were taken on board within [this commit](https://github.com/epiphany-platform/epiphany/pull/2807/commits/5f575f515e7a519252a891e3585a0fc8ff329efe).

docs/changelogs/CHANGELOG-1.3.md
romsok24: The remarks were taken on board within [this commit](https://github.com/epiphany-platform/epiphany/pull/2807/commits/95e08dbfdcdfe64730425287770d4dbdb8547f22).

ansible/playbooks/roles/upgrade/tasks/node-exporter.yml
```diff
@@ -18,30 +18,32 @@
...
           {{ helm_list.stdout | from_json
-                              | map(attribute='name')
-                              | select('==', exporter_defaults.node_exporter_helm_chart_name)
-                              | list }}
+                              | map(attribute='namespace')
+                              | list
+                              | first }}
 
-    - name: Reinstall Node Exporter as DaemonSet from default namespace to dedicated namespace for "k8s as cloud service"
+    - name: "Reinstall Node Exporter as DaemonSet within ns for K8s as cloud service mode"
```
atsikham: Did you want to mention namespace here or remove 'ns' as block names are not displayed in logs? If there are no placeholders, quoting is not necessary.

atsikham (changes requested):

ansible/playbooks/roles/node_exporter/tasks/uninstall-node-exporter-as-daemonset.yml
```diff
@@ -1,9 +1,9 @@
...
 
-- name: Uninstall Helm release "{{ node_exporter_helm_chart_name }}"
+- name: "Uninstall Helm release {{ node_exporter_helm_chart_name }} from {{ node_exporter_release_ns }} ns"
 
   delegate_to: localhost
   become: false
   run_once: true
 
-  command: helm uninstall {{ node_exporter_helm_chart_name }}
+  command: helm uninstall -n {{ node_exporter_release_ns }} {{ node_exporter_helm_chart_name }}
```
atsikham: What's about quoting here?

ansible/playbooks/roles/upgrade/tasks/node-exporter.yml
```diff
@@ -18,30 +18,32 @@
...
           {{ helm_list.stdout | from_json
-                              | map(attribute='name')
-                              | select('==', exporter_defaults.node_exporter_helm_chart_name)
-                              | list }}
+                              | map(attribute='namespace')
+                              | list
+                              | first }}
 
-    - name: Reinstall Node Exporter as DaemonSet from default namespace to dedicated namespace for "k8s as cloud service"
+    - name: Reinstall Node Exporter as DaemonSet within for K8s as cloud service mode
```
atsikham:
```suggestion
    - name: Reinstall Node Exporter as DaemonSet for K8s as cloud service mode
```

atsikham (changes requested):

atsikham (dismissed):

seriva (dismissed):

atsikham:
Tested on AWS/Ubuntu with `k8s_as_cloud_service: true`, disabled `node-exporter` feature mapping (not to install as the system service) and copied kubeconfig to build directory.

✅  Data seems to be in place.

image

✅  Helm chart version and app version are updated successfully from epicli v1.0, verified with another chart installed.

Before:

➜  epicli git:(feature/upgr-node-eporter-inanyns-2782) ✗ helm list -A
NAME         	NAMESPACE     	REVISION	UPDATED                              	STATUS  	CHART              	APP VERSION
aaa          	default       	1       	2021-12-28 12:07:54.7313194 +0000 UTC	deployed	jenkins-3.10.2     	2.319.1    
node-exporter	epi-monitoring	1       	2021-12-28 10:40:51.513357 +0000 UTC 	deployed	node-exporter-1.1.2	1.0.1  

After:

➜  epicli git:(feature/upgr-node-eporter-inanyns-2782) ✗ helm list -A
NAME         	NAMESPACE     	REVISION	UPDATED                              	STATUS  	CHART               	APP VERSION
aaa          	default       	1       	2021-12-28 12:07:54.7313194 +0000 UTC	deployed	jenkins-3.10.2      	2.319.1    
node-exporter	epi-monitoring	1       	2021-12-28 14:22:18.6141643 +0000 UTC	deployed	node-exporter-2.3.17	1.3.1

✅  Created/updated #2821 task.

🟡  No CI pipelines checked, as discussed with @przemyslavic there are no pipelines for that case.

@romsok24 **Following things were found:**

1. Task `Get information about helm releases`: need to add `changed_when: false`.
2. Task `Check if node exporter release exist in default namespace and set fact`: the name is wrong, namespace is not always default.
3. Task `Uninstall Helm release node-exporter from epi-monitoring ns`: always executed, even when version is the same and upgrade is not required.
4. Task `Install Node-Exporter's Helm chart ...`: now always has `changed` status, but when the change above is applied, should be skipped together with the task from p.3.

romsok24:
> 1. Task `Get information about helm releases`: need to add `changed_when: false`.
> 3. Task `Uninstall Helm release node-exporter from epi-monitoring ns`: always executed, even when version is the same and upgrade is not required.
> 4. Task `Install Node-Exporter's Helm chart ...`: now always has `changed` status, but when the change above is applied, should be skipped together with the task from p.3.

As I 've noticed, the similar code is being used also in [other roles](https://github.com/epiphany-platform/epiphany/blob/develop/ansible/playbooks/roles/upgrade/tasks/filebeat.yml) so probably we should create a new issue ( after you're done with testing this role ) as your comments can be applicable also in these other places.

atsikham:
> > 1. Task `Get information about helm releases`: need to add `changed_when: false`.
> > 2. Task `Uninstall Helm release node-exporter from epi-monitoring ns`: always executed, even when version is the same and upgrade is not required.
> > 3. Task `Install Node-Exporter's Helm chart ...`: now always has `changed` status, but when the change above is applied, should be skipped together with the task from p.3.
> 
> As I 've noticed, the similar code is being used also in [other roles](https://github.com/epiphany-platform/epiphany/blob/develop/ansible/playbooks/roles/upgrade/tasks/filebeat.yml) so probably we should create a new issue ( after you're done with testing this role ) as your comments can be applicable also in these other places.

You're right, but the first 2 points is a quick change.

atsikham (dismissed):

atsikham:
@romsok24 could you create an issue you mentioned in [this](https://github.com/epiphany-platform/epiphany/pull/2807#issuecomment-1002443657) comment?

romsok24:
> could you create an issue 
HYA:  https://github.com/epiphany-platform/epiphany/issues/2827

romsok24:
> 3. Task `Uninstall Helm release node-exporter from epi-monitoring ns`: always executed, even when version is the same and upgrade is not required.
> 4. Task `Install Node-Exporter's Helm chart ...`: now always has `changed` status, but when the change above is applied, should be skipped together with the task from p.3.

@atsikham, [this commit](https://github.com/epiphany-platform/epiphany/pull/2807/commits/54f051d5f6c54e8a16c587dd8352a0f6cfb9a49a) answers your remark from p3 and p4.
However it does not cover the scenario when the versions ( both current and the expected one ) are the same but - on the other hand - the current chart installation namespace is the default namespace ( which was probably the idea of the code before ).

I'm not sure however if it's the real  deal breaker....

ansible/playbooks/roles/upgrade/tasks/node-exporter.yml
```diff
@@ -17,31 +17,52 @@
...
+      debug:
+        msg:
+          - "Installed version: {{ node_exporter_current_version }}"
+          - "Target version: {{ exporter_defaults.exporter.version }}"
+
+    - name: Reinstall Node Exporter as DaemonSet for K8s as cloud service mode
       when:
         - node_exporter_release_exists
+        - exporter_defaults.exporter.version != node_exporter_current_version or
+          node_exporter_release_ns == "default"
```
atsikham: `exporter_defaults.monitoring_chart_namespace != node_exporter_release_ns` would be better without hardcoded value

docs/changelogs/CHANGELOG-1.3.md
```diff
@@ -43,6 +44,7 @@
   - Upgrade Flannel to v0.14.0
   - Upgrade Kubernetes dashboard to v2.3.1
   - Upgrade Kubernetes metrics-scraper to v1.0.7
+- [#2747](https://github.com/epiphany-platform/epiphany/issues/2747) - Upgrade Node-Exporter to v1.3.1
```
atsikham, line 47: Could you place 2 new lines close to each other as they are related to similar topic?

atsikham (changes requested):
Almost LGTM

atsikham (approved):

ansible/playbooks/roles/upgrade/tasks/node-exporter.yml
romsok24: Done within this [change](https://github.com/epiphany-platform/epiphany/pull/2807/commits/5d44e643724a93f48dad5ccf3b9df9e064ace0f4). 
Successfully tested with scenario of moving this chart from default to dedicated ns.

seriva (approved):

PR #2810, opened 2021-12-21
Docs improvement

romsok24:
This PR is basically about improving Epiphany docs or just making them a bit more accurate 
from the perspective of the person that is a newbie to the project.
This can allow for better understanding the prerequisites needed to start the Epiphany implementation.


atsikham (approved):

rafzei (approved):
why not

PR #2811, opened 2021-12-22
Fix postgres nodes assertion

seriva:


cli/engine/ApplyEngine.py
```diff
@@ -123,25 +124,30 @@ def assert_no_master_downscale(self):
...
-                next_postgres_count = int(components['postgresql']['count'])
-
-                if prev_postgres_count != next_postgres_count:
-                    raise Exception("Postgresql scaling is not supported yet. Please revert your 'postgresql' count to previous value.")
+            prev_postgres_node_count = len(existing_inventory.list_hosts(pattern='postgresql'))
+            postgres_available = [x for x in feature_mapping.specification.available_roles if x.name == 'postgresql']
+            if postgres_available[0].enabled:
+                for key, roles in feature_mapping.specification.roles_mapping.items():
+                    if ('postgresql') in roles:
+                        if key in components:
```
atsikham: May be combine them?

atsikham (changes requested):

plirglo (dismissed):

cli/engine/ApplyEngine.py
seriva: Good point.

plirglo (approved):

atsikham (approved):

atsikham:
✅  Checked different combinations of scaling up/down and feature mapping changes for apply. Everything works as expected - there is an error message in case of PostgreSQL nodes number change.

przemyslavic:
@seriva I think it can be merged as it has already been tested by @atsikham.

PR #2822, opened 2021-12-29
Added better validation for hostname_extension.

seriva:


rafzei (approved):

schema/common/validation/epiphany-cluster.yml
```diff
@@ -122,7 +122,7 @@ properties:
         default: ''
         examples:
         - extension
-        pattern: "^(.*)$"
+        pattern: "^[a-zA-Z0-9.-]{0,20}$"
```
atsikham, line 125: Do we support capital letters in hostname? https://github.com/epiphany-platform/epiphany/blob/dd972cda8ff0f77cf609d7e7051cf66dfe3768d8/schema/any/validation/infrastructure/machine.yml#L12

atsikham (changes requested):

schema/common/validation/epiphany-cluster.yml
seriva: Technically this is a domain extension, as in somehostname.thispart. I have seen from time to time the use of capitals there but don't think its part of official spec.

atsikham: So we leave it as it is?

seriva: Also Azure does not have not have any issues with this: 
image

seriva: For me we can leave it as is.

atsikham (approved):

schema/common/validation/epiphany-cluster.yml
atsikham: Yes, this is because that validation is for 'any' provider.

przemyslavic:
@seriva Dots are also not allowed.
I will check it more deeply, please wait a second.

OK now I know why it failed for me :)
```
Re-applying a different hostname_domain_extension might lead to data loss and/or other issues. Preserving the previous hostname_domain_extension
```
So probably it's OK. Checking again.

atsikham:
> @seriva Dots are also not allowed. I will check it more deeply, please wait a second.

But there can be names like `ec2-54-161-153-125.compute-1.amazonaws.com`

seriva:
> > @seriva Dots are also not allowed. I will check it more deeply, please wait a second.
> 
> But there can be names like `ec2-54-161-153-125.compute-1.amazonaws.com`

Why I left them, and I think that there even was an example from testing that @rafzei  showed me somewhere.

przemyslavic:
@seriva I would leave it as it is, although it's not perfect ;)
We can have 8 characters for the prefix and 20 for the cluster name, so for the opendistro component we have the pattern `12345678-12345678901234567890-opendistro-for-elasticsearch-vm-0`, which is 63 characters long without the domain name, and the full name cannot exceed 64 characters. The validation will pass, but Azure will return an error. This is more complex to have all scenarios covered by validation so I think we can leave it as it is if you agree.
In addition, from what I can see, Azure has removed the dots and made it lowercase:
`hostname_domain_extension: ..-----890TEST.`
```
[operations@ci-dmnazurrhelcanal-load-balancer-vm-0 ~]$ hostnamectl
Static hostname: ci-dmnazurrhelcanal-load-balancer-vm-0.-----890test
```

przemyslavic (approved):

atsikham (approved):

PR #2829, opened 2021-12-31
Fix edge case in K8s total memory assertion

atsikham:
PR to allow 'apply' on 2 GB/GiB RAM machines and have the [exact](https://github.com/kubernetes/kubernetes/blob/13e97453f9024f347a40940f27630e62c4540fa5/cmd/kubeadm/app/constants/constants.go#L366) value as Kubeadm has.


erzetpe (approved):

romsok24 (approved):
LGTM 
A specially can be useful in the test environments with lower amount of mem on the host machine.

atsikham:
> LGTM A specially can be useful in the test environments with lower amount of mem on the host machine.

In dev environment you can disable that 🌝

Issue #2866, opened 2022-01-13
[BUG] Postgres Exporter upgrade broken - no vars generated during upgrade


przemyslavic:
**Describe the bug**
Postgres exporter upgrade fails on task `[postgres_exporter : Include specification vars from postgres_exporter role]`:
```
2022-01-13T18:37:24.2525347Z[38;21m18:37:24 INFO cli.engine.ansible.AnsibleCommand - TASK [postgres_exporter : Include specification vars from postgres_exporter role] ***
2022-01-13T18:37:24.4826409Z[31;21m18:37:24 ERROR cli.engine.ansible.AnsibleCommand - fatal: [ec2-1-1-1-1.eu-west-1.compute.amazonaws.com]: FAILED! => {"ansible_facts": {"exporter_vars": {}}, "ansible_included_var_files": [], "changed": false, "message": "Could not find or access 'roles/postgres_exporter/vars/main.yml'
Searched in:
\t/shared/build/12to13armawcecalico/ansible/roles/postgres_exporter/vars/roles/postgres_exporter/vars/main.yml
\t/shared/build/12to13armawcecalico/ansible/roles/postgres_exporter/roles/postgres_exporter/vars/main.yml
\t/shared/build/12to13armawcecalico/ansible/roles/postgres_exporter/tasks/upgrade/vars/roles/postgres_exporter/vars/main.yml
\t/shared/build/12to13armawcecalico/ansible/roles/postgres_exporter/tasks/upgrade/roles/postgres_exporter/vars/main.yml
\t/shared/build/12to13armawcecalico/ansible/vars/roles/postgres_exporter/vars/main.yml
\t/shared/build/12to13armawcecalico/ansible/roles/postgres_exporter/vars/main.yml on the Ansible Controller.
If you are using a module and expect the file to exist on the remote, see the remote_src option"}

```
The reason for this is that in unrelated PR #2819 postgres exporter has been removed from roles with defaults 😢 
https://github.com/epiphany-platform/epiphany/pull/2819/files#diff-c347f852c48b4116fd8f4282377ec7c782e8301b32a6650e9ea8072ed40237f4L72
Thanks @cicharka for investigating the issue.

**How to reproduce**
1. deploy v1.2 cluster with PostgreSQL component enabled
3. execute `epicli upgrade` to v1.3

**Expected behavior**
Upgrade should be successful

**Config files**

**Environment**
- Cloud provider: [all]
- OS: [all]

[...]

PR #2868, opened 2022-01-13
Return back postgres_exporter defaults generation

atsikham:
See description in #2866


seriva:
Is this something needed for 1.3 as well? If so it needs to added to the 1.3 branch @przemyslavic created.

cicharka (approved):

seriva (approved):

PR #2921, opened 2022-01-25
[v1.0] Sync .gitignore with develop branch

to-bar:


.gitignore
```diff
@@ -1,73 +1,129 @@
...
+.dmypy.json
+dmypy.json
+
+# Pyre type checker
+.pyre/
+
+# pycharm settings
+.idea/
+
+# epicli specific
```
seriva, line 122: Some if these will be core/src/... in old Epicli < 1.3

seriva (changes requested):

.gitignore
seriva: And I mean the lines below ># epicli specific ;)

to-bar: This rule is wide, i.e. applies to any `.idea` sub-directory including `core/src/.idea`.

seriva (approved):

atsikham (approved):

Issue #2847, opened 2022-01-05
[FEATURE REQUEST] Upgrade Ansible to 5.2.0


to-bar:
**Is your feature request related to a problem? Please describe.**
If there is Python 2 & 3 at the same time, ansible-core 2.11 will choose Python 2 by default. Due to Ubuntu release upgrade we've changed setting `interpreter_python` from `auto_legacy_silent` to `auto` in ansible.cfg. However, this config file may by not applied when running playbooks manually (for developing) unless `ANSIBLE_CONFIG` is set. ansible-core 2.12 solves this issue since it uses different default value for `interpreter_python` (`auto`).

**Describe the solution you'd like**
Ansible in latest version.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

---

**DoD checklist**

* [x] Changelog updated
* [x] COMPONENTS.md updated / doesn't need to be updated
* [x] Schema updated / doesn't need to be updated
* [x] Feature has automated tests
* [x] Automated tests passed (QA pipelines)
  * [x] apply
  * [x] upgrade
* [x] Idempotency tested
* [x] Documentation added / updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Solution meets requirements and is done according to design doc
* [x] Usage compliant with license
* [x] Backport tasks created / doesn't need to be backported

asciible:
waiting for https://app.zenhub.com/workspaces/epiphany-platform-5ef9d3c749afc20019782391/issues/epiphany-platform/epiphany/2852 to be merged for component docs update

przemyslavic:
Tested scenarios:

:heavy_check_mark: `apply` + :x: re-apply x86_64 (fails on Ubuntu #2940)
:heavy_check_mark: `apply` + re-apply aarch64
:heavy_check_mark: `upgrade` from v1.0 x86_64
:heavy_check_mark: `upgrade` from v1.2 aarch64
:heavy_check_mark: `epicli init`
:heavy_check_mark: `epicli prepare`
:heavy_check_mark: `epicli delete`
:heavy_check_mark: `epicli test`
:x: `epicli backup` #2942
:x: `epicli recovery` #2942
:heavy_check_mark: offline installation x86_64
:heavy_check_mark: offline installation aarch64

Issues reported:
#2940
#2942

PR #2933, opened 2022-01-26
Update Ansible to 5.2 and Python to 3.10

asciible:


cli/licenses.py
```diff
@@ -1681,16 +1639,5 @@
...
-        "License text": "MIT License\n\nCopyright (c) [year] [fullname]\n\nPermission is hereby granted, free of charge, to any person obtaining a copy\nof this software and associated documentation files (the \"Software\"), to deal\nin the Software without restriction, including without limitation the rights\nto use, copy, modify, merge, publish, distribute, sublicense, and/or sell\ncopies of the Software, and to permit persons to whom the Software is\nfurnished to do so, subject to the following conditions:\n\nThe above copyright notice and this permission notice shall be included in all\ncopies or substantial portions of the Software.\n\nTHE SOFTWARE IS PROVIDED \"AS IS\", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR\nIMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,\nFITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE\nAUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER\nLIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,\nOUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE\nSOFTWARE.\n"
     }
 ]
```
erzetpe: Please, add enter at the end of file.

.devcontainer/pyproject.toml
```diff
@@ -5,15 +5,15 @@ description = "Epiphany Dev Container"
...
+python = "^3.10"
 pyyaml = "*"
 jinja2 = "*"
 boto3 = "*"
 jsonschema = "*"
 python-json-logger = "*"
-ansible = "==2.10.7"
 "ruamel.yaml" = "*"
 azure-cli = "^2.32.0"
+ansible = "^5.2.0"
```
erzetpe: Maybe this may be be more specific, like using tilde instead of caret?

erzetpe (changes requested):

cli/licenses.py
seriva: This is automatically generated each time we update it, so lets leave as is.

.devcontainer/pyproject.toml
seriva: Agreed, use '''poetry add ansible@5.2.0''' to fix it in place please. We only want to update this periodically, not every time we update the dependencies.

asciible: Do we want to lock python to 3.10 as well? It was specified as ^3.7

seriva: For me its ok, as this is coming from the container anyway which is fixed. What I did see is that I also did not lock the azure-cli version when I did the terraform update. Could you pls also remove and add azure-cli in again in this pull locking it to 2.32.0?

`poetry add azure-cli@2.32.0`

cli/licenses.py
erzetpe: We need to fix this and replace the line:  licenses_content = textwrap.dedent(licenses_content) + json.dumps(all_deps_data, indent=4) in file: .devcontainer/gen-dependency-info.py with for example:  licenses_content = f"{textwrap.dedent(licenses_content)}{json.dumps(all_deps_data, indent=4)}\n". This enter at the end of file is there for a reason.

.devcontainer/pyproject.toml
erzetpe: This can be a good idea to lock this to python 3.10 as this may be additional safety net.

seriva: So lets lock it.

atsikham:
Changelog

erzetpe (dismissed):

docs/changelogs/CHANGELOG-2.0.md
```diff
@@ -17,6 +17,8 @@
   - Azurerm provider 1.38.0 to 2.91.0
   - AWS provider 2.26 to 3.71.0
   - Upgraded Azure-cli 2.29 to 2.32
+  - Ansible-core 2.10 to 2.12 (Ansible 5.2.0)
```
seriva: Just name it Ansible whater to 5.2.0 and the entries under the root issue #2847

cli/licenses.py
seriva: @asciible pls consider adding the extra new line fix suggested by @erzetpe

erzetpe (approved):

seriva (approved):

PR #2938, opened 2022-01-28
Moved version file to root of the project

seriva:
Moved version.txt.py to the root of the project in VERSION file.  version.txt.py made sense when we packaged the project in a Python package but it doesn't anymore with the new project structure introduced in 1.3.


rafzei (approved):
:one:

erzetpe (approved):

asciible (approved):

przemyslavic (approved):

Issue #2833, opened 2022-01-04
[FEATURE REQUEST] Removal of Logstash component


erzetpe:
**Is your feature request related to a problem? Please describe.**
The functionality that Logstash is serving in Epiphany - exporting csv can right now be achieved using Open Distro for Elasticsearch plugin.

**Describe the solution you'd like**
We want to remove unused Logstash component as it double the functionality with Open Distro for Elasticsearch plugin.

**Describe alternatives you've considered**
None.

**Additional context**
Logstash in this version is prone to log4j issue and this is also the reason why should be removed.

---

**DoD checklist**

* [x] Changelog updated
* [x] COMPONENTS.md updated / doesn't need to be updated
* [x] Schema updated / doesn't need to be updated
* [ ] Feature has automated tests
* [x] Automated tests passed (QA pipelines)
  * [x] apply
  * [ ] upgrade
* [x] Idempotency tested
* [x] Documentation added / updated / doesn't need to be updated
* [x] All conversations in PR resolved
* [x] Solution meets requirements and is done according to design doc
* [x] Usage compliant with license
* [x] Backport tasks created / doesn't need to be backported

seriva:
Important is to remove everything:

- Ansible
- Python (if present)
- Defaults
- Validation
- Documentation
- etc

From documentation only leave things that are part of design docs.

rafzei:
Hi @cicharka, shouldn't you also remove spec test for logstash? see [here](https://github.com/epiphany-platform/epiphany/blob/develop/tests/spec/spec/filebeat/filebeat_spec.rb#L57-L68)

cicharka:
Hello @rafzei, as I understand it, the test you pointed to is related to `filebeat` component.
Unfortunately name 'logstash' was not only used for `logstash` component name, but also as username for `elasticsearch` [here](https://github.com/epiphany-platform/epiphany/blob/66e11f3e2e3d3be750952cc0e010af26b04ddc69/ansible/playbooks/roles/opendistro_for_elasticsearch/tasks/configure-es.yml#L217) and also in `filebeat` [here](https://github.com/epiphany-platform/epiphany/blob/2b12e87535df2d45c84b9308bc242b0e913f31af/ansible/playbooks/roles/filebeat/templates/filebeat.yml.j2#L315).
Please correct me if I wrong @przemyslavic or point out who can verify it :)

rafzei:
[...]

PR #2941, opened 2022-02-03
Removal of Logstash component - next steps #2833

cicharka:
* Add information about manual package removal that
  needs to be performed by cluster admin

### DIFF
--- a/docs/home/DEPRECATION-NOTE.md
+++ b/docs/home/DEPRECATION-NOTE.md
@@ -35,12 +35,6 @@ Issue: [2833](https://github.com/epiphany-platform/epiphany/issues/2833)
 
 ## Next steps
 
-If Logstash is installed in your cluster, and you plan Epiphany upgrade from v1.x to v2.x, it is recommended to remove Logstash from your machines.
-
-**RedHat clusters**:
-1. Check if Logstash package is installed: `sudo yum info logstash-oss`
-2. Remove: `sudo yum remove logstash-oss`
-
-**Debian clusters**:
-1. Check if Logstash package is installed: `sudo apt --installed list | grep logstash-oss`
-2. Remove: `sudo apt-get remove logstash-oss`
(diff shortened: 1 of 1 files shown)


przemyslavic:
Maybe worth mentioning to remove the `logstash` group from inventory if it is there.

docs/home/DEPRECATION-NOTE.md
```diff
@@ -32,3 +32,15 @@ The functionality that the Logstash is serving in Epiphany - exporting csv can r
...
+
+If Logstash is installed in your cluster, and you plan Epiphany upgrade from v1.x to v2.x, it is recommended to remove Logstash from your machines.
+
+**RedHat clusters**:
+1. Check if Logstash package is installed: `sudo yum info logstash-oss`
+2. Remove: `sudo yum remove logstash-oss`
+
+**Debian clusters**:
+1. Check if Logstash package is installed: `sudo apt --installed list | grep logstash-oss`
+2. Remove: `sudo apt-get remove logstash-oss`
```
atsikham: Should we describe the way of packages removal or just list them? For me these commands look obviously.

atsikham (changes requested):

docs/home/DEPRECATION-NOTE.md
cicharka: I talked with @to-bar about packages removal, and like you can do it in at least 2 ways (the other ones are rpm and dpkg, according to the OS), they don't provide any logging for that action (e.g. yum history, or /var/log/apt/history.log). I don't know how much do we care about it, but I can either add comment about it in doc, or just leave part `If Logstash is installed in your cluster, and you plan Epiphany upgrade from v1.x to v2.x, it is recommended to remove Logstash from your machines.` What do you think?

erzetpe (approved):

atsikham (approved):

Issue #2945, opened 2022-02-03
[BUG] epicli apply sleeps 10 seconds after creating inventory


to-bar:
**Describe the bug**
We have unconditional sleep call in epicli code:
https://github.com/epiphany-platform/epiphany/blob/c687e8eca0347fda6dd7a0b092deafeccdb53fb6/cli/engine/ansible/AnsibleRunner.py#L109

**How to reproduce**
1. execute `epicli apply ...`

**Expected behavior**
Run without wasting time for sleeping or wait for something conditionally (explicitly) as long as needed instead of static 10 seconds.

**Environment**
- Any

**epicli version**: 1.3.0 and older

**Additional context**
n/a

---

**DoD checklist**

- Changelog
  - [x] updated
  - [ ] not needed
- COMPONENTS.md
  - [ ] updated
  - [x] not needed
- Schema
  - [ ] updated
  - [x] not needed
- Backport tasks
  - [ ] created
  - [x] not needed
- Documentation
  - [ ] added
  - [ ] updated
  - [x] not needed
- [ ] Feature has automated tests
- [ ] Automated tests passed (QA pipelines)
  - [x] apply
  - [x] upgrade
  - [ ] backup/restore
- [ ] Idempotency tested
- [x] All conversations in PR resolved

przemyslavic:
Tested together with #2944.

Issue #2944, opened 2022-02-03
Refactor InitEngine class to be agnostic to changes in ApplyEngine and UpgradeEngine


to-bar:
**Is your feature request related to a problem? Please describe.**
We have temporary workarounds here:
https://github.com/epiphany-platform/epiphany/blob/c687e8eca0347fda6dd7a0b092deafeccdb53fb6/cli/engine/InitEngine.py#L63

One was added in PR #2861

**Describe the solution you'd like**
Be able to add options for `apply` and `upgrade` mode without breaking `init` mode.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

---

**DoD checklist**

- Changelog
  - [x] updated
  - [ ] not needed
- COMPONENTS.md
  - [ ] updated
  - [x] not needed
- Schema
  - [ ] updated
  - [x] not needed
- Backport tasks
  - [ ] created
  - [x] not needed
- Documentation
  - [ ] added
  - [ ] updated
  - [x] not needed
- [ ] Feature has automated tests
- [ ] Automated tests passed (QA pipelines)
  - [x] apply
  - [x] upgrade
  - [ ] backup/restore
- [ ] Idempotency tested
- [x] All conversations in PR resolved
- [ ] Solution meets requirements and is done according to design doc
- [x] Usage compliant with license

przemyslavic:
Tested `init`, `init --full`, `apply` and `upgrade` commands with `--ping-retries` and `--ansible-forks` flags.
Reported 2 minor issues (not related to these changes, but related to the init engine ):
#2966
#2968

PR #2948, opened 2022-02-04
Fixes for #2944 and #2945

seriva:
- Refactor InitEngine class to be agnostic to changes in ApplyEngine and UpgradeEngine: #2944
- Removed epicli apply sleeps 10 seconds after creating inventory: #2945
- Rename __ping_retries to ping_retries for consistency.
- Abstract loading of inventory done in a few places to build_io.py
- Minor fixes in dump function.

@przemyslavic Attached are the original init outputs for every provider, full and non-full to compare against:

[any-full.txt](https://github.com/epiphany-platform/epiphany/files/8001903/any-full.txt)
[any-minimal.txt](https://github.com/epiphany-platform/epiphany/files/8001906/any-minimal.txt)
[aws-full.txt](https://github.com/epiphany-platform/epiphany/files/8001907/aws-full.txt)
[aws-minimal.txt](https://github.com/epiphany-platform/epiphany/files/8001908/aws-minimal.txt)
[azure-full.txt](https://github.com/epiphany-platform/epiphany/files/8001909/azure-full.txt)
[azure-minimal.txt](https://github.com/epiphany-platform/epiphany/files/8001910/azure-minimal.txt)

Note that the any-full.txt will be slightly different because its also merged now, rest should be all the same.


cicharka (dismissed):

cli/helpers/build_io.py
```diff
@@ -53,6 +56,11 @@ def save_inventory(inventory, cluster_model, build_dir=None):
     save_to_file(file_path, content)
 
 
+def load_inventory(inventory_path):
+    inventory = InventoryManager(loader=DataLoader(), sources=inventory_path)
```
sbbroot: Just return InventoryManager in this line, no need for variable.

cli/engine/ApplyEngine.py
```diff
@@ -39,7 +36,7 @@ def __init__(self, input_data):
         self.infrastructure_docs = []
         self.manifest_docs = []
 
-        self.__ping_retries: int = input_data.ping_retries
+        self.ping_retries: int = input_data.ping_retries
```
sbbroot, line 39: Is there any reason for this variable to be public? I don't see any direct usage of it.

seriva: I would agree with you, however the __ paradigm is currently not used  inside the CLI code. So for me this is currently just about consistency. New stuff like the dependency code you are working on or PoC work we should use it and when time allows add it in bulk to the current CLI code as well.

sbbroot (approved):

cicharka (approved):

przemyslavic (approved):

PR #2956, opened 2022-02-08
K8s upgrade tasks refactoring

atsikham:


cicharka (dismissed):

seriva:
@atsikham Would it make sense to also move the upgrade code to the actual Kubernetes component? As we have been doing for some of the other ones in previous upgrades already.

atsikham:
> @atsikham Would it make sense to also move the upgrade code to the actual Kubernetes component? As we have been doing for some of the other ones in previous upgrades already.

I think yes, but this part is quite large and there are a few roles related to K8s, so for me the target structure is not clear. @to-bar started doing so in the scope of a separate collection, so I think these changes should be also consistent

seriva (dismissed):

przemyslavic (approved):

cicharka (approved):

erzetpe (approved):

PR #2967, opened 2022-02-09
CLI refactor to make it more consistent 

### LINKED ISSUE #2968: [BUG] `epicli init` should generate `specification.cloud.subscription_name` for minimal cluster config.
**Describe the bug**
When executing `epicli init` for Azure provider there is no `specification.cloud.subscription_name` key in the output file.
It means that I'm not able just to run `apply` after `init`, but I have to manually add `subscription_name` field, or run `init --full` to have full config available. But then what's the point of using `epicli init` since the configuration is insufficient for a cluster deployment.

**How to reproduce**
1. execute `epicli init -p azure -n test`
2. adjust it to your needs
3. execute `epicli apply`

**Expected behavior**
Minimal cluster config should contain minimum dataset for cluster deployment.
IMO `epicli init` should generate exactly what we have in https://github.com/epiphany-platform/epiphany/blob/develop/schema/azure/defaults/epiphany-cluster.yml - to be agreed.

Note: Fix should be applied to all providers.

This may be done together with #2966.

**Environment**
- Cloud provider: [Azure]
- OS: [all]

**epicli version**: [`2.0.0dev`]

**Additional context**

---

**DoD checklist**

- Changelog
  - [ ] updated
  - [ ] not needed
- COMPONENTS.md
  - [ ] updated
  - [ ] not needed
- Schema
  - [ ] updated
  - [ ] not needed
- Backport tasks
  - [ ] created
  - [ ] not needed
- Documentation
  - [ ] added
  - [ ] updated
  - [ ] not needed
- [ ] Feature has automated tests
- [ ] Automated tests passed (QA pipelines)
  - [ ] apply
  - [ ] upgrade
  - [ ] backup/restore
- [ ] Idempotency tested
- [ ] All conversations in PR resolved

seriva:
- Re-named **epiphany/cli/engine** to **epiphany/cli/src**
- Moved epiphany/cli/models to **epiphany/cli/src/models**
- Moved epiphany/cli/helpers to **epiphany/cli/src/helpers**
- Moved the **CommandsEngine.py** inside **epiphany/cli/src/commands**
- Move **Log.py, Step.py, Config.py** to **epiphany/cli/src**
- Re-named the **CommandsEngine.py** files and classes to **Commands.py**
- Sorted all imports in all files
- Consolidated some of the helper functions
- Fix up all tests
- Also fixes #2968


cli/src/helpers/data_loader.py
```diff
@@ -6,12 +6,12 @@
...
 
-from cli.helpers.yaml_helpers import safe_load_all, safe_load
-from cli.helpers.objdict_helpers import dict_to_objdict
+from cli.src.helpers.yaml_helpers import safe_load_all, safe_load
+from cli.src.helpers.objdict_helpers import dict_to_objdict
 
 
 BASE_DIR_PROD = '/epicli'
-BASE_DIR_DEV = os.path.join(os.path.dirname(__file__), '../../')
+BASE_DIR_DEV = os.path.join(os.path.dirname(__file__), '../../../')
```
sbbroot, line 13: I don't think using relative paths is a good idea. How about defining `BASE_DIR_DEV` in one of the files inside `cli/src` dir and import it here? You would not need to add '../' then.
For example put it in config.py and move config.py to `cli/src`. I think that's a good idea for these reasons:
- Config is not a helper, it's an essential class
- Config does not need separate directory
- No matter how you change the file layout `src/` will always be one directory below the base dir.

sbbroot (changes requested):

cicharka (dismissed):
LGTM

cli/src/helpers/data_loader.py
seriva: Agreed about moving Config.py. 

When it comes to BASE_DIR_DEV we could also just make it fixed as BASE_DIR_PROD as these are just the locations in which the epiphany content lives. For production container its just in /epicli:

image

For our development container it used to be the case that VSCode would generate some random hash folder if I remember correctly. But that has changed already a long time ago so we could just use: /workspaces/epiphany

image

For me no need to move these constants for now.

seriva: Regarding Config.py, we can probably move Log.py and Step.py as well to /cli/src

sbbroot (dismissed):

cicharka (dismissed):

cicharka (dismissed):

schema/azure/defaults/configuration/minimal-cluster-config.yml
```diff
@@ -9,6 +9,7 @@ specification:
     name: operations # YOUR-ADMIN-USERNAME
     key_path: /shared/.ssh/epiphany-operations/id_rsa # YOUR-SSH-KEY-PATH
   cloud:
+    subscription_name: YOUR-SUB-NAME
```
cicharka, line 12: Add Changelog note for that fix

cicharka (dismissed):

przemyslavic (approved):

cicharka (approved):

PR #2972, opened 2022-02-10
Review tools and remove outdated ones

atsikham:
Removed:

- service_user_disable_enable, documentation adjusted. Alternative: `ansible <hosts> -m ansible.builtin.shell -a 'usermod -s /sbin/nologin <user>'`
- file_hash_generator, there were no references in the doc
- ip_change, there were no references in the doc
- patch_cgroup_driver: done in latest versions automatically, contained deprecated K8s configuration, references checked

Memory load check tools were not removed to have an ability to adjust kubelet [settings](https://github.com/epiphany-platform/epiphany/blob/2e56f032792ea9d5c29816fb38284d8f68f594c9/ansible/playbooks/roles/kubernetes_master/defaults/main.yml).


plirglo:
What about ip_change scripts? None of those scripts are valid or we agreed to remove all of them?

atsikham:
> What about ip_change scripts? None of those scripts are valid or we agreed to remove all of them?

Forgot to mention that, but there are also no references to these scripts, that's why it was removed. If you know someone who uses that, I can revert. In general this looks like it shouldn't be a part of this repository.

plirglo:
> 

I've confirmed with client for whom it was created and they don't use it. It can be removed.

plirglo (approved):

seriva (approved):

PR #2992, opened 2022-02-24
Fix swap entry removal from `/etc/fstab`

plirglo:


ansible/playbooks/roles/common/tasks/main.yml
```diff
@@ -25,11 +25,11 @@
     - sysctl
 
 - name: Remove swap from /etc/fstab
-  mount:
-    backup: true
-    fstype: swap
-    path: swap
+  lineinfile:
+    path: /etc/fstab
+    regexp: ^((?!#).*\sswap\s.*\d\s+\d$)
```
to-bar:
```suggestion
    regexp: ^(?!#)\s*(?P<fs_spec>\S+)\s+(?P<fs_file>\S+)\s+(?P<fs_vfstype>swap)\s.*$
```
Longer version but with named groups since `fstab` has static format:
https://manpages.org/fstab/5

to-bar (approved):

seriva (approved):

PR #3023, opened 2022-03-15
Add RHEL upgrade playbook

asciible:


ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+          - name: rhel-7-server-extras-rpms
+            repo_file: /etc/yum.repos.d/redhat.repo
+
+    reboot:
+      connect_timeout: 5 # maximum seconds to wait for a successful connection to the managed hosts before next trial
+      update_timeout: 500
+      upgrade_timeout: 3600 # Long wait required by Leapp utility service
+
+  tasks:
+    ### PRE-UPGRADE ###
```
erzetpe: Maybe all of the tasks from pre-upgrade can be taken into separate block and the code can be separated in blocks reflecting the intent of each piece? Can be also split in separate files to be shorter and more readable?

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+        src: /etc/leapp/files/leapp-data15.tar.gz
+        dest: /etc/leapp/files/
+        remote_src: true
+
+    # Address upgrade inhibitors
+    - name: Disable ssh root login
+      lineinfile:
+        dest: /etc/ssh/sshd_config
+        regexp: '^PermitRootLogin'
+        line: 'PermitRootLogin no'
```
erzetpe: Isn't there any module for sshd config in ansible to change settings?

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+          vars:
+            query: "results[?yumstate=='installed'].envra"
+
+        - name: Remove only duplicates
+          when: installed_kernel_devel_packages | length > 1
+          yum:
+            name: "{{ installed_kernel_devel_packages[1:] }}"
+            state: absent
+
+    ### UPGRADE ###
```
erzetpe: Maybe it will be good to move upgrade to separate file?

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+        - pata_acpi
+        - floppy
+
+    - name: Keep only one kernel-devel package when multiple are present
+      block:
+        - name: List duplicated kernel-devel packages
+          yum:
+            list: kernel-devel
+          register: kernel_devel_packages
+
```
erzetpe: In the block there are too many spaces for me, maybe the layout can be more tight?

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+        - name: Register version of installed packages
+          set_fact:
+            installed_kernel_devel_packages: "{{ kernel_devel_packages|json_query(query) }}"
+          vars:
+            query: "results[?yumstate=='installed'].envra"
+
+        - name: Remove only duplicates
+          when: installed_kernel_devel_packages | length > 1
```
erzetpe: Is this certain that installed_kernel_devel_packages will be always defined?

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+          yum:
+            name: "{{ installed_kernel_devel_packages[1:] }}"
```
erzetpe: Is this certain that installed_kernel_devel_packages will be always defined?

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+      command:
+        cmd: "leapp upgrade {{ '--no-rhsm' if provider != 'non_cloud' }}"
+
+    - name: Reboot system to complete leapp upgrade procedure
+      reboot:
+        msg: Reboot initiated by Ansible due to major release upgrade
+        connect_timeout: "{{ reboot.connect_timeout }}"
+        reboot_timeout: "{{ reboot.upgrade_timeout }}"
+
+    ### POST-UPGRADE -- CLEAN-UP AND VERIFICATION ###
```
erzetpe: Maybe this can be moved to separate file?

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+    - name: Verify post-upgrade state
+      block:
+        - name: Gather post-upgrade facts
+          setup:
+            gather_subset: "!all"
+
+        - name: Assert major version is as expected
+          assert:
+            that:
+              - ansible_distribution_version is version('8.4','=')
```
erzetpe: Cannot be this also 8.5? @to-bar has requested that images on Azure and AWS should be on RHEL 8.5.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+          loop: "{{ old_kernel_versions.stdout_lines }}"
+
+        - name: Remove old kernels from bootloader entries
+          when: not old_kernel_versions.failed
+          command: "/bin/kernel-install remove {{ item }} /lib/modules/{{ item }}/vmlinuz"
+          loop: "{{ old_kernel_versions.stdout_lines }}"
+
+    - name: Restore root login only if it was originally enabled
+      when: sshd_config.changed
+      lineinfile:
```
erzetpe: Is there a module to do this?

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+        dest: /etc/ssh/sshd_config
+        regexp: '^PermitRootLogin no'
+        line: 'PermitRootLogin yes'
+        state: present
+
+    - *RESTART_SSHD
```
erzetpe: What is this?

erzetpe (changes requested):

ci/ansible/playbooks/os/redhat/upgrade_release.yml
asciible: *RESTART_SSHD executes &RESTART_SSHD defined in line 140

ci/ansible/playbooks/os/redhat/register_subscription.yml
```diff
@@ -0,0 +1,19 @@
+# Provide Redhat subscription credentials (username and password) for
+# RHSM registered systems (e.g. vagrant) via extra-vars file
+---
+- name: Force subscription registration and auto-attach
+  command: "subscription-manager register --force --auto-attach --username {{ username }} --password {{ password }}"
```
to-bar: When `--force` option is used it's a good practice to add a comment with reason.

ci/ansible/playbooks/os/redhat/register_subscription.yml
```diff
@@ -0,0 +1,19 @@
...
+
+- name: Register subscription status
+  command: "subscription-manager status"
+  register: "subscription_status"
+  ignore_errors: true # May return code 1 even when correctly registered
```
to-bar: Please do not use `ignore_errors` if a better alternative exists.
Is there any specific message when rc == 1?
We can use `failed_when`.

ci/ansible/playbooks/os/redhat/register_subscription.yml
```diff
@@ -0,0 +1,19 @@
...
+
+- name: Assert current and known subscription status/system purpose
+  assert:
+    that:
+      - "'Overall Status: Current' in subscription_status.stdout"
+      - "'System Purpose Status: Unknown' not in subscription_status.stdout"
+    fail_msg: >
+      "Registration has been rejected, check the availability
+       of sufficient entitlements in your Red Hat subscription"
```
to-bar:
```suggestion
      Registration has been rejected, check the availability
      of sufficient entitlements in your Red Hat subscription
```

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
+# Based on https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html-single/upgrading_from_rhel_7_to_rhel_8/index
+# and partially on https://docs.microsoft.com/en-us/azure/virtual-machines/workloads/redhat/redhat-in-place-upgrade
+# Limitations: ansible connection as root is not possible,
+# root login via ssh is an Inhibitor of release upgrade using LEAPP
```
to-bar:
```suggestion
```
I would remove this comment since:

> Title: Possible problems with remote login using root account
> Summary: OpenSSH configuration file does not explicitly state the option PermitRootLogin in sshd_config file, which will default in RHEL8 to "prohibit-password".
> Remediation: [hint] If you depend on remote root logins using passwords, consider setting up a different user for remote administration or adding "PermitRootLogin yes" to sshd_config.

source: https://access.redhat.com/solutions/6139192

In CI pipeline we will use  `ec2-user` / `operations` user which is added to sudoers.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+---
+- name: In-place RHEL release upgrade
+  hosts: "{{ target | default('all') }}"
+  become: true
+  connection: paramiko # Ensures connection stability during leapp utility execution
```
to-bar: Do we need paramiko?
1) In inventory file we use `ansible_ssh_extra_args='-o ServerAliveInterval=60 -o ServerAliveCountMax=1'`. Have you tried them without paramiko?
2) In Ansible [docs](https://docs.ansible.com/ansible/latest/user_guide/connection_details.html#controlpersist-and-paramiko) they say "If your control machine uses an older version of OpenSSH that does not support ControlPersist, Ansible will fallback to a Python implementation of OpenSSH called ‘paramiko’." If 1) doesn't help, maybe we could try other SSH options like [described here](https://medium.com/analytics-vidhya/ansible-tips-and-tricks-part-3-66e44b77038a).

to-bar (commented):
Please rename `redhat` sub-directory to `rhel`. We're going to use RedHat name for OS family only and RHEL/AlmaLinux for disto to reduce ambiguity.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
to-bar: This playbook is meant to be run in CI so I would exceptionally vote for limiting amount of files since they will be downloaded from GH using curl to avoid cloning whole repo.

ci/ansible/playbooks/os/redhat/leapp_metadata.yml
```diff
@@ -0,0 +1,11772 @@
+$ANSIBLE_VAULT;1.1;AES256
+37373839313765333931663935386630383835303064633931396365613830343961396634306531
+3935333430373638623061316437376636363838613863340a323731336435636538653562356265
+39363033373735346639383132643565633035306563616635333831353166343664633063643566
+3663323937363161340a633462616365313863663639313937346462666535316264366162643534
+39343563373461646265616136376334303830366134616239663030323639396235373234326263
+31306233646139313463333138366164616432323162386465376264653064373439373762326430
```
erzetpe: We were discussing about this with @to-bar and we have doubts if this should be even in encrypted form in our repository because of the licensing. Probably we should remove this?

erzetpe (dismissed):

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+    ### PRE-UPGRADE ###
+    - name: Assert initial major version
+      assert:
+        that: ansible_distribution_major_version == "7"
+        fail_msg: "Release upgrade not possible in current major version"
+        quiet: true
+
+    - name: Detect and set provider type
+      block:
+        - name: Register presence of distinctive AWS repository file
```
to-bar: Instead of two sequential tasks you can consider using just one, for example run command `yum repolist -q` then check whether distinctive repo is present in the output.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+          register: aws_repo
+
+        - name: Register presence of distinctive Azure repository file
+          stat:
+            path: "{{ leapp_dependencies.repos.azure.0.repo_file }}"
+          register: azure_repo
+
+        - name: Set provider
+          set_fact:
+            provider: "{% if aws_repo.stat.exists %}aws{% elif azure_repo.stat.exists %}azure{% else %}non_cloud{% endif %}"
```
to-bar: Using Jinja directly here is IMO overkill, preferred version would be (just example):
```
provider: >-
  {{ 'aws' if aws_repo.stat.exists else
     'azure' if azure_repo.stat.exists else
     'non_cloud' }}
```

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+
+    - name: Ensure Redhat subcription on non-cloud hosted systems
+      when: provider == 'non_cloud'
+      include_tasks: register_subscription.yml
+
+    - name: Register SELinux state
```
to-bar: Why not use `ansible_selinux` fact?
See https://www.reddit.com/r/ansible/comments/6jo1xe/checking_for_selinux/

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+    - name: Install packages that provide the leapp utility
+      package:
+        name: "{{ leapp_dependencies.packages[provider] }}"
+        state: present
+
+    # The following leapp_metadata archive is tied to a subscription,
+    # thus it has been encrypted to avoid licensing conflicts
+    - name: Create leapp metadata archive
+      copy:
+        content: "{{ metadata_archive | b64decode }}"
```
to-bar: I think we can provide this file in different way - without adding it to public GH repo. Maybe add it to Kubernetes on the build cluster?

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+            list: kernel-devel
+          register: kernel_devel_packages
+
+        - name: Register version of installed packages
+          set_fact:
+            installed_kernel_devel_packages: "{{ kernel_devel_packages|json_query(query) }}"
+          vars:
+            query: "results[?yumstate=='installed'].envra"
+
+        - name: Remove only duplicates
```
to-bar, line 168: Does this task always remove the oldest versions keeping the latest?

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+          when: installed_kernel_devel_packages | length > 1
+          yum:
+            name: "{{ installed_kernel_devel_packages[1:] }}"
+            state: absent
+
+    ### UPGRADE ###
+    - name: Provide leapp answer about pam_pkcs11_module removal
+      command:
+        cmd: "leapp answer --section remove_pam_pkcs11_module_check.confirm=True --add"
```
to-bar:
```suggestion
      command: leapp answer --section remove_pam_pkcs11_module_check.confirm=True --add
```

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+
+    - name: Start leapp upgrade
+      command:
+        cmd: "leapp upgrade {{ '--no-rhsm' if provider != 'non_cloud' }}"
```
to-bar:
```suggestion
      command: leapp upgrade {{ '--no-rhsm' if provider != 'non_cloud' }}
```

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+        value: ''
+
+    - name: Remove leapp_resume service unit
+      block:
+        - name: Disable leapp_resume service # This unit may remain even when the leapp packages are removed
+          systemd:
+            name: leapp_resume
+            state: stopped
+            enabled: false
+            daemon-reload: true
```
to-bar: Do we need this just after stopping and disabling service?

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+            force: true
+
+        - name: Remove leapp_resume unit file # Package removal does not remove this units automatically
```
to-bar: I would first remove leapp packages then clean up left-overs.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+      file:
+        path: /etc/leapp
+        state: absent
+
+    - name: Remove remaining RHEL 7 packages
+      block:
+        - name: Register remaining RHEL7 packages
+          shell: "rpm -qa | grep -e '\\.el7' | grep -vE '^(gpg-pubkey|libmodulemd|katello-ca-consumer)'"
+          register: remaining_rhel7_packages
+          ignore_errors: true
```
to-bar, line 258: This is bad practice.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+
+        - name: Remove remaining RHEL7 packages
+          when: not remaining_rhel7_packages.failed
+          dnf:
+            name: "{{ remaining_rhel7_packages.stdout_lines }}"
```
to-bar, line 263: Can we specify list of packages here directly using patterns with `{{ package_name }}*.el7*`?

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+          when: not old_kernel_versions.failed
+          shell: "[ -x /usr/sbin/weak-modules ] && /usr/sbin/weak-modules --remove-kernel {{ item }}"
+          loop: "{{ old_kernel_versions.stdout_lines }}"
+
+        - name: Remove old kernels from bootloader entries
+          when: not old_kernel_versions.failed
+          command: "/bin/kernel-install remove {{ item }} /lib/modules/{{ item }}/vmlinuz"
+          loop: "{{ old_kernel_versions.stdout_lines }}"
+
+    - name: Restore root login only if it was originally enabled
```
to-bar: If I'm not wrong the upgrade only requires `PermitRootLogin` to be defined and both options are acceptable. So we only need to ensure that without restoring - please use `no` when undefined.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+        - provider == "non_cloud"
+        - "'Extra Packages for Enterprise Linux 7' in pre_upgrade_active_repositories.stdout"
+      block:
+        - name: Remove epel7 repository
+          package:
+            name: epel-release
+            state: absent
+
+        - name: Install epel8 repository
+          command: "yum -y install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm"
```
to-bar: Can we use `dnf` module?
Do we need remove epel-release first or it would be just upgraded (replaced) by epel-release-latest-8.noarch.rpm?

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+          register: cloud_init_cache
+
+        - name: Remove cloud-init cache files
+          file:
+            path: "{{ item.path }}"
+            state: absent
+          loop: "{{ cloud_init_cache.files }}"
+
+    - name: Restart cloud-init service # Azure /mnt mount point is cloud-init dependent
+      when: provider == 'azure'
```
to-bar: The same condition is used multiple times sequentially - please use block for grouping.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
```diff
@@ -0,0 +1,378 @@
...
+
+    # SELINUX configuration restore
+    - name: Re-enable SELinux
+      when:
+        - not pre_upgrade_selinux_status.stdout == "disabled"
+        - pre_upgrade_selinux_mode == "Enforcing"
+      ansible.posix.selinux:
+        state: enforcing
+
+    - name: Remove cloud-init cache # Initially populated with python2 dependencies that are unmet in RHEL8
```
to-bar, line 321: Could we do that before rebooting? Then maybe the `Remount azure specific cloud-init filesystem` or `Restart temp-disk-dataloss-warning service` tasks would be not needed.

to-bar (changes requested):

ci/ansible/playbooks/os/redhat/upgrade_release.yml
to-bar: Refactored. We decided to have everything in one file to easily download it in CI pipeline - just with simple `curl` instead of `git clone`.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
to-bar: It seems there is only module for SSH client config (`community.general.ssh_config`). For sshd I found only external role in Galaxy (https://galaxy.ansible.com/willshersystems/sshd) so we keep `lineinfile`.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
to-bar: Refactored.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
to-bar: Refactored.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
to-bar: Already answered.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
to-bar: Added additional tasks.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
to-bar: Already answered.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
to-bar: Paramiko is not needed when Epiphany inventory is used.

ci/ansible/playbooks/os/redhat/leapp_metadata.yml
to-bar: Removed from repo, provided via Azure File share to build agents.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
to-bar: Refactored.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
to-bar: Refactored.

ci/ansible/playbooks/os/redhat/upgrade_release.yml
to-bar: Refactored.

to-bar (approved):

PR #3031, opened 2022-03-18
Fix element selection from ansible_mounts

atsikham:


przemyslavic:
1. changelog
2. this also has to be fixed: https://github.com/epiphany-platform/epiphany/blob/develop/ansible/playbooks/roles/postgresql/tasks/preflight/upgrade.yml#L116

przemyslavic (approved):

plirglo (approved):

PR #3070, opened 2022-04-08
Fix for #3065

### LINKED ISSUE #3065: [BUG][AZURE] Flag `delete_os_disk_on_termination` has no effect when removing cluster
**Describe the bug**
In configuration of virtual-machines there is flag `delete_on_termination` set to false, but it doesn't have effect during cluster removal (`epicli delete`) and all of the OS disks are removed anyway. Also, when using similar flag for data disk (`delete_data_disks_on_termination`), it has no effect.

**How to reproduce**
1. execute `epicli apply ...`
3. execute `epicli delete ...` on that build

**Expected behavior**
When setting `delete_on_termination` flag to false (which is default) OS disks should not be removed during `epicli delete`.

**Config files**
Any config

**Environment**
- Cloud provider: Azure 
- OS: Ubuntu 20.04.3 LTS

**epicli version**: [`epicli --version`]
- 2.0.0dev
- all of the previous versions that use that flag

**Additional context**
On AWS this flag can preserve disks from being removed when removing cluster with `epicli delete`.
According to [this comment](https://github.com/hashicorp/terraform-provider-azurerm/issues/1701#issuecomment-409870513) it is because deletion of whole Resource Group.

---

**DoD checklist**

- Changelog
  - [x] updated
  - [ ] not needed
- COMPONENTS.md
  - [ ] updated
  - [x] not needed
- Schema
  - [x] updated
  - [ ] not needed
- Backport tasks
  - [x] created
  - [ ] not needed
- Documentation
  - [x] added
  - [ ] updated
  - [ ] not needed
- [ ] Feature has automated tests
- [x] Automated tests passed (QA pipelines)
  - [x] apply
  - [ ] upgrade
  - [ ] backup/restore
- [ ] Idempotency tested
- [x] All conversations in PR resolved

cicharka:
* fixes #3065
* update changelog, docs and message prompt before destroy procedure
* update schema and terraform templates

### DIFF
--- a/cli/epicli.py
+++ b/cli/epicli.py
@@ -237,7 +237,9 @@ def delete_parser(subparsers):
     sub_parser._action_groups.append(optional)
 
     def run_delete(args):
-        if not query_yes_no('Do you really want to delete your cluster?'):
+        if not query_yes_no('''You are trying to delete your cluster.
+If your configuration does not allow to keep existing disks used in the cluster, you will lose your data.
+Make sure your data is safe. Do you really want to delete your cluster?'''):
             return 0
         adjust_paths_from_build(args)
         with Delete(args) as cmd:
--- a/cli/src/commands/Delete.py
+++ b/cli/src/commands/Delete.py
@@ -1,4 +1,5 @@
 from cli.src.helpers.build_io import delete_directory, load_manifest
+from cli.src.helpers.cli_helpers import query_yes_no
 from cli.src.helpers.doc_list_helpers import select_single
 from cli.src.Step import Step
 from cli.src.terraform.TerraformRunner import TerraformRunner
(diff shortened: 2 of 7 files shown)


seriva (dismissed):

cli/epicli.py
```diff
@@ -237,7 +237,9 @@ def delete_parser(subparsers):
     sub_parser._action_groups.append(optional)
 
     def run_delete(args):
-        if not query_yes_no('Do you really want to delete your cluster?'):
+        if not query_yes_no('You are trying to delete your cluster. '
+                            'If your configuration does not allow to keep existing disks used in the cluster, you will lose your data. '
+                            'Make sure your data is safe. Do you really want to delete your cluster?'):
```
przemyslavic: Maybe let's add some new lines here? Now the message is quite long and a bit unreadable.
```
epiuser@a48587a15ad1:/shared$ epicli delete -b /shared/build/testdelete
You are trying to delete your cluster. If your configuration does not allow to keep existing disks used in the cluster, you will lose your data. Make sure your data is safe. Do you really want to delete your cluster? [Y/n]
```

cicharka: Sure, I will improve it.

cicharka: Is it good enough?
```
vscode@792f1ffdfe3e:/workspaces/azure-disk-removal-b1$ epicli delete -b clusters/rook-v6/
You are trying to delete your cluster.
If your configuration does not allow to keep the existing disks used in the cluster, you will lose your data.
Make sure your data is safe. Do you really want to delete your cluster? [Y/n] 
```

przemyslavic (approved):

erzetpe (approved):

PR #3076, opened 2022-04-11
[2.0.1] Allow temporary credentials (session token parameter)

atsikham:
[Examples](https://docs.aws.amazon.com/cli/latest/reference/sts/get-session-token.html#examples).


erzetpe:
Please add validation, changelog and documentation.

seriva:
@atsikham I think we can move this to Ready to review?

to-bar:
How about using full names in specification?

```
specification:
  cloud:
    credentials:
      key: <access_key_id>
      secret: <secret_access_key>
      token: <session_token>
```
vs:
```
specification:
  cloud:
    credentials:
      access_key_id: <access_key_id>
      secret_access_key: <secret_access_key>
      session_token: <session_token>
```
IMO this would better correspond to AWS env variables: `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`

rafzei:
I think target branch should be https://github.com/epiphany-platform/epiphany/tree/v2.0 (with this changelog)

seriva:
> I think target branch should be https://github.com/epiphany-platform/epiphany/tree/v2.0 (with this changelog)

Not for current 2.0.0 release anymore. Once we do 2.0.1 will probably just do a merge from dev to v2.0 branch as there will be some more changes.

atsikham:
> How about using full names in specification?
> 
> ```
> specification:
>   cloud:
>     credentials:
>       key: <access_key_id>
>       secret: <secret_access_key>
>       token: <session_token>
> ```
> 
> vs:
> 
> ```
> specification:
>   cloud:
>     credentials:
>       access_key_id: <access_key_id>
>       secret_access_key: <secret_access_key>
>       session_token: <session_token>
> ```
> 
> IMO this would better correspond to AWS env variables: `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_SESSION_TOKEN`

Parameters are renamed, but it was decided not to backport this change not to broke current configurations.

rafzei (dismissed):

docs/changelogs/CHANGELOG-2.0.md
```diff
@@ -81,5 +87,6 @@
 - Upgrade of Terraform components in issue [#2825](https://github.com/epiphany-platform/epiphany/issues/2825) and [#2853](https://github.com/epiphany-platform/epiphany/issues/2853) will make running re-apply with infrastructure break on existing 1.x clusters. The advice is to deploy a new cluster and migrate data. If needed a manual upgrade path is described [here.](../home/howto/UPGRADE.md#terraform-upgrade-from-epiphany-1.x-to-2.x)
 - Kubernetes container runtime changed. Dockershim and Docker are no longer on Kubernetes hosts.
 - Filebeat `docker` input replaced by `container` input. New field provided for Filebeat as system service installation: `container.id`. Field `kubernetes.container.name` is no longer valid.
+- AWS credentials configuration parameters are renamed from `specification.cloud.credentials.key` and `specification.cloud.credentials.secret` to `specification.cloud.credentials.access_key_id` and `specification.cloud.credentials.secret_access_key`.
```
to-bar: Since this feature is for 2.0.1 shouldn't we move this info to `[2.0.1]` section?

seriva: Agreed, this willl be 2.0.1 so it can be moved there.

atsikham: Good catch, but is it ok to do such renaming between 2.0.0 and 2.0.1?

seriva: Yeah in this case I am ok with it as we know that 2.0.0 will not be used in that sense. As I posted yesterday on main 2.0.1 and 2.0.2 will be the versions which are actually gonna be used down the road.

schema/common/validation/epiphany-cluster.yml
```diff
@@ -79,19 +79,24 @@ properties:
...
+          secret_access_key:
             "$id": "#/properties/specification/properties/cloud/properties/credentials/properties/secret"
             type: string
             title: The Secret Schema
             pattern: "^(.*)$"
+          session_token:
+            "$id": "#/properties/specification/properties/cloud/properties/credentials/properties/token"
+            type: string
+            title: The session token
+            pattern: "^(.*)$"
```
to-bar: How about changing this pattern to not allow spaces?
Then we could avoid `credentials.session_token.strip()` in `APIProxy.py`.

https://stackoverflow.com/questions/41917270/use-regex-in-json-schema-to-validate-a-string-for-no-spaces

seriva: @atsikham ?

to-bar (dismissed):

erzetpe (dismissed):

seriva (approved):

cicharka (approved):

PR #3078, opened 2022-04-12
Fix issue with `allow_mismatch: false` flag

sbbroot:
* An exception will be raised when `allow_mismatch` flag is set to `false`
  while downloading images

### DIFF
--- a/ansible/playbooks/roles/repository/files/download-requirements/src/mode/base_mode.py
+++ b/ansible/playbooks/roles/repository/files/download-requirements/src/mode/base_mode.py
@@ -98,7 +98,7 @@ class BaseMode:
                             requirement_file: Path,
                             sha_algorithm: str) -> bool:
         """
-        Check if checksum matches with `requirement` and downloaded image `requirement_file`.
+        Check if checksum matches with `requirement` and downloaded file `requirement_file`.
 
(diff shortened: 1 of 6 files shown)


ansible/playbooks/roles/repository/files/download-requirements/src/mode/base_mode.py
```diff
@@ -131,21 +161,33 @@ def __download_files(self, files: Dict[str, Dict], dest: Path):
...
 
-                logging.info(f'- {file}')
-                self._download_file(file, filepath)
+                if not self.__is_checksum_valid(files, req_file, tmpfile, 'sha256'):
+                    tmpfile.unlink()
+                    raise ChecksumMismatch(f'- {req_file}')
 
-                if files[file]['sha256'] != get_sha256(filepath):
-                    raise ChecksumMismatch(f'- {file}')
+                move(str(tmpfile), str(filepath))
```
to-bar: Can we avoid moving from tmp by default to not slow down?

przemyslavic (approved):

seriva (approved):

erzetpe (approved):

rafzei (approved):

PR #3100, opened 2022-04-22
OWASP ZAP scans

Harshini097t:
Web Application Vulnerability Scanners are automated tools that scan web applications, normally from the outside, to look for security vulnerabilities such as Cross-site scripting, SQL Injection, Command Injection, Path Traversal and insecure server configuration.
So, here this can be achieved by installing ZAPCLI

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ COPY . /epicli
 RUN : INSTALL APT REQUIREMENTS \
     && apt-get update \
     && apt-get install --no-install-recommends -y \
-        autossh wget default-jre curl gcc jq libcap2-bin libc6-dev libffi-dev make musl-dev openssh-client procps psmisc rsync ruby-full sudo tar unzip vim \
+        autossh curl gcc jq libcap2-bin libc6-dev libffi-dev make musl-dev openssh-client procps psmisc rsync ruby-full sudo tar unzip vim \
 \
     && : INSTALL HELM BINARY \
     && curl -fsSLO https://get.helm.sh/helm-v${HELM_VERSION}-linux-amd64.tar.gz \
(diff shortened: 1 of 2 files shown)


seriva:
Hey! Thanks for the contribution. Can you add a description what this would be for pls?

seriva:
How is this related directly to Epiphany as our container is just used to deploy infrastructure, not directly manage/scan applications.

Dockerfile
```diff
@@ -15,7 +15,7 @@ COPY . /epicli
 RUN : INSTALL APT REQUIREMENTS \
     && apt-get update \
     && apt-get install --no-install-recommends -y \
-        autossh curl gcc jq libcap2-bin libc6-dev libffi-dev make musl-dev openssh-client procps psmisc rsync ruby-full sudo tar unzip vim \
+        autossh wget default-jre curl gcc jq libcap2-bin libc6-dev libffi-dev make musl-dev openssh-client procps psmisc rsync ruby-full sudo tar unzip vim \
```
to-bar: We would like to keep alphabetic order.

Dockerfile
```diff
@@ -63,6 +63,9 @@ RUN : INSTALL APT REQUIREMENTS \
     && mv /epicli/cli/epicli /bin/epicli \
     && chmod +x /bin/epicli
 
+RUN echo "Installing zap scan binary ..."
```
to-bar: These two commands can be combined to single one in order to create only one layer. For more details, please see https://docs.docker.com/develop/develop-images/dockerfile_best-practices/.

to-bar:
I would recommend to use separate (dedicated) image for scanning.

seriva:
> I would recommend to use separate (dedicated) image for scanning.

Yeah, it can be based  on our image using:

`
FROM epiphanyplatform/epicli:x.x.x
....
`

seriva:
@Harshini097t for me this is something that does not need to be part of our repository as this seems really project specific.  @to-bar would you agree with that?

to-bar:
> @Harshini097t for me this is something that does not need to be part of our repository as this seems really project specific. @to-bar would you agree with that?

I agree, one can just run `pip3 install zapcli` in their container if ZAP is needed or keep this Dockerfile in their fork or another repo. I'm not aware of business case in our organization that would require ZAP.

seriva:
Closing because of lack of feedback. Please recreate the pull if this is still relevant.

PR #2761, opened 2021-11-29
Issue #2747 upgrade node_exporter

romsok24:
This PR was created to satisfy the work scope described in [EPI#2747](https://github.com/epiphany-platform/epiphany/issues/2747), that is to upgrade Prometheus Node Exporter to the newest version.

What was done:
- requrements.... file --> updated for every distro 
- tested with epicli on ubu18
- run the: ```epicli test -b ....... -g <component ie. node_exporter>```
- tested ```epicli upgrade -b build/dir/```  (node_exporter is not in the scope of epicli upgrade command )

PR #3111, opened 2022-04-26
[1.0.3] Fix Node Exporter errors

atsikham:
* Created absent dir.
* Verified the list of used flags.
* Modified escaping like in the latest version.

### DIFF
--- a/core/src/epicli/data/common/ansible/playbooks/roles/node_exporter/tasks/install-node-exporter-as-system-service.yml
+++ b/core/src/epicli/data/common/ansible/playbooks/roles/node_exporter/tasks/install-node-exporter-as-system-service.yml
@@ -25,7 +25,7 @@
   vars:
     file_name: "{{ exporter_file_name }}"
 
-- name: Create /opt/node_exporter directories
+- name: Create node_exporter and textfile_collector directories
   file:
     path: "{{ item }}"
     recurse: true
@@ -33,8 +33,9 @@
     group: "node_exporter"
     mode: u=rwx,g=rx,o=
     state: directory
-  with_items:
+  loop:
     - /opt/node_exporter
+    - /var/lib/prometheus/node-exporter/textfile_collector
 
 - name: Unpack node_exporter binary
   unarchive:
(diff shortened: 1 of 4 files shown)


to-bar (dismissed):

core/src/epicli/data/common/ansible/playbooks/roles/node_exporter/tasks/install-node-exporter-as-system-service.yml
```diff
@@ -25,16 +25,17 @@
...
+- name: Create node_exporter and textfile_collector directories
   file:
     path: "{{ item }}"
     recurse: true
     owner: root
     group: "node_exporter"
     mode: u=rwx,g=rx,o=
     state: directory
-  with_items:
+  loop:
```
to-bar, line 36: I would consider separating permissions, like:
```
  loop:
    - {path: /opt/node_exporter, mode: u=rwx,g=rx,o=}
    - {path: /var/lib/prometheus/node-exporter/textfile_collector, mode: ug=rwx,o=rx}
```

since for `textfile_collector` directory I see [they use `775`](https://github.com/cloudalchemy/ansible-node-exporter/blob/c2401cfed554906843520cbac7ae2a3a132c590e/tasks/configure.yml#L40).

atsikham: This is done [the same](https://github.com/epiphany-platform/epiphany/pull/2761/files) as in newer versions. If this is necessary to change, shoudn't it be fixed in latest and backported? But in general, I would not extend permissions until this is required.

to-bar: For standard metrics, I think you are right. For custom, we require `sudo` so being member of `node_exporter` group will not allow you to export custom metrics. Can be improved one day - maybe on request 😉

rafzei (dismissed):

to-bar (approved):

rafzei (approved):

Issue #3105, opened 2022-04-25
[FEATURE REQUEST][download-requirements] Manifest file parsing


sbbroot:
**Is your feature request related to a problem? Please describe.**
In order to optimize downloading requirements, we want to download only what's needed. To do so we need to parse the manifest file produced by `epicli init` and/or `epicli prepare`. There also should be added verbose flag to the script `--verbose/-v` which prints out current enabled mapping and also requirements which will be downloaded after parsing is complete.

**Describe the solution you'd like**
- an input arg for the download-script should be added called `--manifest/-m` which accepts a Path to the manifest file produced by `epicli init` and/or `epicli prepare`
- the manifest file should be parsed by pyYaml and stored in a separate class tight to the src/config/config.py's Config class.
- a verbose flag should be added called `--verbose/-v` which will print out enabled roles mapping and requirements which will be downloaded

**Describe alternatives you've considered**
.

**Additional context**
This is related to issues: (#3103), (#3097)

---

**DoD checklist**

- Changelog
  - [x] updated
  - [ ] not needed
- COMPONENTS.md
  - [ ] updated
  - [x] not needed
- Schema
  - [x] updated
  - [ ] not needed
- Backport tasks
  - [ ] created
  - [x] not needed
- Documentation
  - [ ] added
  - [x] updated
  - [ ] not needed
- [ ] Feature has automated tests
- [x] Automated tests passed (QA pipelines)
  - [x] apply
  - [x] upgrade
  - [ ] backup/restore
- [x] Idempotency tested
- [x] All conversations in PR resolved
- [ ] Solution meets requirements and is done according to design doc
- [ ] Usage compliant with license

PR #3119, opened 2022-04-29
[2.0.1] Split available_roles and roles_mapping into separate yaml documents …

sbbroot:
…(#3097)

* available_roles splitted into roles-mapping and roles-enabled
  documents

* roles-mapping added to the Init by default

Issue #3103, opened 2022-04-25
[FEATURE REQUEST][download-requirements] Add config files merging to `epicli prepare`


sbbroot:
**Is your feature request related to a problem? Please describe.**
Currently we support merging only for `epicli apply` and `epicli upgrade` but not for `epicli prepare`. This is required for the download-requirement script to get the data about roles mapping and other info for downloading only the needed requirements for the user.

**Describe the solution you'd like**
Merging should be refactored and used across all three modes of apply/upgrade/prepare

**Describe alternatives you've considered**
.

**Additional context**
This is related to (#3097)

---

**DoD checklist**

- Changelog
  - [ ] updated
  - [ ] not needed
- COMPONENTS.md
  - [ ] updated
  - [ ] not needed
- Schema
  - [ ] updated
  - [ ] not needed
- Backport tasks
  - [ ] created
  - [ ] not needed
- Documentation
  - [ ] added
  - [ ] updated
  - [ ] not needed
- [ ] Feature has automated tests
- [ ] Automated tests passed (QA pipelines)
  - [ ] apply
  - [ ] upgrade
  - [ ] backup/restore
- [ ] Idempotency tested
- [ ] All conversations in PR resolved
- [ ] Solution meets requirements and is done according to design doc
- [ ] Usage compliant with license

sbbroot:
I'm closing this issue as merging for the `epicli prepare` is no longer needed. The download-requirements script will now use the unmerged `input-manifest` rather than merged version.

PR #3130, opened 2022-05-05
[2.0.1] Add manifest file parsing (#3105)

sbbroot:
* Add `-m/--manifest` flag to accept manifest.yml produced by
  `epicli init/prepare`

* Add `-v/--verbose` mode for printing out parsed manifest data

* Add ManifestReader class used for paring the manifest.yml file

* Move src/command/*.py to debian/redhat subdirs where needed


seriva:
@sbbroot Can you resolve the conflicts before review?

seriva (dismissed):

seriva:
@sbbroot Could we pls make a feature branch for this for now and not merge it straight into develop?

seriva:
Also @sbbroot 2 scenarios to think about:

1. People are using it after the prepare step, so they will include their manifest.yml file. Just wondering if in that case there should be a merge/validate procedure in place like we have for apply etc.
2.  People are just including the manifest in apply. do we want apply to also only generate the partial downloading of the requirements from the start?

sbbroot:
@seriva 

1. Yes, there is a separate issue for this: [#3103](https://github.com/epiphany-platform/epiphany/issues/3103)
2. I assume "yes", default manifest.yml alone provides enough information to reduce the number of downloaded files.

ansible/playbooks/roles/repository/files/download-requirements/src/config/config.py
```diff
@@ -93,6 +101,12 @@ def __create_parser(self) -> ArgumentParser:
         parser.add_argument('--no-logfile', action='store_true', dest='no_logfile',
                             help='no logfile will be created')
 
+        parser.add_argument('--verbose', '-v',  action='store_true', dest='verbose',
+                            help='more verbose output will be provided')
+
+        parser.add_argument('--manifest', '-m', metavar='MANIFEST_PATH', type=Path, action='store', dest='manifest',
+                            help='manifest file generated by `epicli apply` and/or `epicli prepare`')
```
to-bar: Is it possible to use manifest generated by `epicli init --full` or merged version is always required?

ansible/playbooks/roles/repository/files/download-requirements/src/config/manifest_reader.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+def load_yaml_file(filename: Path) -> Any:
+    return load_yaml_file_all(filename)[0]
+
+
+class ManifestReader:
+    """
+    Load the manifest file and call defined parser methods to process required documents.
+    Main running method is :func:`~manifest_reader.ManifestReader.parse_manifest` which returns formatted manifest output.
```
to-bar, line 26: Does `~` here has a special meaning or this is typo?

ansible/playbooks/roles/repository/tasks/download-requirements.yml
```diff
@@ -11,6 +11,8 @@
         /var/www/html/epirepo \
         "{{ download_requirements_os_name }}" \
         --repos-backup-file /var/tmp/enabled-system-repos.tar \
+        --verbose \
```
to-bar: How about alphabetical order?

ansible/playbooks/roles/repository/files/download-requirements/src/config/manifest_reader.py
sbbroot: Yes, this has a special meaning in pydoc, it creates a link to that method.

ansible/playbooks/roles/repository/files/download-requirements/src/config/config.py
sbbroot: The only minimum data required is: sections `epiphany-cluster` and `configuration/feature-mappings`.
Mappings is implemented in #3119

to-bar (approved):

erzetpe (approved):

Issue #3131, opened 2022-05-06
[FEATURE REQUEST][download-requirements] Optimize Grafana dashboards downloading


sbbroot:
**Is your feature request related to a problem? Please describe.**
Currently we download all dashboards as is, but we would like to download only when Grafana service is used.

**Describe the solution you'd like**
Add check after parsing the `manifest` to remove dashboards from the requirements list if there is no Grafana.
Add info about which dashboards will be downloaded when the `-v/--verbose` is used.

**Describe alternatives you've considered**
.

**Additional context**
.

---

**DoD checklist**

- Changelog
  - [x] updated
  - [ ] not needed
- COMPONENTS.md
  - [ ] updated
  - [x] not needed
- Schema
  - [x] updated
  - [ ] not needed
- Backport tasks
  - [ ] created
  - [x] not needed
- Documentation
  - [ ] added
  - [x] updated
  - [ ] not needed
- [ ] Feature has automated tests
- [x] Automated tests passed (QA pipelines)
  - [x] apply
  - [x] upgrade
  - [ ] backup/restore
- [x] Idempotency tested
- [x] All conversations in PR resolved
- [ ] Solution meets requirements and is done according to design doc
- [ ] Usage compliant with license

PR #3150, opened 2022-05-17
[2.0.1] Optimize Grafana dashboards downloading (#3131)

sbbroot:


ansible/playbooks/roles/repository/files/download-requirements/src/config/config.py
```diff
@@ -231,6 +231,13 @@ def __print_parsed_manifest_data(self, output: Dict[str, Any]):
         for feature in output['detected-features']:
             lines.append(f'- {feature}')
 
+        dashboards = requirements['grafana-dashboards']
+        if dashboards:
+            lines.append('')
```
rafzei, line 236: \n instead is possible?

erzetpe (approved):

ansible/playbooks/roles/repository/files/download-requirements/src/config/config.py
sbbroot: Lines are joined using '\n', so if you'd use '\n' you would get '\n\n' instead. See line 234/241.

rafzei: I mean lines.append('\nDashboards to download:') but up to you :v:

rafzei (approved):

Issue #3143, opened 2022-05-12
[FEATURE REQUEST] Update ARM doc for 2.0.1


rafzei:
**Is your feature request related to a problem? Please describe.**
ARM installation moved from CentOS to AlmaLinux but the documentation does not reflect this.

**Describe the solution you'd like**
Update [ARM doc](https://github.com/epiphany-platform/epiphany/blob/develop/docs/home/ARM.md)

**Describe alternatives you've considered**
*

**Additional context**
*

---

**DoD checklist**

- Changelog
  - [ ] updated
  - [ ] not needed
- COMPONENTS.md
  - [ ] updated
  - [ ] not needed
- Schema
  - [ ] updated
  - [ ] not needed
- Backport tasks
  - [ ] created
  - [ ] not needed
- Documentation
  - [ ] added
  - [ ] updated
  - [ ] not needed
- [ ] Feature has automated tests
- [ ] Automated tests passed (QA pipelines)
  - [ ] apply
  - [ ] upgrade
  - [ ] backup/restore
- [ ] Idempotency tested
- [ ] All conversations in PR resolved
- [ ] Solution meets requirements and is done according to design doc
- [ ] Usage compliant with license

seriva:
Fixed with #3151

PR #3151, opened 2022-05-17
[2.0.1] Add ARM architecture support for AlmaLinux 8.4

plirglo:


seriva:
@plirglo I think that this one: https://github.com/epiphany-platform/epiphany/issues/3143 will also be solved with this pull?

plirglo:
> 

Correct, i updated this document

seriva:
> > 
> 
> Correct, i updated this document

Perfect, I updated and connected the issue to reflect this. thx.

seriva (dismissed):

ansible/playbooks/roles/repository/files/download-requirements/src/mode/red_hat_family_mode.py
```diff
@@ -98,14 +98,16 @@ def _add_third_party_repositories(self):
...
-        for repo in ['https://dl.2ndquadrant.com/default/release/get/10/rpm',  # for repmgr
-                     'https://dl.2ndquadrant.com/default/release/get/13/rpm']:
-            Command('curl', self._cfg.retries, [repo]) | Command('bash', self._cfg.retries)  # curl {repo} | bash
-
-        # script adds 2 repositories, only 1 is required
-        for repo in ['2ndquadrant-dl-default-release-pg10-debug',
-                     '2ndquadrant-dl-default-release-pg13-debug']:
-            self._tools.dnf_config_manager.disable_repo(repo)
+        # repmgr is supported only with x86_64 architecture
+        if 'x86_64' in self.__archs:
```
sbbroot: We don't want to use magic consts.

Please change this conditional in such a way:
if self._cfg.os_arch == OSArch.X86_64

Don't forget to import OSArch type on top of the file.
from src.config.os_type import OSArch

ansible/playbooks/roles/repository/files/download-requirements/src/config/os_type.py
```diff
@@ -5,7 +5,7 @@
 class OSArch(Enum):
     """ Supported architecture types """
     X86_64 = 'x86_64'
-    ARM64 = 'arm64'
+    ARM64 = 'aarch64'
```
sbbroot, line 8: Didn't we want to have `arm64` in all of our code instead of `aarch64`? Correct me if I'm wrong.

sbbroot (changes requested):

ansible/playbooks/roles/repository/files/download-requirements/src/config/os_type.py
plirglo: I used `aarch64` here because ansible_architecture variable, which in almalinux systems has value: `aarch64`
image

image
I used `aarch64` to be consistent with "real value" from system

sbbroot (approved):

seriva (approved):

PR #3162, opened 2022-05-25
Fix for upgrade from v1.0.3 LTS to v1.3.1

rafzei:


cicharka (dismissed):

ansible/playbooks/roles/docker/tasks/main.yml
```diff
@@ -1,6 +1,32 @@
...
+
+    - name: Print containerd.io versions
+      debug:
+        msg:
+          - "Installed version: {{ containerd_installed }}"
+          - "Target version: {{ containerd_version }}"
+
+    # apt module doesn't support --allow-downgrades so we remove packages as workaround
+    - name: Remove newer containerd.io package installed as dependency if they exist
+      apt:
```
to-bar, line 26: How about adding condition to run `apt` module only on Debian OS family or using `package` module instead?

seriva (approved):

to-bar (approved):

przemyslavic (approved):

PR #3181, opened 2022-06-02
[2.0.1] Fix `use_network_security_groups` is set to `false`

seriva:
- Fix use_network_security_groups is set to false
- Also minor fix for SonarQube reported bug: https://sonarcloud.io/project/issues?resolved=false&types=BUG&id=epiphany-platform_epiphany&open=AYEfJldDUaTriuEfx6nc

### DIFF
--- a/cli/src/spec/SpecCommand.py
+++ b/cli/src/spec/SpecCommand.py
@@ -58,5 +58,4 @@ These need to be installed to run the cluster spec tests from epicli'''
         for entry in listdir:
             if os.path.isdir(f'{SPEC_TEST_PATH}/spec/{entry}'):
                 groups = groups + [entry]
-        sorted_group = sorted(groups, key=str.lower)
-        return sorted_group
+        return sorted(groups, key=str.lower)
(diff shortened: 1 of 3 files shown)


norbix (dismissed):

cli/src/spec/SpecCommand.py
```diff
@@ -58,5 +58,5 @@ def get_spec_groups():
         for entry in listdir:
             if os.path.isdir(f'{SPEC_TEST_PATH}/spec/{entry}'):
                 groups = groups + [entry]
-        sorted(groups, key=str.lower)
-        return groups
+        sorted_group = sorted(groups, key=str.lower)
+        return sorted_group
```
sbbroot: Just return sorted()? I don't think there is a need for the sorted_group variable.

sbbroot (changes requested):

cli/src/spec/SpecCommand.py
seriva: Good point:)

przemyslavic (approved):

sbbroot (approved):

PR #3188, opened 2022-06-14
Adaptive mode for downloading requirements

sbbroot:


ansible/playbooks/roles/repository/tasks/copy-download-requirements.yml
```diff
@@ -55,6 +55,11 @@
         dest: "{{ download_requirements_dir }}/{{ item }}"
       loop: "{{ _files }}"
 
+    - name: Copy the manifest file
+      synchronize:
```
to-bar, line 59:
```suggestion
      copy:
```
Isn't `synchronize` overkill for a single file?

cli/src/commands/Upgrade.py
```diff
@@ -23,6 +25,8 @@ def __init__(self, input_data):
         self.input_docs = []
         self.ping_retries: int = input_data.ping_retries
 
+        Config().input_manifest_path = Path(self.file) or Path(self.build_dir) / 'manifest.yml'
```
to-bar: Let's talk about this line 😉

przemyslavic: Good catch :)
```
Manifest used: /var/tmp/epi-download-requirements/manifest.yml", "Verbose mode: True", "Retries count: 3", "--------------------------------------------------", "2022-06-15 08:04:27,850 [ERROR]: ManifestReader - could not find document(s): {'configuration/feature-mappings'}"
```

ansible/playbooks/roles/repository/tasks/copy-download-requirements.yml
seriva: Agreed ^

seriva (approved):

ansible/playbooks/roles/repository/tasks/copy-download-requirements.yml
sbbroot: I don't understand why it's an overkill? Just because rsync has more options which I don't use here? Nevertheless it offers same speed for a small file. I just wanted to be consistent here and don't mix copy with synchronize.

cli/src/commands/Upgrade.py
sbbroot: No longer valid.

przemyslavic (approved):

PR #3193, opened 2022-06-22
[3190] - feat: Introducing new field 'specification.enable_controller…

norbix:
…_attach_detach' for K8s worker nodes kubelet config.


norbix:
Guys, how to test eval of 'specification.enable_controller_attach_detach' field ?

seriva:
> Guys, how to test eval of 'specification.enable_controller_attach_detach' field ?

Should be added to defaults and validation:
https://github.com/epiphany-platform/epiphany/blob/develop/schema/common/defaults/configuration/kubernetes-master.yml
https://github.com/epiphany-platform/epiphany/blob/develop/schema/common/validation/configuration/kubernetes-master.yml

ansible/playbooks/roles/kubernetes_master/templates/kubeadm-config.yml.j2
```diff
@@ -61,3 +61,4 @@ kind: KubeletConfiguration
 apiVersion: kubelet.config.k8s.io/v1beta1
 cgroupDriver: systemd
 rotateCertificates: true
+enableControllerAttachDetach: {{ specification.enable_controller_attach_detach }}
```
cicharka: You have to verify if in any ansible task in kubernetes_master we import specification from kubernetes-node - otherwise variable `enable_controller_attach_detach` could be not accessible.

ansible/playbooks/roles/kubernetes_master/templates/kubeadm-config.yml.j2
```diff
@@ -61,3 +61,5 @@ kind: KubeletConfiguration
 apiVersion: kubelet.config.k8s.io/v1beta1
 cgroupDriver: systemd
 rotateCertificates: true
+#enableControllerAttachDetach: {{ specification.extensions.k8s_worker_nodes_capabilities.enable_controller_attach_detach }}
```
cicharka: Adding comment with single `#` sign (or with double `##`) in j2 template will not make j2 engine think that the line is a comment -> it will try to render that line, and access undefined variable which will result in playbook failure "AnsibleUndefinedVariable: 'dict object' has no attribute 'extensions'".

In order to comment line in j2 template you have to use `{# statement #}` syntax.

sbbroot:
Please fix git issues first, this PR is unreadable.

PR #3202, opened 2022-06-24
Allow excluding test groups

to-bar:


cli/src/commands/Test.py
```diff
@@ -40,8 +43,29 @@ def test(self):
...
+        if selected_test_groups:
+            spec_command = SpecCommand()
+            if 'all' in selected_test_groups:
+                selected_test_groups = ['all']
+            else:
+                self.logger.info(f'Selected test groups: {", ".join(selected_test_groups)}')
+
+            spec_command.run(spec_output, path_to_inventory, admin_user.name, admin_user.key_path, selected_test_groups)
+        else:
+            raise Exception('No test group to test')
```
seriva: `No test group specified to run`  maybe? As your not testing a test group but running a test group to test.

cli/src/helpers/argparse_helpers.py
```diff
@@ -0,0 +1,13 @@
+import argparse
+
+
```
seriva, line 3: Maybe add some comments while refactoring this?

seriva (changes requested):

seriva (approved):

sbbroot (approved):

PR #3206, opened 2022-07-01
Provide kubeconfig file for spec tests

to-bar:


seriva (dismissed):

cli/src/commands/Test.py
```diff
@@ -23,16 +28,62 @@ def __enter__(self):
...
+        inventory_groups = load_inventory(self.inventory_path).list_groups()
+        if append_implicit:
+            inventory_groups.append('common')
+        return inventory_groups
+
+    def __get_available_test_groups(self) -> list[str]:
+        inventory_groups = self.__get_inventory_groups(True)
+        return [group for group in self.all_groups if group in inventory_groups]
+
+
```
sbbroot: remove endline

cli/src/commands/Test.py
```diff
@@ -23,16 +28,62 @@ def __enter__(self):
...
+    def __get_selected_test_groups(self) -> list[str]:
```
sbbroot, line 48:
```python
-> list[str]:
```
This is not officially supported with < python3.10
I suggest to stick with the `List` type from the `typing` module.

cli/src/commands/Test.py
```diff
@@ -23,16 +28,62 @@ def __enter__(self):
...
+        selected_groups = self.included_groups
+
+        if 'all' in selected_groups:
+            selected_groups = ['all']
```
sbbroot: Make it a single line:

```python
selected_groups = ['all'] if 'all' in self.included_groups else self.included_groups
```

cli/src/commands/Test.py
```diff
@@ -23,16 +28,62 @@ def __enter__(self):
...
+
+        # exclude test groups
+        if self.excluded_groups:
+            included_groups = self.included_groups
+            if 'all' in included_groups:
+                included_groups = self.available_groups
```
sbbroot:
```python
included_groups = self.available_groups if 'all' in self.included_groups else self.included_groups
```

cli/src/commands/Test.py
```diff
@@ -23,16 +28,62 @@ def __enter__(self):
...
+            if any(group in kubectl_groups for group in self.selected_groups):
+                return True
+            if 'all' in self.selected_groups and any(group in kubectl_groups for group in self.available_groups):
+                return True
+
+        return False
+
+    def __prepare_env(self):
+        if self.__is_env_preparation_needed():
+            self.ansible_command = AnsibleCommand()
```
sbbroot: Declaring new object members not in a constructor is probably a bad idea.

cli/src/commands/Test.py
```diff
@@ -23,16 +28,62 @@ def __enter__(self):
...
+            playbook_path = str(SPEC_TESTS_PATH) + '/pre_run/ansible/kubernetes_master/copy-kubeconfig.yml'
+            self.ansible_command.run_playbook(inventory=self.inventory_path,
+                                              playbook_path=playbook_path,
+                                              extra_vars=[f'kubeconfig_remote_path={self.kubeconfig_remote_path}'])
```
sbbroot:
```python
            AnsibleCommand().run_playbook(inventory=self.inventory_path,
                                                                  playbook_path=playbook_path,
                                                                  extra_vars=[f'kubeconfig_remote_path={self.kubeconfig_remote_path}'])
```

cli/src/commands/Test.py
```diff
@@ -23,16 +28,62 @@ def __enter__(self):
...
+
+    def __clean_up_env(self):
+        if self.__is_env_preparation_needed():
+            playbook_path = str(SPEC_TESTS_PATH) + '/post_run/ansible/kubernetes_master/undo-copy-kubeconfig.yml'
+            self.ansible_command.run_playbook(inventory=self.inventory_path,
```
sbbroot: Same as in `prepare_env()`

cli/src/commands/Test.py
```diff
@@ -43,28 +94,16 @@ def test(self):
...
-            else:
-                self.logger.info(f'Selected test groups: {", ".join(selected_test_groups)}')
 
-            spec_command.run(spec_output, path_to_inventory, admin_user.name, admin_user.key_path, selected_test_groups)
+            if 'all' not in self.selected_groups:
+                self.logger.info(f'Selected test groups: {", ".join(self.selected_groups)}')
+
+            spec_command.run(spec_output, self.inventory_path, admin_user.name, admin_user.key_path, self.selected_groups)
+            self.__clean_up_env()
         else:
```
sbbroot:
```python 
else
```
Not needed, if we will not reach the `if` statement this method will always raise an exception.
Line 71/110 will never be reached.

cli/src/commands/Test.py
```diff
@@ -1,20 +1,25 @@
 import os
```
sbbroot, line 1: It would be nice to add some pydoc to this module.

sbbroot (changes requested):

cli/src/commands/Test.py
to-bar: `epicli` image is based on `python:3.10-slim` image.

cli/src/commands/Test.py
to-bar: The idea was to have lazy init, since the object is needed conditionally - based on result from `__is_env_preparation_needed()`.

cli/src/commands/Test.py
to-bar: Added.

sbbroot (approved):

seriva (approved):

Issue #3207, opened 2022-07-05
[FEATURE REQUEST][download-requirements] Add filtering mechanism for the sensitive data


sbbroot:
**Is your feature request related to a problem? Please describe.**
Currently we copy manifest.yml file to the remote "as is". This is a potential security issue with the manifest holding multiple passwords as well as cloud provider info.

**Describe the solution you'd like**
Add mechanism for filtering the sensitive data on the Ansible side.

**Describe alternatives you've considered**
-

**Additional context**
An Ansible module would be a good solution here.

---

**DoD checklist**

- Changelog
  - [x] updated
  - [ ] not needed
- COMPONENTS.md
  - [ ] updated
  - [x] not needed
- Schema
  - [ ] updated
  - [x] not needed
- Backport tasks
  - [ ] created
  - [x] not needed
- Documentation
  - [ ] added
  - [ ] updated
  - [x] not needed
- [ ] Feature has automated tests
- [ ] Automated tests passed (QA pipelines)
  - [x] apply
  - [ ] upgrade
  - [ ] backup/restore
- [ ] Idempotency tested
- [x] All conversations in PR resolved
- [ ] Solution meets requirements and is done according to design doc
- [ ] Usage compliant with license

PR #3208, opened 2022-07-05
Add filtering mechanism for the sensitive data (#3207)

sbbroot:


seriva (dismissed):

ansible/playbooks/roles/repository/library/filter_credentials.py
```diff
@@ -0,0 +1,131 @@
...
+        required: true
+        type: str
+    dest:
+        description: Path to the newly created, filtered manifest
+        required: false
+        type: str
+    in_place:
+        description: When set to True will modify existing manifest passed as `src`
+        required: false
+        type: false
```
to-bar:
```suggestion
        type: bool
        default: false
```

seriva:
@to-bar I think this filter should cover everything, I went quickly through the defaults and besides **cloud** and **...password...** I don't think anything more is there. Or maybe you know something more:)

@sbbroot Maybe we can also apply this to the dump_debug_info function in epicli.py:

image
image

Would be nice to kill 2 birds with one stone.

ansible/playbooks/roles/repository/library/filter_credentials.py
```diff
@@ -0,0 +1,131 @@
...
+import yaml
+from ansible.module_utils.basic import AnsibleModule
+
+
+def _get_filtered_manifest(manifest_path: Path) -> str:
+    """
+    Load the manifest file and remove any sensitive data.
+
+    :param manifest_path: manifest file which will be loaded
+    :returns: filtered manifset doc
```
to-bar:
```suggestion
    :returns: filtered manifest
```

ansible/playbooks/roles/repository/library/filter_credentials.py
```diff
@@ -0,0 +1,131 @@
...
+        for line in mhandler.readlines():
+            if 'password' not in line.lower():
+                lines.append(line)
+
+    docs = yaml.safe_load_all(''.join(lines))
+    epiphany_cluster = next(docs)
+
+    # remove provider credentials:
+    try:
+        del epiphany_cluster['specification']['cloud']
```
to-bar: How about removing only `credentials` subkey?

ansible/playbooks/roles/repository/tasks/copy-download-requirements.yml
```diff
@@ -55,9 +55,16 @@
         dest: "{{ download_requirements_dir }}/{{ item }}"
       loop: "{{ _files }}"
 
+    - name: Filter sensitive data from the manifest
+      filter_credentials:
+        src: "{{ input_manifest_path }}"
+        dest: /tmp/filtered_manifest.yml
+      delegate_to: localhost
+      when: not full_download and input_manifest_path
```
to-bar, line 59: The next task has the same condition so I would suggest to use `block` to have just one `when`.

to-bar (changes requested):

to-bar:
> @to-bar I think this filter should cover everything, I went quickly through the defaults and besides **cloud** and **...password...** I don't think anything more is there. Or maybe you know something more:)

I have `grep`-ed `schema` subdir using `secret` pattern. Found some keys but by default they are commented out, e.g:
```
  - name: rabbitmq
    #image_pull_secret_name: regcred # optional
```
```
#  image_registry_secrets:
#  - email: emaul@domain.com
#    name: secretname
#    namespace: default
#    password: docker-registry-pwd
#    server_url: docker-registry-url
#    username: docker-registry-user
```
Maybe emails should also be filtered out?

```
#     - name: 'opsgenie'
#       opsgenie_config:
#         api_key: <secret> | default = global.opsgenie_api_key
#         api_url: <string> | default = global.opsgenie_api_url
```

ansible/playbooks/roles/repository/library/filter_credentials.py
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/python
+
+from __future__ import (absolute_import, division, print_function)
+
+from pathlib import Path
+from typing import Callable, Dict, List
```
to-bar: We support only Python 3.10 so you can consider using just `list` and `dict`.

ansible/playbooks/roles/repository/library/filter_credentials.py
```diff
@@ -0,0 +1,148 @@
...
+                                                            'configuration/feature-mappings',
+                                                            'configuration/image-registry']]
+
+    FILTER_DATA: Dict[str, Callable] = {
+        'any': _filter_common,
+        'azure': _filter_azure,
+        'aws': _filter_aws
+    }
+
+    FILTER_DATA[filtered_docs[0]['provider']](filtered_docs)
```
to-bar, line 105: @seriva can input manifest contain multiple providers or we always have just one?

ansible/playbooks/roles/repository/library/filter_credentials.py
```diff
@@ -0,0 +1,148 @@
...
+        module.fail_json(msg='Cannot use `in_place` and `dest` at the same time', **result)
+
+    if not module.params['in_place'] and not module.params['dest']:  # to stdout
+        result['manifest'] = manifest
+    elif module.params['in_place']:  # overwrite existing manifest
+        with input_manifest.open(mode='w', encoding='utf-8') as mhandler:
+            mhandler.write(manifest)
+    elif module.params['dest']:  # write to a new location
+        with Path(module.params['dest']).open(mode='w', encoding='utf-8') as output_manifest:
+            output_manifest.write(manifest)
```
to-bar:
```suggestion
        with Path(module.params['dest']).open(mode='w', encoding='utf-8') as output_file:
            output_file.write(manifest)
```

ansible/playbooks/roles/repository/library/filter_credentials.py
```diff
@@ -0,0 +1,148 @@
...
+
+    result['changed'] = True
```
to-bar: IMO to support this properly we should compare content before and after filtering. If manifest was already filtered the task should not be marked as changed as content would be the same.

.devcontainer/python.env
```diff
@@ -2,4 +2,4 @@
 
 # PYTHONPATH can contain multiple locations separated by os.pathsep: semicolon (;) on Windows and colon (:) on Linux/macOS.
 # Invalid paths are ignored. To verify use "python.analysis.logLevel": "Trace".
-PYTHONPATH=ansible/playbooks/roles/repository/files/download-requirements
+PYTHONPATH=ansible/playbooks/roles/repository:ansible/playbooks/roles/repository/files/download-requirements
```
to-bar:
```suggestion
PYTHONPATH=ansible/playbooks/roles/repository/library:ansible/playbooks/roles/repository/files/download-requirements
```

ansible/playbooks/roles/repository/library/filter_credentials.py
```diff
@@ -0,0 +1,172 @@
...
+
+    :param manifest_path: manifest file which will be loaded
+    :returns: filtered manifset
+    """
+    docs = yaml.safe_load_all(manifest_path.open())
+    filtered_docs = [doc for doc in docs if doc['kind'] in ['epiphany-cluster',
+                                                            'configuration/feature-mappings',
+                                                            'configuration/image-registry']]
+
+    FILTER_DATA: list[str, Callable] = {
```
to-bar:
```suggestion
    FILTER_DATA: dict[str, Callable] = {
```

to-bar (changes requested):

ansible/playbooks/roles/repository/library/filter_credentials.py
sbbroot: We support only single provider in a manifest.

to-bar (dismissed):

to-bar (dismissed):

seriva:
> @to-bar I think this filter should cover everything, I went quickly through the defaults and besides **cloud** and **...password...** I don't think anything more is there. Or maybe you know something more:)
> 
> @sbbroot Maybe we can also apply this to the dump_debug_info function in epicli.py:
> 
> image image
> 
> Would be nice to kill 2 birds with one stone.

As discussed with @sbbroot lets defer this to a separate issue as this is not entirely related to this issue. @sbbroot pls create it:)

seriva (dismissed):

przemyslavic (approved):

seriva (approved):

Issue #3222, opened 2022-07-15
[BUG] Adaptive mode: not all files are downloaded when `k8s_as_cloud_service` flag is used


przemyslavic:
**Describe the bug**
There is a bug in adaptive mode causing that not all necessary files are downloaded when `k8s_as_cloud_service` flag is set to `true`.

**How to reproduce**
1. Prepare manifest file: enable only repository component and set `k8s_as_cloud_service` flag to true
2. Run download-requirements.py script

Output:
```
2022-07-15 13:15:22,603 [INFO]: Manifest summary:
--------------------------------------------------
Components requested:
- repository

Features requested:
- filebeat
- firewall
- image-registry
- node-exporter
- repository

Files to download:
- https://charts.bitnami.com/bitnami/node-exporter-2.3.17.tgz
- https://charts.rook.io/release/rook-ceph-cluster-v1.8.8.tgz
- https://charts.rook.io/release/rook-ceph-v1.8.8.tgz
- https://helm.elastic.co/helm/filebeat/filebeat-7.12.1.tgz

Images to download:
- registry:2.8.0
```
There is no node-exporter binary included.

When the flag is set to false then node exporter appears:
```
2022-07-15 13:16:22,109 [INFO]: Manifest summary:
--------------------------------------------------
Components requested:
- repository

Features requested:
- filebeat
- firewall
- image-registry
- node-exporter
- repository

Files to download:
- https://github.com/prometheus/node_exporter/releases/download/v1.3.1/node_exporter-1.3.1.linux-amd64.tar.gz

Images to download:
- registry:2.8.0
```
The same with logging component.
When the flag is set to false there are listed a few files related to opensearch:
```
Files to download:
- https://artifacts.opensearch.org/releases/bundle/opensearch-dashboards/1.2.0/opensearch-dashboards-1.2.0-linux-x64.tar.gz
- https://artifacts.opensearch.org/releases/bundle/opensearch/1.2.4/opensearch-1.2.4-linux-x64.tar.gz
- https://github.com/opensearch-project/perftop/releases/download/1.2.0.0/opensearch-perf-top-1.2.0.0-linux-x64.zip
- https://github.com/prometheus/node_exporter/releases/download/v1.3.1/node_exporter-1.3.1.linux-amd64.tar.gz
```
and when the flag is set to true there are only helm chart files:
```
Files to download:
- https://charts.bitnami.com/bitnami/node-exporter-2.3.17.tgz
[...]

PR #3242, opened 2022-08-03
Fix k8s_as_cloud_service flag used in download-requirements (#3222)

sbbroot:


seriva (dismissed):

ansible/playbooks/roles/repository/files/download-requirements/requirements/x86_64/files.yml
```diff
@@ -41,19 +41,19 @@ files:
...
 
   'https://helm.elastic.co/helm/filebeat/filebeat-7.12.1.tgz':
     sha256: 5838058fe06372390dc335900a7707109cc7287a84164ca245d395af1f9c0a79
-    deps: [kubernetes-master]
+    deps: [kubernetes-master, k8s-as-cloud-service]
 
   'https://charts.rook.io/release/rook-ceph-v1.8.8.tgz':
     sha256: f67e474dedffd4004f3a0b7b40112694a7f1c2b1a0048b03b3083d0a01e86b14
-    deps: [kubernetes-master]
+    deps: [kubernetes-master, k8s-as-cloud-service]
```
cicharka: I was asked yesterday about rook and k8s_as_cloud_service by @to-bar - and my response was that we never tested rook with k8s_as_cloud_service (no requirements to do so). Therefore I would say that we don't have to add `k8s-as-cloud-service` group for `rook-ceph` and `rook-ceph-cluster` charts.

cicharka (dismissed):

ansible/playbooks/roles/repository/files/download-requirements/requirements/x86_64/files.yml
seriva: Maybe lets test if it works, if not remove?:D

cicharka (approved):
Tested by:
- using different components and analyzing downloaded files / download-requirements.py logs
- full cluster with all components (without k8s) deployed alongside AKS
- discussed with @przemyslavic test results

przemyslavic (approved):

PR #3263, opened 2022-08-23
Ansible upgrade to 6.2

seriva:
- Bump Python requirements from 3.10.4 to 3.10.6
- Locked Python version to 3.10.6
- Upgrade Ansible from 5.2 to 6.2
- Fixed linter issues with ansible-lint 6.5
- Locked ansible-lint version to 6.5
- Upgraded non versions Python packages
- For now disabled [jinja[spacing]](https://github.com/epiphany-platform/epiphany/pull/3263/commits/0e7e0878e85a54121fdc7f8d746c265146efc136) as the latest changes to this pull don`t take some cases into account like >- constructions: https://github.com/ssbarnea/ansible-lint/commit/b29342950c3cadfe10bc27f42845027c20bd345f


ansible/playbooks/roles/jmx_exporter/tasks/main.yml
```diff
@@ -13,7 +13,7 @@
     createhome: false
     system: true
 
-- name: jmx-exporter | create directories
+- name: Jmx-exporter | create directories
```
tomasz-baran:
```suggestion
- name: Create JMX Exporter directories
```

ansible/playbooks/roles/applications/tasks/deploy-k8s-app.yml
```diff
@@ -1,22 +1,22 @@
 ---
-- name: "{{ app_dir_name | capitalize }} | Assert input parameters"
+- name: "Application | {{ app_dir_name | capitalize }} | Assert input parameters"
```
tomasz-baran: IMO these changes will produce some kind of redundant output since Ansible will prepend the role name and we would have something like: `TASK [applications : Application | Rabbitmq | Assert input parameters]`. How about using `# noqa name[casing]` instead? [Link to docs](https://ansible-lint.readthedocs.io/en/latest/rules/#false-positives-skipping-rules).

ansible/playbooks/roles/upgrade/tasks/load_balancer.yml
```diff
@@ -1,20 +1,20 @@
 ---
-- name: load_balancer | Remove legacy haproxy
+- name: Load_balancer | Remove legacy haproxy
```
tomasz-baran: This is component name so I would keep the original name using `# noqa name[casing]`.

.devcontainer/pyproject.toml
```diff
@@ -5,16 +5,16 @@ description = "Epiphany Dev Container"
 authors = ["Epiphany Platform"]
 
 [tool.poetry.dependencies]
-python = "3.10.4"
+python = "3.10.6"
```
tomasz-baran, line 8: We define minor version here so how about doing the same (for consistency) in Dockerfile? I mean `FROM python:3.10-slim` -> `FROM python:3.10.6-slim` or using e.g. `python = "~3.10"` here without changing Dockerfile.

seriva: Yeah that is actually a good plan to also lock down the Python version.

tomasz-baran (dismissed):

tomasz-baran (dismissed):

tomasz-baran (approved):

przemyslavic (approved):

PR #3274, opened 2022-09-19
Upgrade Keycloak to 19.0.2

tomasz-baran:


ansible/playbooks/roles/keycloak/tasks/preflight/apply.yml
```diff
@@ -0,0 +1,38 @@
...
+          assert:
+            that:
+              - groups.postgresql is defined
+              - groups.postgresql | count > 0
+            fail_msg: At least one postgresql component has to be defined
+
+        - name: Execute if Postgresql is not standalone instance
+          when: groups.postgresql | count > 1
+          block:
+            - name: Include vars from application role
```
sbbroot, line 27: You import these vars more than once, can you import them in one place and use them globally in keycloak tasks?

tomasz-baran: Preflight tasks are run from separate `preflight` role and I would like to have possibility to run this role independently hence there is dedicated include here.  However, I've removed one include (from other place) so now there are just two.

PR #3384, opened 2023-08-17
Python updates

seriva:
- Updated Python from 3.10.6 to 3.10.12
- Update all dependencies except AzureCLI


seriva:
Created a new pull.

